PR #12, opened 2015-12-03
use latest error decoder from original foundry code

pnepywoda:


pnepywoda:
@uschi2000

uschi2000:
looks ok. @markelliot what was your rationale for changing this?

markelliot:
I think maybe Paul edited the Foundry copy after I moved this out here.

PR #14, opened 2015-12-08
Add TextDelegateDecoder to standard Feign clients

uschi2000:
@markelliot

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Optional;
 import com.palantir.remoting.http.errors.SerializableErrorErrorDecoder;
 import feign.OptionalAwareDecoder;
+import feign.TextDelegateDecoder;
 import feign.jackson.JacksonDecoder;
 import feign.jackson.JacksonEncoder;
 import feign.jaxrs.JAXRSContract;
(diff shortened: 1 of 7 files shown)


http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -67,9 +68,9 @@ public static FeignClientFactory withMapper(ObjectMapper mapper) {
         return FeignClientFactory.of(
                 new JAXRSContract(),
                 new JacksonEncoder(mapper),
-                new OptionalAwareDecoder(new JacksonDecoder(mapper), SerializableErrorErrorDecoder.INSTANCE),
+                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(mapper)),
```
markelliot: is the order here correct?

http-clients/src/test/java/feign/TextDelegateDecoderTest.java
```diff
@@ -0,0 +1,113 @@
...
+        Object decodedObject = textDelegateDecoder.decode(response, String.class);
+
+        assertEquals(decodedObject, DELEGATE_RESPONSE);
+    }
+
+    @Test
+    public void testStandardClientsUseTextDelegateEncoder() {
+        assertThat(service.getString("string"), is("string"));
+
+    }
```
markelliot, line 142: can we also cover that a null response comes back with content type text/plain and you get the right `Optional` behavior?

uschi2000, line 142: do you mean in interface like this?

```
        @Produces(MediaType.TEXT_PLAIN)
        Optional<String> getString(@QueryParam("value") String value);
```

Not really sure that actually makes sense.

uschi2000, line 142: added a few tests to clarify/specify the behaviour.

PR #17, opened 2016-01-07
FeignClientFactory can be configured with timeout Request.Options

uschi2000:


uschi2000:
@markelliot @ellisjoe

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -49,13 +51,15 @@ private FeignClientFactory(
             Decoder decoder,
             ErrorDecoder errorDecoder,
             Function<Optional<SSLSocketFactory>, Client> clientSupplier,
-            BackoffStrategy backoffStrategy) {
+            BackoffStrategy backoffStrategy,
+            Request.Options options) {
```
markelliot, line 55: should we support a default constructor which does not require providing this parameter to allow compatibility?

markelliot, line 55: nvm, i get it now

ellisjoe, line 55: The first .of method is still the same and this constructor is private so we should be ok. The only method we may want is a .of with backoffStrategy and no Request.Options but I'm not sure that's necessary.

markelliot, line 55: Yeah, I cut the release with this as is, we can add the other method later, as necessary.

PR #18, opened 2016-01-12
Remove Feign dependency from error-handling project

uschi2000:


uschi2000:
@markelliot 
technically a back-compat break.

markelliot:
jsr305 doesn't appear to be required either?

uschi2000:
it is -- for @Nullable

markelliot:
cool, thanks

PR #21, opened 2016-01-14
remove unused dependency versions in gradle.properties

### LINKED ISSUE #20: dependency versions should be removed from gradle.properties
The non-buildscript dependency versions are unused. Removing them would reduce confusion, especially when versions don't match between gradle/lib.gradle and gradle.properties (e.g. feign).

gracew:
Fixes #20: Reduces confusion especially when the versions don't match between
gradle.properties and lib.gradle.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,13 +2,4 @@
 baselineVersion=0.2.2
 bintrayPlugin=1.4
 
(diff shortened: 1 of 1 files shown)


markelliot:
:+1:

PR #26, opened 2016-01-22
Add ServiceUrls

leeavital:


leeavital:
@pnepywoda @ashwinr

qinfchen:
FYI,  I was trying to add this before and we concluded that http-remoting may not be best place for serviceUrls , as there is no internal/external concept on the client side.

markelliot:
Yep, not sure http-clients wants an API of more than Set<String> for URLs.

uschi2000:
What's the motivation for this? (Sorry, I may have missed some conversation)

pnepywoda:
Compass was about to put a ServiceUrl's class in their compass-client and I
pointed out that I didn't think anything more should be added to
compass-client - it should be removed as much as possible (note how we're
also trying to remove our foundry-clients)

markelliot:
thoughts on just adding another project here called 'service-config' and putting it in there? This class is not an http-client concern.

qinfchen:
Would the highline library i am currently working on be the right place for that? since it focuses on app/bundle configuration, i was planning on adding a highline-urls or config for that

markelliot:
sure, is that something you're planning on making an OSS contrib?

qinfchen:
Yes, that is the plan.

leeavital:
@rob Set<String> won't always suffice. For example, we need to configure
urls that are accessible both by internal host and by public DNS.

rob:
Thanks @leeavital

markelliot:
Given that @qinfchen is going to open source this elsewhere, going to close this out.

PR #27, opened 2016-01-27
Hide message and exception type when stack trace is supressed

leeavital:


leeavital:
@jtamayo

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
```diff
@@ -59,7 +60,9 @@ public final Response toResponse(T exception) {
                 error = SerializableError.of(message, exception.getClass(),
                         Arrays.asList(stackTrace));
             } else {
-                error = SerializableError.of(message, exception.getClass());
+                String errorId = UUID.randomUUID().toString();
+                log.error("error has been logged as {}", errorId, exception);
```
uschi2000: think you want to reverse the messages

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
```diff
@@ -59,7 +60,9 @@ public final Response toResponse(T exception) {
...
+                error = SerializableError.of(errorId, RuntimeException.class);
```
jtamayo, line 64: Could we have a SerializableError constructor that doesn't take an exception class? Passing RuntimeException here seems unnecessary and even misleading.

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
markelliot: in particular, I think you want to log something like:

```
log.error("Error {}: {}", errorId, message, exception);
```

leeavital: Yup, nice catch. Mark, is there any reason to log the exception message? Won't it get logged with log.error()?

leeavital:
Also, how do we feel about special-casing on the exception type when setting the status code? Lots of people use `throw BadRequestException` to indicate that a 400 should be returned.

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
markelliot: I think it's worth the clarity to perhaps print it twice/will make grep a LOT easier.

markelliot:
You want to suppress always for 400s but not for others?

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
uschi2000: yes, the message should get logged with `log.error("foo", e)`, but a simple test will tell you. also note that as it stands you log the thing twice, cf. line 55.

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
uschi2000, line 64: you _want to_ pass an exception class since it will get thrown on the client side.

leeavital, line 64: ^ yup. Although I think it may make more sense to default to `InternalServerError` since that doesn't leak much information but does indicate that there was a server side exception. Thoughts?

markelliot, line 64: `RuntimeException` seems fine, I think. It's not necessarily an ISE (we're not throwing a 500).

leeavital:
I'm proposing that any exception which is a subclass of `WebApplicationException` uses the error code associated with the exception (and not the one passed in the constructor). This allows us to use the pattern of throwing WAEs to control the error codes our APIs return.

If we do that, then it feels natural to always serialize exceptions as subclasses of WAE if possible and fall back to ISE when the exception is something else. ISE seems okay to me, because any exception we throw that isn't a WAE is unexpected, and a problem with the server.

markelliot:
Is it reasonable to assume that ISE exists on the client?

leeavital:
It's a member of `javax.ws.rs` so I would say so.

leeavital:
@jtamayo brang up a good point -- ISE uses static methods off `RuntimeDelegate` to create response objects. This requires either jersey or some other JAX-RS implementation on your classpath -- some clients might have trouble deserializing an ISE.

markelliot:
So let's stick with `RuntimeException`, I think it's fine.

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
markelliot: I'd still like to clean up the quality of this error text, how about:

```
log.error("Suppressed Stacktrace for Error: {}", errorId, exception)
```

PR #29, opened 2016-01-28
Unpack parameterized types in OptionalAwareDecoder

leeavital:
This fixes an issue where a `TextDelegateDecoder` wrapped with an `OptionalAwareDecoder` fails to deserialize `Optional<String>`.

### DIFF
--- a/http-clients/src/main/java/feign/OptionalAwareDecoder.java
+++ b/http-clients/src/main/java/feign/OptionalAwareDecoder.java
@@ -16,9 +16,12 @@
 
 package feign;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.base.Optional;
 import feign.codec.Decoder;
 import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 
 /**
@@ -35,11 +38,21 @@ public final class OptionalAwareDecoder implements Decoder {
 
     @Override
     public Object decode(Response response, Type type) throws IOException, FeignException {
-        if (response.status() == 204 && Types.getRawType(type).equals(Optional.class)) {
-            return Optional.absent();
+        if (Types.getRawType(type).equals(Optional.class)) {
+            if (response.status() == 204) {
+                return Optional.absent();
+            } else {
+                Object decoded = checkNotNull(delegate.decode(response, getInnerType(type)),
+                        "Unexpected null content for response status %d", response.status());
+                return Optional.of(decoded);
+            }
         } else {
             return delegate.decode(response, type);
         }
     }
-}
 
+    private static Type getInnerType(Type type) {
+        ParameterizedType paramType = (ParameterizedType) type;
+        return paramType.getActualTypeArguments()[0];
+    }
+}
--- a/http-clients/src/test/java/feign/TextDelegateDecoderTest.java
+++ b/http-clients/src/test/java/feign/TextDelegateDecoderTest.java
@@ -33,7 +33,6 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.common.net.HttpHeaders;
 import com.palantir.remoting.http.FeignClients;
-import feign.codec.DecodeException;
 import feign.codec.Decoder;
 import io.dropwizard.Configuration;
 import io.dropwizard.testing.junit.DropwizardAppRule;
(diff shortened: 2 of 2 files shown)


http-clients/src/main/java/feign/OptionalAwareDecoder.java
```diff
@@ -38,8 +39,26 @@ public Object decode(Response response, Type type) throws IOException, FeignExce
         if (response.status() == 204 && Types.getRawType(type).equals(Optional.class)) {
             return Optional.absent();
         } else {
+            System.out.println("here I am");
```
leeavital: whoops

http-clients/src/main/java/feign/OptionalAwareDecoder.java
```diff
@@ -38,8 +39,26 @@ public Object decode(Response response, Type type) throws IOException, FeignExce
...
+            // try and decode the containing type first
+            if (isOptional(type)) {
```
uschi2000: how about:

```
if (Types.getRawType(type).equals(Optional.class)) {
  if (response.status() == 204) {
    return Optional.absent();
  } else {
    return Optional.of(delegate.decode(response, getInnerType(type)));
  }
} else {
  return delegate.decode(response, type);
}
```

leeavital:
Sorry about the pre-mature squash

http-clients/src/main/java/feign/OptionalAwareDecoder.java
```diff
@@ -35,11 +36,19 @@ public OptionalAwareDecoder(Decoder delegate) {
 
     @Override
     public Object decode(Response response, Type type) throws IOException, FeignException {
-        if (response.status() == 204 && Types.getRawType(type).equals(Optional.class)) {
-            return Optional.absent();
+        if (Types.getRawType(type).equals(Optional.class)) {
+            if (response.status() == 204) {
+                return Optional.absent();
+            } else {
+                return Optional.of(delegate.decode(response, getInnerType(type)));
```
markelliot: is `delegate.decode` guaranteed to return non-null?

leeavital: I would say if `delegate.decode` returns null, the response code should have been a 204.

Should I change line 43 to `ofNullable`?

markelliot: Throwing probably makes sense, it's basically a programming error. Problem is the NPE is going to be a pain to debug. How about this?

```
Optional.of(checkNotNull(delegate.decode(response, getInnerType(type)),
     "Unexpected null content for response status %d", response.status()));
```

markelliot: - should add test coverage to enforce the contract

leeavital: It looks like any resource that returns null from jersey will be a 204, so there's no way to write this test without mocking -- which doesn't seem useful to me.

http-clients/src/main/java/feign/OptionalAwareDecoder.java
```diff
@@ -35,11 +38,20 @@ public OptionalAwareDecoder(Decoder delegate) {
...
     public Object decode(Response response, Type type) throws IOException, FeignException {
-        if (response.status() == 204 && Types.getRawType(type).equals(Optional.class)) {
-            return Optional.absent();
+        if (Types.getRawType(type).equals(Optional.class)) {
+            if (response.status() == 204) {
+                return Optional.absent();
+            } else {
+                Object decoded = checkNotNull(delegate.decode(response, getInnerType(type)),
+                        "Unexpected null content for response status %d", response.status());
+                return Optional.of(decoded);            }
```
markelliot: you have a weird line break thing here

http-clients/src/main/java/feign/OptionalAwareDecoder.java
markelliot: ok, can skip the test and/or forget the null check

http-clients/src/main/java/feign/OptionalAwareDecoder.java
leeavital: yeah, I think my eclipse is set up is weird.  It should be fixed now.

leeavital:
Squashed. If no one else has comments, this should be ready to merge.

Issue #22, opened 2016-01-21
2-way SSL support


markelliot:
Modify trust-stores to have a configurable key store for use with 2-way SSL and add http-clients tests to verify that this is sufficient for 2-way SSL to function.

nmiyake:
Fixed by #35 -- should be able to close this now

PR #31, opened 2016-02-05
Add ability to specify keystore for SSLSocketFactory

nmiyake:
Motivation is to add the ability to leverage this library when making TLS connections that require client certificates. In these situations, both a key store and a trust store need to be specified separately. Verified this with Axiom -- can't establish a connection with custom keys/trust before this change, but can do so after.

This isn't ready to merge as-is, but wanted to submit for feedback before I work on it any further.

My own thoughts:
- Naming of `TrustStoreConfiguration` makes it slightly awkward to use for key store as well, but this code would be exactly the same, so don't think that duplicating it wholesale as `KeyStoreConfiguration` makes much sense.
  - Could rename to something like `StoreConfiguration`, but would be a backcompat-break
- Similar for entire package/project -- named `trust-stores` and `trust`.
- Currently only added one more factory method with 2 optionals. Mirroring all of the methods for trust would make it too verbose.
- Could over-haul the approach completely and write `TrustStores` as a builder instead (where trust and key config can be set as parameters) to avoid factory method combinatorics issues, but that would be a major change so didn't want to do it without checking first.


markelliot:
Awesome, thanks. Have this tagged as #22 :)

In terms of impl: I think it might make more sense for TrustStoreConfiguration to collect the keyStore, path and password and for there still to be a single factory. 

The class and project name probably ought to change, though Java annoyingly calls these things "key stores", and per your comments, not really sure what a better name would be. This whole thing is in service of making an SSLSocketFactory, so maybe we can go with "ssl-config" and "SslConfiguration"?

uschi2000:
+1 for single factory

since we will rely on this to establish a 2-way trusted channel, it would be imported to have integration tests verifying that connections succeed/fail depending on the configured trust setup

nmiyake:
OK great, thanks for the feedback! I'll change the config to collect everything, give a first stab at naming, add some tests and put back up.

nmiyake:
I think I messed this up by force-pushing on my branch instead of adding a commit. Closing and re-opening.

PR #33, opened 2016-02-06
SerializableErrorErrorDecoder wraps remote exception in local excepti…

uschi2000:
…on in order to produce local stacktrace

@pnepywoda @markelliot

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
@@ -59,12 +59,27 @@ public enum SerializableErrorErrorDecoder implements ErrorDecoder {
                 try {
                     SerializableError error = MAPPER.readValue(bodyAsString, SerializableError.class);
                     Class<? extends Exception> exceptionClass = error.getExceptionClass();
-                    Exception exception = exceptionClass.getConstructor(String.class).newInstance(error.getMessage());
+
+                    // Construct remote exception
+                    Exception remoteException =
+                            exceptionClass.getConstructor(String.class).newInstance(error.getMessage());
                     List<StackTraceElement> stackTrace = error.getStackTrace();
                     if (stackTrace != null) {
-                        exception.setStackTrace(stackTrace.toArray(new StackTraceElement[stackTrace.size()]));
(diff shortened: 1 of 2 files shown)


http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -59,12 +60,27 @@ public Exception decode(String methodKey, Response response) {
...
+
+                    // Construct local exception that wraps the remote exception
+                    Constructor<? extends Exception> constructor;
+                    try {
+                        constructor = exceptionClass.getConstructor(String.class, Throwable.class);
+                    } catch (NoSuchMethodException e) {
+                        constructor = exceptionClass.getConstructor(String.class);
                     }
-                    return exception;
+                    Exception localException = constructor.newInstance(error.getMessage(), remoteException);
```
pnepywoda: doesn't this break on exception classes that only have String and not (String, Throwable) ?

uschi2000: ah shoot, i thought of that case and then implemented it wrong ... (see try/catch above)

uschi2000: try again

pnepywoda:
Seems good. Do we care about exceptions that only have default constructors?

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -59,12 +59,27 @@ public Exception decode(String methodKey, Response response) {
                 try {
                     SerializableError error = MAPPER.readValue(bodyAsString, SerializableError.class);
                     Class<? extends Exception> exceptionClass = error.getExceptionClass();
-                    Exception exception = exceptionClass.getConstructor(String.class).newInstance(error.getMessage());
+
+                    // Construct remote exception
+                    Exception remoteException =
+                            exceptionClass.getConstructor(String.class).newInstance(error.getMessage());
```
markelliot: we probably ought to do this construction and the one on line 74 in a utility method that safely tries to construct each of these and failing that just produces a RuntimeException holding data and shoves the classname of the uninstantiable exception into the error message.

uschi2000:
@pnepywoda

PR #40, opened 2016-02-09
Disallow PathParam Optionals

markelliot:


markelliot:
@uschi2000

pnepywoda:
+!

punya:
Even though http-remoting is written in Java 7, many consumers will use it in Java 8 servers. Therefore it seems desirable to block the use of Java 8 Optional as well - it should be possible to do this by looking up the class reflectively by name.

pnepywoda:
People aren't using this lib expecting java8 optional support.

PR #41, opened 2016-02-09
Allow endpoints to return InputStreams

j-baker:
At present, Jackson attempts to decode application/octet-stream bodies,
even though they are not JSON. Instead, decode application/octet-stream
responses with return type byte[] to a byte array, and
application/octet-stream requests with body type byte[] as a byte array
also.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -19,6 +19,8 @@ package com.palantir.remoting.http;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Optional;
 import com.palantir.remoting.http.errors.SerializableErrorErrorDecoder;
+import feign.InputStreamDelegateDecoder;
+import feign.InputStreamDelegateEncoder;
 import feign.OptionalAwareDecoder;
 import feign.TextDelegateDecoder;
 import feign.jackson.JacksonDecoder;
@@ -41,8 +43,11 @@ public final class FeignClients {
     public static FeignClientFactory standard() {
         return FeignClientFactory.of(
                 new GuavaOptionalAwareContract(new JAXRSContract()),
-                new JacksonEncoder(ObjectMappers.guavaJdk7()),
-                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(ObjectMappers.guavaJdk7()))),
+                new InputStreamDelegateEncoder(new JacksonEncoder(ObjectMappers.guavaJdk7())),
+                new OptionalAwareDecoder(
+                        new InputStreamDelegateDecoder(
+                                new TextDelegateDecoder(
+                                        new JacksonDecoder(ObjectMappers.guavaJdk7())))),
                 SerializableErrorErrorDecoder.INSTANCE,
                 FeignClientFactory.okHttpClient());
     }
@@ -53,8 +58,11 @@ public final class FeignClients {
     public static FeignClientFactory vanilla() {
         return FeignClientFactory.of(
                 new GuavaOptionalAwareContract(new JAXRSContract()),
-                new JacksonEncoder(ObjectMappers.vanilla()),
-                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(ObjectMappers.vanilla()))),
+                new InputStreamDelegateEncoder(new JacksonEncoder(ObjectMappers.vanilla())),
+                new OptionalAwareDecoder(
+                        new InputStreamDelegateDecoder(
+                                new TextDelegateDecoder(
+                                        new JacksonDecoder(ObjectMappers.vanilla())))),
                 SerializableErrorErrorDecoder.INSTANCE,
                 FeignClientFactory.okHttpClient());
     }
@@ -65,8 +73,11 @@ public final class FeignClients {
     public static FeignClientFactory withMapper(ObjectMapper mapper) {
         return FeignClientFactory.of(
                 new GuavaOptionalAwareContract(new JAXRSContract()),
-                new JacksonEncoder(mapper),
-                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(mapper))),
+                new InputStreamDelegateEncoder(new JacksonEncoder(mapper)),
+                new OptionalAwareDecoder(
+                        new InputStreamDelegateDecoder(
+                                new TextDelegateDecoder(
+                                        new JacksonDecoder(mapper)))),
                 SerializableErrorErrorDecoder.INSTANCE,
                 FeignClientFactory.okHttpClient());
     }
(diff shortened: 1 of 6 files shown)


http-clients/src/main/java/feign/ByteArrayDelegateDecoder.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public Object decode(Response response, Type type) throws IOException, DecodeException, FeignException {
+        Collection<String> contentTypes = response.headers().get(HttpHeaders.CONTENT_TYPE);
+        if (contentTypes == null) {
+            contentTypes = ImmutableSet.of();
+        }
+        // In the case of multiple content types, or an unknown content type, we'll use the delegate instead.
+        if (contentTypes.size() == 1 && type.equals(byte[].class)
+                && Iterables.getOnlyElement(contentTypes, "contentTypes").equals(MediaType.APPLICATION_OCTET_STREAM)) {
```
markelliot: thinking this through, I think we should maybe skip this filter -- if your class expects a byte[], we should hand it the byte[]. Also wondering if we should special case handling of `InputStream`, so that if your class expects an `InputStream` we give it the body `InputStream`?

http-clients/src/main/java/feign/InputStreamDelegateDecoder.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public InputStreamDelegateDecoder(Decoder delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Object decode(Response response, Type type) throws IOException, FeignException {
+        if (type.equals(InputStream.class)) {
+            byte[] body = Util.toByteArray(response.body().asInputStream());
+            return new ByteArrayInputStream(body);
```
markelliot, line 39: why not `return response.body().asInputStream()`?

http-clients/src/main/java/feign/InputStreamDelegateEncoder.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public InputStreamDelegateEncoder(Encoder delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public void encode(Object object, Type bodyType, RequestTemplate template) throws EncodeException {
+        if (bodyType.equals(InputStream.class)) {
+            try {
+                template.body(Util.toByteArray((InputStream) object), Charsets.UTF_8);
```
markelliot, line 40: surprised that there's no symmetric injecting of an InputStream here

markelliot, line 40: also, does this close the InputStream after it's complete? Either way, class javadoc should probably indicate behavior.

markelliot:
generally fine except confused why we still need to produce `byte[]`s

http-clients/src/main/java/feign/InputStreamDelegateDecoder.java
markelliot, line 39: talked offline: Feign annoyingly closes the body's inputstream

http-clients/src/main/java/feign/InputStreamDelegateEncoder.java
markelliot, line 40: talked offline: requestemplate actually stores a `byte[]`

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -41,8 +43,11 @@ private FeignClients() {}
     public static FeignClientFactory standard() {
         return FeignClientFactory.of(
                 new GuavaOptionalAwareContract(new JAXRSContract()),
-                new JacksonEncoder(ObjectMappers.guavaJdk7()),
-                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(ObjectMappers.guavaJdk7()))),
+                new InputStreamDelegateEncoder(new JacksonEncoder(ObjectMappers.guavaJdk7())),
+                new OptionalAwareDecoder(
```
uschi2000, line 47: I wonder if delegation is really the right thing here?

markelliot, line 47: what's the alternative? if it finds an InputStream it processes, otherwise it delegates (this matches the rest of our stuff?)

http-clients/src/test/java/feign/TestServer.java
```diff
@@ -166,6 +185,18 @@ public String getString(@Nullable String value) {
         String getString(@QueryParam("value") @Nullable String value);
 
         @GET
+        @Path("/writeInputStream")
+        @Consumes(MediaType.TEXT_PLAIN)
+        @Produces(MediaType.TEXT_PLAIN)
+        InputStream writeInputStream(@QueryParam("value") @Nullable String bytes);
```
uschi2000, line 191: why isn't this an OutputStream?

uschi2000, line 191: ah, nevermind

http-clients/src/test/java/feign/InputStreamDelegateEncoderTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        verify(delegate).encode(data, String.class, requestTemplate);
+    }
+
+    @Test
+    public void testStandardClientsUseByteArrayDelegateEncoder() {
+        String data = "bytes";
+        assertThat(service.readInputStream(new ByteArrayInputStream(bytes(data))), is(data));
+    }
+
```
uschi2000, line 86: would like to see what happens for an empty input stream. do you get the empty string back? or null? or something else?

http-clients/src/test/java/feign/InputStreamDelegateDecoderTest.java
```diff
@@ -0,0 +1,93 @@
...
+        String decodedObject = (String) inputStreamDelegateDecoder.decode(response, String.class);
+        assertEquals(returned, decodedObject);
+    }
+
+    @Test
+    public void testStandardClientsUseInputStreamDelegateDecoder() throws IOException {
+        String data = "bytes";
+        assertThat(Util.toByteArray(service.writeInputStream(data)), is(bytes(data)));
+    }
+
```
uschi2000, line 89: symmetric to the encoder test: what happens if data is empty or null?

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
uschi2000, line 47: many of the decoders/encoders are orthogonal in the sense that exactly one is going to consume and return without delegating any further. in terms of patterns, this is really more chain-of-responsibility than delegate. for the latter you would expect the delegation to walk the entire chain.

markelliot, line 47: Open to refactoring here. Also wondering if we still need an OptionalAwareDecoder given that we just fixed the Contract stuff.

I suppose one pitch would be we make:

```
 interface LimitedApplicationDecoder extends Decoder {
      boolean isSuitable(Response response, Type type)
 }
```

And then we can make a Decoder that takes an ordered list of Decoders to try?

uschi2000, line 47: yeah, something like that. i had pitched essentially the same design to the feign folks a while ago, but it was too flexible for their liking: https://github.com/Netflix/feign/pull/287/files

haven't looked at the OptionalAwareDecoder/Contract interaction in detail, but yeah, remove away :)

PR #42, opened 2016-02-09
Fix Checkstyle and FindBugs settings

nmiyake:
Update configuration to ignore generated code

### DIFF
--- a/.baseline/findbugs/excludeFilter.xml
+++ b/.baseline/findbugs/excludeFilter.xml
@@ -1,5 +1,9 @@
 <!-- See http://findbugs.sourceforge.net/manual/filter.html for syntax -->
 <FindBugsFilter>
+    <!-- ignore generated code -->
+    <Match>
+        <Class name="~.*\.*Immutable.*" />
+    </Match>
 
     <!-- @NonnullByDefault doesn't play nicely with overriding @Nullable parameters/outputs -->
     <Match>


nmiyake:
Settings are based on those used in the Foundry projects

.baseline/checkstyle/checkstyle-suppressions.xml
```diff
@@ -7,6 +7,10 @@
 <!-- IMPORTANT ECLIPSE NOTE: If you change this file, you must restart Eclipse
  for your changes to take effect in its Checkstyle integration. -->
 <suppressions>
+    <!-- Eclipse/Idea APT generated code -->
+    <suppress files="[/\\]generated_src[/\\]" checks="." />
+    <suppress files="[/\\]src[/\\]generated[/\\]java[/\\]" checks="." />
```
uschi2000: IDEA uses generated_src (at least with the processors plugin), and so does Eclipse (I believe).

nmiyake: Sorry, don't understand the feedback -- doesn't `[/\\]generated_src[/\\]"` cover that case? FWIW, these lines were copied directly from the ".baseline" in Foundry, and in Eclipse I confirmed that Checkstyle warnings show up for these files before this change, but don't after the change.

uschi2000: I believe the second case (src/generated) may be redundant. (And it is also redundant in Foundry)

nmiyake: OK cool, will verify and remove if so.

nmiyake:
Updated PR; ready for review

.baseline/findbugs/excludeFilter.xml
```diff
@@ -1,5 +1,9 @@
 <!-- See http://findbugs.sourceforge.net/manual/filter.html for syntax -->
 <FindBugsFilter>
+    <!-- ignore generated code -->
+    <Match>
+        <Class name="~.*\.*Immutable.*" />
```
j-baker, line 5: Why is FindBugs running on generated_src? Are you using an outdated version of processors? (to be clear, this is a gradle-processors bug if anything).

https://github.com/palantir/gradle-processors/blob/master/src/main/groovy/org/inferred/gradle/ProcessorsPlugin.groovy#L143

.baseline/checkstyle/checkstyle-suppressions.xml
```diff
@@ -7,6 +7,9 @@
 <!-- IMPORTANT ECLIPSE NOTE: If you change this file, you must restart Eclipse
  for your changes to take effect in its Checkstyle integration. -->
 <suppressions>
+    <!-- Eclipse/Idea APT generated code -->
+    <suppress files="[/\\]generated_src[/\\]" checks="." />
```
j-baker: again, this sounds like something gradle-processors should be doing/does, not baseline/remoting. The only reason it's generated_src is because gradle-processors defines it to be so. If you use idea, you should also be excluding generated_testSrc.

.baseline/findbugs/excludeFilter.xml
markelliot, line 5: this project does have an outdated processors plugin, we should bump the version to 0.2.4 (latest)

.baseline/checkstyle/checkstyle-suppressions.xml
qinfchen: agree, i believe eclipse default generated folder is `apt_generated`.

nmiyake: If you check out the repo, run "./gradlew eclipse" and then load the generated projects into Eclipse, the generated Immutables files in "generated_src" cause Checkstyle errors and FindBugs warnings in the workspace. Adding these lines makes those errors and warnings go away.

If there's something else that can be done for the generated Eclipse project or other settings that are incorrect on my Eclipse, happy to fix that way as well, but believe strongly that having the default Eclipse projects report errors/warnings is poor form.

.baseline/findbugs/excludeFilter.xml
nmiyake, line 5: I'll try that locally and see if that resolves the issue. Issue I'm seeing right now is that default Eclipse projects generated currently report FindBugs warnings and Checkstyle errors.

nmiyake, line 5: Baseline is on 0.2.2 -- bumping to 0.2.4 locally didn't help.

We're using version 1.1 of `gradle-processors`, and looking at the version history I don't think there are any changes between 1.1 and 1.1.5 (current release) that would fix this issue.

uschi2000:
:+1: this all looks good to me

nmiyake:
Rob's change at #44 addresses the Checkstyle issue, which is the most important one.

Keeping this PR for discussion on whether FindBugs change should go in -- with Rob's change, FindBugs warnings are still shown in Eclipse for the generated files. This is less of an issue than the Checkstyle errors, but still better to fix than not.

uschi2000:
@nmiyake happy to merge the findbugs thing here if you rebase. 
I don't think we can really include this kind of config in the gradle-processors plugin since it cannot know about the specific checkstyle/findbugs config.

nmiyake:
Thanks, will rebase and push back up.

nmiyake:
Pushed rebased commit

PR #44, opened 2016-02-10
Rfink/update plugins

uschi2000:


uschi2000:
New baseline config as per `gradlew updateBaselineConfig`

uschi2000:
@markelliot

markelliot:
:+1:

nmiyake:
+1, verified that this fixes the Checkstyle errors.

I still see FindBugs warnings in the generated files, but those are less important since they aren't errors. Will change my other PR to only include FindBugs change and that can be discussed there.

uschi2000:
rebasing

PR #45, opened 2016-02-15
Decode general web exceptions

### LINKED ISSUE #24: SerializableErrorErrorDecoder.java fails on exceptions without a single-string constructor
BLUF: Failure to parse error body for any HTTP statuses other than a few of the most common

In https://github.com/palantir/http-remoting/blob/develop/http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java, "exceptionClass.getConstructor(String.class)" requires that the exception has a constructor that takes a single String. This is reasonable for the HTTP statuses that have a a specific exception made for them (e.g. BadRequestException, ForbiddenException), but for all the other statuses, you have to use either a ClientErrorException or ServerErrorException, which do not have single-string constructors.

Hacky workarounds are possible: I wrote code in Contour the catches HTTP-remoting exception and searches the message for the right HTTP status code. :/

TheBeruriahIncident:
Previously, exceptions corresponding to specific HTTP error codes (e.g. BadRequestException) would be properly decoded, but more general exceptions that just hold a status value (e.g. ClientErrorException) would fail to decode.

We now handle:
- WebApplicationException
- ServerErrorException
- ClientErrorException

Fixes #24

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
@@ -28,6 +28,10 @@ import java.io.Reader;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.List;
+import javax.annotation.CheckForNull;
+import javax.ws.rs.ClientErrorException;
+import javax.ws.rs.ServerErrorException;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -69,7 +73,8 @@ public enum SerializableErrorErrorDecoder implements ErrorDecoder {
                 }
 
                 // Construct remote exception and fill with remote stacktrace
-                Exception remoteException = constructException(error.getExceptionClass(), error.getMessage());
+                Exception remoteException = constructException(error.getExceptionClass(), error.getMessage(),
+                        response.status(), null);
                 List<StackTraceElement> stackTrace = error.getStackTrace();
                 if (stackTrace != null) {
                     remoteException.setStackTrace(stackTrace.toArray(new StackTraceElement[stackTrace.size()]));
@@ -78,7 +83,8 @@ public enum SerializableErrorErrorDecoder implements ErrorDecoder {
                 // Construct local exception that wraps the remote exception and fill with stack trace of local
                 // call (yet without the reflection overhead).
                 Exception localException =
-                        constructException(error.getExceptionClass(), error.getMessage(), remoteException);
+                        constructException(error.getExceptionClass(), error.getMessage(), response.status(),
+                                remoteException);
                 localException.fillInStackTrace();
 
                 return localException;
@@ -101,18 +107,17 @@ public enum SerializableErrorErrorDecoder implements ErrorDecoder {
         }
     }
 
-    private static Exception constructException(Class<? extends Exception> exceptionClass, String message) {
-        try {
-            return exceptionClass.getConstructor(String.class).newInstance(message);
-        } catch (Exception e1) {
-            return new RuntimeException(String.format(
-                    "Failed to construction exception of type %s, constructing RuntimeException instead: %s%n%s",
-                    exceptionClass.toString(), e1.toString(), message));
+    // wrappedException may be null to indicate an unknown cause
+    private static Exception constructException(Class<? extends Exception> exceptionClass, String message, int status,
+            @CheckForNull Throwable wrappedException) {
+        if (exceptionClass.equals(ClientErrorException.class)) {
+            return new ClientErrorException(message, status, wrappedException);
+        } else if (exceptionClass.equals(ServerErrorException.class)) {
+            return new ServerErrorException(message, status, wrappedException);
+        } else if (exceptionClass.equals(WebApplicationException.class)) {
+            return new WebApplicationException(message, wrappedException, status);
         }
-    }
 
-    private static Exception constructException(
-            Class<? extends Exception> exceptionClass, String message, Throwable wrappedException) {
         // Note: If another constructor is added, then we should refactor the construction logic in order to avoid
         // nested try/catch
         try {
@@ -123,7 +128,7 @@ public enum SerializableErrorErrorDecoder implements ErrorDecoder {
                 return exceptionClass.getConstructor(String.class).newInstance(message);
             } catch (Exception e1) {
                 return new RuntimeException(String.format(
-                        "Failed to wrap exception as %s, wrapping exception as RuntimeException instead: %s%n%s",
+                        "Failed to construct exception as %s, constructing RuntimeException instead: %s%n%s",
                         exceptionClass.toString(), e1.toString(), message),
                         wrappedException);
             }
--- a/http-clients/src/test/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoderTests.java
+++ b/http-clients/src/test/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoderTests.java
@@ -18,6 +18,7 @@ package com.palantir.remoting.http.errors;
 
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -25,25 +26,35 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import feign.Response;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Collections;
 import javax.annotation.CheckForNull;
+import javax.ws.rs.ClientErrorException;
+import javax.ws.rs.ServerErrorException;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response.Status;
 import org.junit.Test;
 
 public final class SerializableErrorErrorDecoderTests {
 
     private static final SerializableErrorErrorDecoder decoder = SerializableErrorErrorDecoder.INSTANCE;
+    private static final String message = "hello";
 
     @Test
-    public void testJsonException() throws JsonProcessingException {
-        Object error = SerializableError.of("msg", IllegalArgumentException.class, null);
-        String json = new ObjectMapper().writeValueAsString(error);
-        Response response = getResponse(MediaType.APPLICATION_JSON, json);
-        Exception decode = decoder.decode("ignored", response);
-        assertThat(decode, is(instanceOf(IllegalArgumentException.class)));
-        assertThat(decode.getMessage(), is("msg"));
+    public void testWebApplicationExceptions() {
+        testEncodingAndDecodingWebException(ClientErrorException.class, Status.NOT_ACCEPTABLE);
+        testEncodingAndDecodingWebException(ServerErrorException.class, Status.BAD_GATEWAY);
+        testEncodingAndDecodingWebException(WebApplicationException.class, Status.NOT_MODIFIED);
+    }
+
+    @Test
+    public void testSpecificException() {
+        Exception exception = encodeAndDecode(new IllegalArgumentException("msg"));
+        assertThat(exception, is(instanceOf(IllegalArgumentException.class)));
+        assertThat(exception.getMessage(), is("msg"));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
...
+                // the message should be of the form "HTTP $STATUS $DESCRIPTION
+                // for example, "HTTP 501 Not Implemented"
+                int prefixLength = "HTTP ".length();
+                String statusAsString = message.substring(prefixLength, prefixLength + 3);
+                int status = Integer.parseInt(statusAsString);
+
+                // WebApplicationException has a different order in its constructor arguments
+                if (exceptionClass.equals(WebApplicationException.class)) {
+                    return exceptionClass.getConstructor(Throwable.class, Integer.TYPE)
+                            .newInstance(wrappedException, status);
```
markelliot: how about:

```
return new WebApplicationException(wrappedException, status);
```

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
         // Note: If another constructor is added, then we should refactor the construction logic in order to avoid
         // nested try/catch
         try {
+            if (exceptionClass.equals(ClientErrorException.class)
```
markelliot: this would also need to appear in the other `constructException` block for it to generally work, which makes me think the test coverage you're committing isn't sufficient.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
...
+                    || exceptionClass.equals(ServerErrorException.class)
+                    || exceptionClass.equals(WebApplicationException.class)) {
+                // the message should be of the form "HTTP $STATUS $DESCRIPTION
+                // for example, "HTTP 501 Not Implemented"
+                int prefixLength = "HTTP ".length();
+                String statusAsString = message.substring(prefixLength, prefixLength + 3);
+                int status = Integer.parseInt(statusAsString);
```
markelliot: I think maybe we should just put the regex in here, don't care about perf in the exception handler and would prefer something more readable

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
markelliot: or do these actually have a single-string arg constructor?

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
         // Note: If another constructor is added, then we should refactor the construction logic in order to avoid
```
uschi2000, line 121: I believe that this comment applies to this change. Introducing more special cases turns this into spaghetti.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
...
         // nested try/catch
         try {
+            if (exceptionClass.equals(ClientErrorException.class)
+                    || exceptionClass.equals(ServerErrorException.class)
+                    || exceptionClass.equals(WebApplicationException.class)) {
+                // the message should be of the form "HTTP $STATUS $DESCRIPTION
+                // for example, "HTTP 501 Not Implemented"
+                int prefixLength = "HTTP ".length();
+                String statusAsString = message.substring(prefixLength, prefixLength + 3);
```
schlosna: This may `throw IndexOutOfBoundsException` if the message doesn't contain "HTTP " or is length less than 5. Could also NPE if message is null.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
schlosna: Would also avoid possible problems where statusAsString is non-numeric and won't parse.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -115,6 +118,25 @@ private static Exception constructException(
         // Note: If another constructor is added, then we should refactor the construction logic in order to avoid
         // nested try/catch
         try {
+            if (exceptionClass.equals(ClientErrorException.class)
+                    || exceptionClass.equals(ServerErrorException.class)
+                    || exceptionClass.equals(WebApplicationException.class)) {
```
schlosna: Why not do `WebApplicationException.class.isAssignableFrom(exceptionClass)` to cover all of these cases or do you want to explicitly exclude `javax.ws.rs.RedirectionException`?

TheBeruriahIncident: We want to exclude all of the Exception classes that were already handled. `BadRequestException`, `NotFoundException`, etc all have single-string constructors, but are children of WebApplicationException.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
TheBeruriahIncident: Ah, the other `constructException` was returning `RuntimeException`, will fix!

TheBeruriahIncident:
@markelliot , I made an oversight in that you can construct these exceptions in such a way that the message itself won't have the status, meaning that the status only lives as a field in the `Response`. As such, the `SerializableException` doesn't necessarily have enough info to actually reconstruct the exception. Is it reasonable to add a `status` field to `SerializableException`?

TheBeruriahIncident:
@markelliot Are you good with adding `status` to `SerializableException`?

uschi2000:
@akroy why do you need it in the exception if it's already in the response?

TheBeruriahIncident:
@uschi2000 We don't currently put the status in the response. We've special cased a few exception types, but most everything just gets a generic 500 as the response status.

(I was planning on following this up with making an exception mapper that actually uses statuses) My (guessed) interpretation was that we were intentionally only using the `SerializableException` info for serialization, but if encoding the status in the response is fine, I'll work on our exception mappers first.

markelliot:
The response code is available from the literal http response, Adam, we shouldn't be additionally serializing it into the data structure.

TheBeruriahIncident:
Great. I'll close this out and add exception mapping that puts the status into the http response, and I'll reopen after that's fixed.

markelliot:
the status is already placed into the HTTP response.

TheBeruriahIncident:
Specific statuses are placed there for `FORBIDDEN` and `NOT_FOUND`, and we have handling for `IllegalArgumentException`, but from what I can tell/observe, most exceptions are handled by the `RuntimeExceptionMapper`, which just uses 500 no matter what.

markelliot:
- @pnepywoda

This doesn't quite jive with my recollection. I'm pretty sure if you return a WebAppException we pull out the error code.

TheBeruriahIncident:
I don't see any relevant mapper in: https://github.com/palantir/http-remoting/tree/develop/http-servers/src/main/java/com/palantir/remoting/http/server . I did write the (fairly small) PR to fix that though, I'll put it up in a sec

markelliot:
Ok, happy to look at the pr.

TheBeruriahIncident:
PR at https://github.com/palantir/http-remoting/pull/48

TheBeruriahIncident:
@markelliot @uschi2000 @schlosna Redid this now that we can assume the status is right

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -112,7 +125,15 @@ private static Exception constructException(Class<? extends Exception> exception
...
     private static Exception constructException(
-            Class<? extends Exception> exceptionClass, String message, Throwable wrappedException) {
+            Class<? extends Exception> exceptionClass, String message, Throwable wrappedException, int status) {
+        if (exceptionClass.equals(ClientErrorException.class)) {
+            return new ClientErrorException(status, wrappedException);
+        } else if (exceptionClass.equals(ServerErrorException.class)) {
+            return new ServerErrorException(status, wrappedException);
+        } else if (exceptionClass.equals(WebApplicationException.class)) {
+            return new WebApplicationException(wrappedException, status);
+        }
```
TheBeruriahIncident, line 119: I'm not enthused about having this similar block twice, but since in one case we wrap the exception, more DRY approaches seem much less clean.

Similarly, I'm not a fan of having two different `if`s for Client and Server, but that lets us remove the reflection and take this out of the try-catch block.

uschi2000, line 119: you can `initCause()`; it's "clean enough" since it enforces (at runtime) that it's called at most once.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -101,7 +106,15 @@ public Exception decode(String methodKey, Response response) {
         }
     }
 
-    private static Exception constructException(Class<? extends Exception> exceptionClass, String message) {
+    private static Exception constructException(Class<? extends Exception> exceptionClass, String message, int status) {
+        if (exceptionClass.equals(ClientErrorException.class)) {
```
uschi2000, line 113: what about the message?

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
TheBeruriahIncident, line 119: Just tried that: apparently all the WebApplicationExceptions call into a single constructor that takes cause, meaning that they're all setting cause to `null` if you don't set it. This prevents us from setting it with `initCause`, because of the "set once" limitation.

TheBeruriahIncident:
@uschi2000 Using message, improved tests

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
uschi2000, line 121: Any opinion here?

TheBeruriahIncident, line 121: That comment seems to be about the nested try-catch and a worry about nesting further. My change doesn't touch any of that, so I don't think it makes that situation worse. It may be possible to clean that block up, but I think that's a separate issue.

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -101,7 +106,15 @@ public Exception decode(String methodKey, Response response) {
         }
     }
 
-    private static Exception constructException(Class<? extends Exception> exceptionClass, String message) {
+    private static Exception constructException(Class<? extends Exception> exceptionClass, String message, int status) {
```
markelliot: looking at http://grepcode.com/file/repo1.maven.org/maven2/javax.ws.rs/javax.ws.rs-api/2.0.1/javax/ws/rs/WebApplicationException.java#65 I think this method ought to simply invoke:

```
return constructException(exceptionClass, message, null, status);
```

...and that that would be a lot less duplicate code.

TheBeruriahIncident: Collapsed the two methods into one

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -101,18 +106,16 @@ public Exception decode(String methodKey, Response response) {
...
     }
 
-    private static Exception constructException(Class<? extends Exception> exceptionClass, String message) {
-        try {
-            return exceptionClass.getConstructor(String.class).newInstance(message);
-        } catch (Exception e1) {
-            return new RuntimeException(String.format(
-                    "Failed to construction exception of type %s, constructing RuntimeException instead: %s%n%s",
-                    exceptionClass.toString(), e1.toString(), message));
+    private static Exception constructException(
```
jlzhao27: I don't think this change is correct. The problem I noticed is that the server side `JsonExceptionMapper` is capturing the web specific exceptions like 405 Not Allowed and throwing them as a 500. I think the fix needs to be in the exception mapper to throw the right status code (and we should also decode the exception).

TheBeruriahIncident: Hey Jason! HTTP-remoting's handling of exceptions was actually doubly broken:
- most exceptions were thrown as 500s on server side (what you're mentioning)
- most exceptions were failing to decode on client side (what this commit is)

Both of these errors were related to the same failure to deal with WebApplication, ClientError, or ServerError exceptions, just in different places. I already fixed the problem you mention at https://github.com/palantir/http-remoting/pull/48 .

jlzhao27: oh doh, thanks adam! neither of these changes are released yet right?

uschi2000: nope. we can release 0.7.0 today. it's a major breaking change, both in
terms of code and config.

TheBeruriahIncident: Not yet! Server side fix is merged (so you could snapshot). On https://github.com/palantir/http-remoting/pull/35, @uschi2000 said that we'll likely release 0.7.0 once my fixes go in.

TheBeruriahIncident:
@markelliot @uschi2000 How does this look?

jlzhao27:
ping on this also, GK is waiting on the 0.7.0 release :)

markelliot:
seems okay to me

TheBeruriahIncident:
Sweet. I'm going to squash and rebase

TheBeruriahIncident:
@markelliot Ready!

http-clients/src/main/java/com/palantir/remoting/http/errors/SerializableErrorErrorDecoder.java
```diff
@@ -101,18 +106,16 @@ public Exception decode(String methodKey, Response response) {
...
+            Class<? extends Exception> exceptionClass, String message, int status, Throwable wrappedException) {
```
uschi2000: @CheckForNull missing

uschi2000: or probably `Optional<Throwable>`

uschi2000: and please leave a comment saying that sub-class of RuntimeException can take a `null` cause parameter to indicate an unknown cause.

uschi2000:
one last comment above, hopefully.

PR #48, opened 2016-02-16
Make ExceptionMappers use proper HTTP status

TheBeruriahIncident:
Previously, any non-standard exceptions (most `WebApplicationException`s) would land in the `RuntimeExceptionMapper`, thus getting a status of 500, regardless of actual status.

The way I solved this is by having each mapper have a `getStatus` method, allowing the mapper to check the status at runtime, instead of hardcoding a status in the constructor. Now a single `WebApplicationExceptionMapper` can handle all `WebApplicationException`s.


http-servers/src/main/java/com/palantir/remoting/http/server/ExceptionMappers.java
```diff
@@ -29,10 +29,9 @@ private ExceptionMappers() {}
 
     public static void visitExceptionMappers(boolean includeStackTrace,
             Consumer<ExceptionMapper<? extends Throwable>> consumer) {
-        consumer.accept(new ForbiddenExceptionMapper(includeStackTrace));
         consumer.accept(new IllegalArgumentExceptionMapper(includeStackTrace));
         consumer.accept(new NoContentExceptionMapper());
-        consumer.accept(new NotFoundExceptionMapper(includeStackTrace));
         consumer.accept(new RuntimeExceptionMapper(includeStackTrace));
+        consumer.accept(new WebApplicationExceptionMapper(includeStackTrace));
```
TheBeruriahIncident, line 35: I was nervous about the ordering here (i.e. I didn't want `RuntimeException`s to be handled first), but this seems to work. I guess something clever must happen with finding the most specific matching `ExceptionMapper`.

http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
```diff
@@ -76,6 +75,8 @@ public final Response toResponse(T exception) {
         return builder.build();
     }
 
+    abstract StatusType getStatus(T exception);
```
TheBeruriahIncident: This is `StatusType` rather than just `Status`, because `Response` only gives you back `StatusType`, which is a superclass of `Status`

markelliot: Ok. This should be protected rather than package.

markelliot: Or wait, it has to be visible...actually make it public.

markelliot: No I take it back. Protected.

pnepywoda:
looks good. Want to add a test for an exception that works now (that didn't before)?

TheBeruriahIncident:
Cool, amending to fix visibility and add test!

http-servers/src/main/java/com/palantir/remoting/http/server/ExceptionMappers.java
uschi2000, line 35: can you add a test to that effect?

TheBeruriahIncident:
@pnepywoda @uschi2000 added tests
@markelliot fixed visibility

TheBeruriahIncident:
@uschi2000 Fleshed out the tests to include both levels of subclasses and fixed naming

http-servers/src/main/java/com/palantir/remoting/http/server/NotFoundExceptionMapper.java
```diff
@@ -1,30 +0,0 @@
...
- */
-
-package com.palantir.remoting.http.server;
-
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.core.Response.Status;
-import javax.ws.rs.ext.Provider;
-
-@Provider
-public final class NotFoundExceptionMapper extends JsonExceptionMapper<NotFoundException> {
```
markelliot, line 24: just to be sure, do NotFoundExceptions actually encode `Status.NOT_FOUND` already?

markelliot, line 24: (we should test this)

markelliot, line 24: ....and you do, so nvm

TheBeruriahIncident, line 24: Yeah, the different specific web exceptions bake their `Status` in. (I can add a test for `ForbiddenException` if you want to be super certain that we're covering all the same things as before, but I feel like we don't want the mess of confirming that every specific exception actually encodes what it claims to encode)

markelliot, line 24: I think we want to guarantee behavior of what we had before.

TheBeruriahIncident, line 24: Sure, I'll add the other test just to be sure there's no regression.

I do think though that we don't want to follow this logic to the conclusion and eventually test that BadRequestExceptions have BAD_REQUEST status, NotAuthorizedExceptions have NOT_AUTHORIZED status, and so on and so on. It seems like at some point we need to trust that Java's WS library is sane.

TheBeruriahIncident, line 24: Test added! :D

markelliot, line 24: I'm fine not following it for all exceptions, but I'd like to make sure we don't change any of the semantics we already had.

TheBeruriahIncident, line 24: +1

TheBeruriahIncident:
Is this good to squash/rebase/merge?

uschi2000:
good to go from my perspective

uschi2000:
(squash and rebase, yes)

TheBeruriahIncident:
@uschi2000 Ready!

Issue #50, opened 2016-02-16
testSslNoClientAuthenticationFailsWithMultipleKeyStoreUnspecified() test is flaky


nmiyake:
This test makes an assumption about the order in which a key store returns the keys that it contains. Should rewrite or remove this test since it looks like this can be system-dependent.

nmiyake:
Fixed by #51

PR #51, opened 2016-02-16
Remove unit test that was system-dependent

nmiyake:
Fixes issue #50. The test in question assumed
that a JKS key store would return its entries
in a deterministic order when an alias was not
specified, but this may not be true. Removing
the test entirely since there is already a test
that verifies that the functionality is correct
when an alias is explicitly specified.


nmiyake:
@akroy , this should fix the issue (it removes the test entirely).

nmiyake:
@uschi2000 or @markelliot , can you merge this change once it verifies?

As explained in the comment, the test I'm removing shouldn't be necessary -- I already have a test for the positive case. Although I would like it if I could test for the negative case, if the order in which a key is returned is non-deterministic (or not well-defined), then any test I write for it will always have a chance of failure (since there's always a chance that the correct key will be returned first, which would cause the test to fail)

Issue #30, opened 2016-01-29
TrustStoreConfiguration is not serializable


jiahuijiang:
Tried to serialize a TrustStoreConfiguration object and got the following exception:

com.fasterxml.jackson.databind.JsonMappingException: Direct self-reference leading to cycle (through reference chain: com.palantir.trust.Json["trustStorePath"]->sun.nio.fs.UnixPath["parent"]->sun.nio.fs.UnixPath["root"])

Is it meant to be like this?

markelliot:
Had assumed this wasn't something that ever needed to be serialized, what's the use-case?

jiahuijiang:
We are trying to pass some information from one process to another spark-submit process using a file. And the trust store configuration is part of it.

schlosna:
It's failing on the path to the trust store on disk, which likely isn't even a valid path on the workers.
What is being serialized that is referencing the trust store config? Do you actually want to serialize the trust store config in that case?

schlosna:
If you do actually want this, you could use https://github.com/FasterXML/jackson-datatype-jdk7 (specifically see https://github.com/FasterXML/jackson-datatype-jdk7/issues/1 )

jiahuijiang:
The driver will be the only one that's using the trust store. And it will be on the same box as the process that forked this spark-submit process.
Thanks for the link!

markelliot:
Pretty simple to fix this. Will adjust in a bit.

markelliot:
hm, actually, I'd expect this to work, will need to do some testing

jiahuijiang:
@markelliot I followed the link @schlosna sent and registered the jdk7 module to the ObjectMapper and it's working now :) Thanks for taking a look at the problem!

tpetracca:
FYI the change made in #52 doesn't address the issue here (although adding the jdk7module does so I'm just going to do that), because SslConfiguration still uses java.nio.file.Path.

Edit: I now realize this specific issue was TrustStoreConfiguration, which no longer exists.  I think this issue may have regressed when we (what I assume was a refactor) removed TrustStoreConfiguration and KeyStoreConfiguration in favor of SslConfiguration.

PR #52, opened 2016-02-16
support URIs instead of just Paths

pnepywoda:
@uschi2000

### DIFF
--- a/ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
+++ b/ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
@@ -16,13 +16,11 @@
 
 package com.palantir.remoting.ssl;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.base.Optional;
-import java.nio.file.Path;
-import java.nio.file.Paths;
+import com.google.common.base.Preconditions;
+import java.net.URI;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Style.ImplementationVisibility;
 
@@ -33,7 +31,7 @@ import org.immutables.value.Value.Style.ImplementationVisibility;
 public abstract class KeyStoreConfiguration {
 
     @Value.Parameter
-    public abstract Path path();
+    public abstract URI uri();
 
     @Value.Parameter
     public abstract String password();
@@ -49,12 +47,12 @@ public abstract class KeyStoreConfiguration {
 
     @Value.Check
     protected final void check() {
-        checkArgument(!path().equals(Paths.get("")), "path cannot be empty");
-        checkArgument(!password().equals(""), "password cannot be empty");
+        Preconditions.checkArgument(!uri().equals(URI.create("")), "uri cannot be empty");
+        Preconditions.checkArgument(!password().equals(""), "password cannot be empty");
     }
 
-    public static KeyStoreConfiguration of(Path path, String password) {
-        return ImmutableKeyStoreConfiguration.of(path, password);
+    public static KeyStoreConfiguration of(URI uri, String password) {
+        return ImmutableKeyStoreConfiguration.of(uri, password);
     }
 
     public static Builder builder() {
--- a/ssl-config/src/main/java/com/palantir/remoting/ssl/TrustStoreConfiguration.java
+++ b/ssl-config/src/main/java/com/palantir/remoting/ssl/TrustStoreConfiguration.java
@@ -16,12 +16,10 @@
 
 package com.palantir.remoting.ssl;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import java.nio.file.Path;
-import java.nio.file.Paths;
+import com.google.common.base.Preconditions;
+import java.net.URI;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Style.ImplementationVisibility;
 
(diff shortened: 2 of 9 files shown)


ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
```diff
@@ -33,7 +32,7 @@
 public abstract class KeyStoreConfiguration {
 
     @Value.Parameter
-    public abstract Path path();
+    public abstract URI path();
```
uschi2000: how does this play with jackson serialization?
- in case path is "foo/bar"
- in case path is "file://foo/bar"
- in case path is "bogusScheme://foo/bar"

pnepywoda: I'm not sure why these are jackson serializable anyway. Path and URI aren't
exactly to/from jsonifiable

uschi2000: TrustStoreConfiguration/KeyStoreConfiguration are often used as part of Dropwizard Configuration objects, which in turn are read from disk using a YAML/JSON mapper.

pnepywoda:
@uschi2000 dropwizard supports URI parsing. added unit test: https://github.com/palantir/http-remoting/pull/52/files#diff-0a17123000e058a85ac57bf2cbb5ef62

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
uschi2000: To clarify: the intent of this change is that a user can pass in an arbitrary URI and the app can call `uri.toURL().openStream()` or `Paths.get(uri)`, assuming they have a handler registered for the URI scheme passed in. Correct?

uschi2000: should this be `uri` now? (we're breaking compatibility anyway with the next release, so we can chose whatever name works best)

ssl-config/src/main/java/com/palantir/remoting/ssl/TrustStoreConfiguration.java
```diff
@@ -32,7 +29,7 @@
 public abstract class TrustStoreConfiguration {
 
     @Value.Parameter
-    public abstract Path path();
+    public abstract URI path();
```
uschi2000: as above

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
uschi2000: @markelliot

markelliot: yeah it should be URI, and there should be some javadoc here to make it explicit that just entering the path is fine.

ssl-config/src/test/java/com/palantir/ssl/DropwizardConfigTests.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
```
markelliot, line 15: missing license

markelliot:
Can you also test serialization of this?

pnepywoda:
When would we ever be serializing it?

markelliot:
There was an issue where someone was serializing it to ship config to worker nodes...see #30

pnepywoda:
Seems like someone should have written a unit test for THAT issue ;)
Anyway, sure, I'll add it to this commit as well

markelliot:
Well it was verified at the time it was serializable, but since we're going from `Path` to `URI` we should be testing :)

pnepywoda:
will squash after +1

pnepywoda:
@markelliot ready for review again

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
```diff
@@ -49,12 +48,11 @@ public String type() {
 
     @Value.Check
     protected final void check() {
-        checkArgument(!path().equals(Paths.get("")), "path cannot be empty");
```
markelliot, line 52: sorry, missed this first time through, I think we want to error if the URI contains the empty string, can we have a check for that somehow?

markelliot:
lgtm except for losing the emptiness check on the uri

pnepywoda:
Empty string in config actually becomes "file:///" in the URL parser. As a
consumer of this we can't actually check for empty string

markelliot:
hm, is there a way to check that if it's a file it actually exists? It'd be nice to maintain some error reporting at the config level.

pnepywoda:
It's a URI, you can't check for existence

markelliot:
`file:///` is invalid, so what if we error if it's equal to URI("")?

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStoreConfiguration.java
pnepywoda, line 52: ok, checking for empty string specifically. dunno when this will happen but sure, one fail-fast check

pnepywoda:
any more comments?

ssl-config/src/main/java/com/palantir/remoting/ssl/TrustStoreConfiguration.java
```diff
@@ -40,13 +37,8 @@ public String type() {
         return "JKS";
     }
 
-    @Value.Check
-    protected final void check() {
```
uschi2000, line 44: why remove this check here? should be symmetric to KeyStoreConfiguration

uschi2000:
one comment, otherwise LGTM

PR #53, opened 2016-02-17
Add QueryMap class and encoder

nmiyake:
Allow Feign clients to use a QueryMap to send GET
requests that have arbitrary query parameter keys
and values. Analogous to the functionality provided
by the @QueryMap annotation in Retrofit.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
@@ -4,14 +4,10 @@
 
 package com.palantir.remoting.http;
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.collect.Multimap;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Style.ImplementationVisibility;
 
-@JsonDeserialize(as = ImmutableQueryMap.class)
-@JsonSerialize(as = ImmutableQueryMap.class)
 @Value.Immutable
 @Value.Style(visibility = ImplementationVisibility.PACKAGE)
 public abstract class QueryMap {
--- a/http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package com.palantir.remoting.http;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.collect.Multimap;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Style.ImplementationVisibility;
+
+@JsonDeserialize(as = ImmutableQueryMap.class)
+@JsonSerialize(as = ImmutableQueryMap.class)
+@Value.Immutable
+@Value.Style(visibility = ImplementationVisibility.PACKAGE)
+public abstract class QueryMap {
+    public abstract Multimap<String, String> queryMap();
+
+    public static QueryMap of(Multimap<String, String> queryMap) {
+        return ImmutableQueryMap.builder().queryMap(queryMap).build();
+    }
+}
--- a/http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
+++ b/http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package feign;
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMultimap;
+import com.palantir.remoting.http.FeignClientFactory;
+import com.palantir.remoting.http.ObjectMappers;
+import com.palantir.remoting.http.QueryMap;
+import com.palantir.remoting.http.errors.SerializableErrorErrorDecoder;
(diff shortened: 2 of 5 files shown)


nmiyake:
@uschi2000 @markelliot 

Not dead-set on getting this merged as part of this repo (could just have this code in my client project as well), but interested in getting your feedback on the general concept/approach.

Our current Retrofit client uses the `@QueryMap` functionality in Retrofit, and this is an attempt to support the same concept in Feign (approach is to define a new object that's passed to a GET request as a `@Param` object, and a custom encoder reads the map provided by the object and embeds them as query params in the request).

Rob, this implements some of the stuff we talked about. Although this makes the most sense for client code, it does support the server scenario pretty naturally as well (see unit test where there's a strongly typed interface, but there can be a "Client" interface that extends it to provide QueryMap support)

uschi2000:
I see two use-cases for this, correct if I'm wrong:
- a method with a large (yet finite) number of optional (yet statically known) parameters, defined by a server interface with hard-coded parameters
- a method with an unbounded (and not statically known) number of parameters

nmiyake:
Correct -- there may be others, but those are the main cases that come to mind for me as well.

http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
```diff
@@ -0,0 +1,23 @@
...
+import com.google.common.collect.Multimap;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Style.ImplementationVisibility;
+
+@JsonDeserialize(as = ImmutableQueryMap.class)
+@JsonSerialize(as = ImmutableQueryMap.class)
+@Value.Immutable
+@Value.Style(visibility = ImplementationVisibility.PACKAGE)
+public abstract class QueryMap {
+    public abstract Multimap<String, String> queryMap();
```
markelliot, line 14: do Multimaps actually serialize reasonably?

http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Before
+    public void before() {
+        String endpointUri = "http://localhost:" + APP.getLocalPort();
+
+        FeignClientFactory factory = FeignClientFactory.of(
+                new JAXRSContract(),
+                // register query map encoder
+                new QueryMapEncoder(new JacksonEncoder(ObjectMappers.guavaJdk7())),
+                new OptionalAwareDecoder(new TextDelegateDecoder(new JacksonDecoder(ObjectMappers.guavaJdk7()))),
```
markelliot, line 50: we really need to pay down the debt on this at some point

markelliot:
thoughts on contributing this directly to Feign?

http-clients/src/main/java/com/palantir/remoting/http/QueryMap.java
nmiyake, line 14: Yes -- just tried it in a unit test, and Multimaps can be written to JSON and read back from JSON properly (at least when using the JDK 7 Guava ObjectMapper).

However, after thinking about it, this object doesn't actually need to support JSON serialization since it's converted to query parameters on the client side and is never actually passed over the wire. Will remove the annotations.

http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
nmiyake, line 50: Can you clarify what debt you're referring to?

nmiyake:
Thought about it, but don't know if it would fit their model exactly -- poked around their source code, and from what I can tell they don't ship with any custom `Encoder` subclasses. Although I haven't spent a ton of time on this, I did look through their codebase a bit to see if there would be a cleaner way to add it there, but couldn't come up with anything that would be straight-forward.

My calculus was basically that this solution was just simple/flexible/clean enough for me to be comfortable to use as a client, but not quite elegant enough build directly into Feign (at least in its current form). However, if either of you have thoughts on how this might be done in Feign directly, please let me know and would be open to looking into it.

http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
markelliot, line 50: Sorry meant to tag Rob with that. We have a growing number of encoders and decoders and should really come up with a different config pattern for adding all of them. Feel free to ignore this.

markelliot:
Would you mind opening an issue and possibly linking to this PR as what it might look like on the Feign repo? They've been super responsive and it'd be cool to float the idea.

nmiyake:
Sure, will do.

http-clients/src/test/java/feign/QueryMapUsingServerWithContextTests.java
nmiyake, line 50: OK cool, yeah, that makes sense. Tagging @uschi2000 and throwing out some ideas that came to my mind (noticed this as well when I was writing this code).

If most of the decoders/encoders are delegating ones, could have a pattern where you just supply them all and have them chained. Should work for most simple cases, although in some cases could lead to some weird dependencies. 

Another approach could be to have some kind of "handler" pattern where the thing you're registering has a predicate to determine whether or not to process the item and will then process it if so. In this scenario, builder/config could take a collection of processors and have a contract that any input will be processed by the first processor whose predicate is satisfied (although this can have corner cases depending on the situation as well like if a handler for a compound type wants to use the "Optional" handler for the sub-components of its type).

uschi2000, line 50: +1, that's pretty much exactly what I had suggested before.

nmiyake:
Closing this out since I think this will use a different approach now.

My PR for adding `@QueryMap` to Feign is progressing, and should hopefully merge soon. This should allow it to be supported client-side natively.

There will still be a bit of work since this annotation will only be on `Contract.Default` and not on `JAXRSContract`, which is what we use for the builders in http-remoting. I may need to implement something like `JAXRSContractWithQueryParams` to add support for the annotation to JAXRS and then have http-remoting use that or something (or make the builders more permissive). Will submit a PR for that appraoch once I have it.

PR #55, opened 2016-02-18
handle null exception message

pnepywoda:
@uschi2000


http-servers/src/main/java/com/palantir/remoting/http/server/JsonExceptionMapper.java
```diff
@@ -47,30 +48,33 @@ public JsonExceptionMapper(boolean includeStackTrace) {
 
     @Override
     public final Response toResponse(T exception) {
-        String message = exception.getMessage();
+        String message = Objects.toString(exception.getMessage());
+        String errorString = message;
```
uschi2000: can you name these errorMessage and logMessage?

uschi2000:
rebase and we're good.

pnepywoda:
rebased

PR #59, opened 2016-02-19
Flatten configuration and match Dropwizard names

nmiyake:


ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -28,18 +30,44 @@
...
+
+    @SuppressWarnings("checkstyle:designforextension")
+    @Value.Default
+    public String keyStoreType() {
+        return "JKS";
+    }
+
+    // alias of the certificate to use in the key store.
+    // If absent, first entry returned by key store is used.
+    public abstract Optional<String> certAlias();
```
markelliot: Even though this is what Dropwizard uses I think it's ambguous here and would prefer something appropriately specific, like keyStoreKeyAlias

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -28,18 +30,44 @@
...
+
+    @Value.Check
+    protected final void check() {
+        Preconditions.checkArgument(!keyStorePath().isPresent() || keyStorePassword().isPresent(),
+                "keyStorePassword cannot be absent if keyStorePath is present");
```
markelliot: You need to also reject if keyStorePassword is present but keyStorePath is not

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
nmiyake: Yeah in this case clarity probably should win out -- will update.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -28,18 +30,44 @@
 @Value.Style(visibility = ImplementationVisibility.PACKAGE)
 public abstract class SslConfiguration {
 
-    @Value.Parameter
-    public abstract TrustStoreConfiguration trust();
+    public abstract URI trustStorePath();
```
markelliot: These are different from the Dropwizard spec in that they take URIs and not paths. Is this still actually a reasonable name?

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
nmiyake: Noticed right after I pushed this and implemented locally -- pushing up now :)

```
    @Value.Check
        Preconditions.checkArgument(keyStorePath().isPresent() == keyStorePassword().isPresent(),
                "keyStorePath and keyStorePassword must both be present or both be absent");
    }
```

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
nmiyake: Fair point. I'm okay with `trustStoreUri` and `keyStoreUri` to keep same general format but to make the type more clear from the param name.

markelliot: I guess it's up to what others want. I don't have strong preferences, but I think this does remove clarity.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -28,18 +30,51 @@
...
+
+    @Value.Check
+    protected final void check() {
+        Preconditions.checkArgument(
+                keyStorePath().isPresent() == keyStorePassword().isPresent(),
+                "keyStorePath and keyStorePassword must both be present or both be absent");
+
+        Preconditions.checkArgument(
+                !keyStoreKeyAlias().isPresent() || keyStorePath().isPresent(),
+                "keyStoreKeyAlias must be present if keyStorePath is present");
```
markelliot: hm, I think `keyStoreKeyAlias` may only be present when `keyStorePath` is present, but is not required.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -75,23 +76,24 @@ private static TrustManagerFactory createTrustManagerFactory(TrustStoreConfigura
         }
     }
 
-    private static KeyManagerFactory createKeyManagerFactory(KeyStoreConfiguration keyStoreConfig) {
+    private static KeyManagerFactory createKeyManagerFactory(SslConfiguration sslConfig) {
```
markelliot: the args to this should probably explicitly be `Path keyStorePath`, `String keyStorePassword`, `String keyStoreType`, `Optional<String> keyStoreKeyAlias`

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -59,12 +60,12 @@ public static SSLContext createSslContext(SslConfiguration config) {
         }
     }
 
-    private static TrustManagerFactory createTrustManagerFactory(TrustStoreConfiguration trustStoreConfig) {
+    private static TrustManagerFactory createTrustManagerFactory(SslConfiguration sslConfig) {
```
markelliot: the args to this should probably explicitly be `Path trustStorePath`, `Optional<String> trustStorePassword`, `String trustStoreType`

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
nmiyake: Oops, sorry, message is incorrect here. What you're saying is correct -- this is meant to throw an error if `keyStoreKeyAlias` is specified when `keyStorePath` is not present.

markelliot:
this lgtm

nmiyake:
Great -- will rebase, squash, and push back up

PR #62, opened 2016-02-19
added discoverable subtype resolver to standard object mapper

jlzhao27:
gatekeeper relies on discoverability for dynamic subtype resolving, we could omit it but people's client will break unless they explicitly modify the om.

### DIFF
--- a/gradle/libs.gradle
+++ b/gradle/libs.gradle
@@ -1,7 +1,8 @@
 project.ext.libs = [
     'dropwizard': [
         'testing': 'io.dropwizard:dropwizard-testing:0.8.2',
-        'core': 'io.dropwizard:dropwizard-core:0.8.2'
+        'core': 'io.dropwizard:dropwizard-core:0.8.2',
+        'jackson': 'io.dropwizard:dropwizard-jackson:0.8.2'
     ],
     'feign': [
         'jackson': 'com.netflix.feign:feign-jackson:8.11.0',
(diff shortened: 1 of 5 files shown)


markelliot:
I think this is probably a bad addition. The ObjectMappers stuff is pluggable, if we need to provide a factory to construct the standard delegates we should do that. Also, we should maybe consider not using the discoverable objectmapper stuff in favor of explicit sub-type resolution.

jlzhao27:
sure, as long as there is a standard. do other projects explicitly register subtypes and provide a unified object mapper?

markelliot:
I've been largely unable to convince others that type polymorphism is practical and reasonable, but GK's data structure make it impossible to ignore. Where we've had to do it (Stemma is in this boat, for instance), explicit sub-type annotations have been used over the discoverable object mapper. I'm generally afraid of taking a dropwizard dep here because it means http-remoting becomes a driver for dropwizard upgrades, which is net negative, IMO.

jlzhao27:
okay, so the recommendation is to have something like: https://github.palantir.build/elements/multipass/pull/221/files#diff-ed973ce939454dce96155647833a2461L1 and GK providing the right object mappers?

markelliot:
No, I'd advise GK switch to explicit annotations and use the standard object mappers, or alternatively keep _just_ a custom `ObjectMappers` class and still use the http-remoting constructors.

jlzhao27:
okay, I'm going to change the GK annotations then to prevent clients from having to care.

Issue #32, opened 2016-02-05
[Needs Investigation] Returning an Optional<String> when Content-Type=application/json causes unexpected errors


leeavital:
```
Exception in thread "main" feign.FeignException: Unexpected character ('/' (code 47)): maybe a (non-standard) comment? (not recognized as one since Feature 'ALLOW_COMMENTS' not enabled for parser)
 at [Source: java.io.BufferedReader@2b662a77; line: 1, column: 2] reading GET http://localhost:8442/.....
    at feign.FeignException.errorReading(FeignException.java:48)
    at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:127)
    at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:71)
    at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:94)
    at com.sun.proxy.$Proxy19.getPath(Unknown Source)
```

This was a problem in 0.3.0 and is still a problem in 0.4.0.

markelliot:
Any chance you could document a minimal example that reliably reproduces the problem? (Or...contribute that as a test?)

nmiyake:
I tried this out for GET requests, and I do get exceptions when return type is annotated with `@Produces(MediaType.APPLICATION_JSON)` rather than `@Produces(MediaType.TEXT_PLAIN)` if the method is returning either `String` or `Optional<String>`. However, exception I get is different:

```
feign.FeignException: Unrecognized token 'foo': was expecting 'null', 'true', 'false' or NaN
 at [Source: java.io.BufferedReader@146d984e; line: 1, column: 7] reading GET http://localhost:61992/optionalStringJson
    at feign.FeignException.errorReading(FeignException.java:48)
    at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:127)
    at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:71)
    at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:94)
    at com.sun.proxy.$Proxy59.getOptionalStringJson(Unknown Source)
    at feign.OptionalAwareDecoderTest.testOptionalStringJson(OptionalAwareDecoderTest.java:87)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #64, opened 2016-02-21
String decoding

nmiyake:


nmiyake:
PR is only for the purposes of possibly illustrating #32 (not meant for merge)

nmiyake:
This was for illustration purposes only and doesn't seem to be progress on the ticket, so closing this out.

PR #65, opened 2016-02-22
Add script to generate test certificates

nmiyake:
- Add 'certSetup.sh' script that generates
  CA, server and client key pairs signed by
  the CA, an empty CRL and associated files.
- Update test security files to be files
  generated by the script.


nmiyake:
@uschi2000 @markelliot 

This doesn't change any functional code, but adds a script that can be used to generate all of the security material required for tests (CA, service key pairs signed by CA, P12 files, JKS files, etc.). Should be useful if there's a need to change any of the material at a later time.

uschi2000:
tested locally, works fine. thanks nick!

PR #66, opened 2016-02-22
add leader election aware client

mgelmi:


uschi2000:
I would like to kick off a discussion around adopting hystrix before we re-implement parts of it here.

mgelmi:
wanted to resurrect this PR...I checked Hystrix and Ribbon and it seems neither of them have a concept of leader node. Essentially we would need to reimplement this logic anyway, just with a different framework. If we think this feature is not common enough to warrant submitting it here, I am also ok with having a custom client for our project. Let me know your thoughts. @NeilRickards @j-baker

markelliot:
@mgelmi sorry, we never did get to this and it's sat here in limbo. To leave it unambiguous, we're electing not to move forward with this. Thinking about it some, it'd be straight forward to make a dynamic proxy that wraps a Feign client to perform the retry, and that's likely cleaner than replacing the internals.

PR #67, opened 2016-02-24
inline FluentIterable.of to be compatible with guava <18

Xenorith:


pwoody:
Can we shade guava?

markelliot:
I'd rather not, would prefer to rewrite this to remove the Guava dep rather than shade just for this one liner.

pwoody:
cool - just something to be wary of for all of our libraries that interact with spark/hadoop. guava versions always cause issues (hence spark shaded theirs).

markelliot:
We use so little of the Guava library that forcing it to whatever version you need is probably fine in the current mode. If we force shading of the Guava Optional class then that sort of ruins how this gets used, so I'm pretty hesitant to do that.

uschi2000:
any objections to merging this for a 0.7.1 release?

http-clients/src/main/java/com/palantir/remoting/http/GuavaOptionalAwareContract.java
```diff
@@ -67,7 +68,7 @@ public GuavaOptionalAwareContract(Contract delegate) {
                     Class<?> cls = parameterTypes[i];
                     if (cls.equals(Optional.class)) {
                         FluentIterable<Class<?>> paramAnnotations =
-                                FluentIterable.of(annotations[i]).transform(EXTRACT_CLASS);
+                                FluentIterable.from(Lists.newArrayList(annotations[i])).transform(EXTRACT_CLASS);
```
pnepywoda, line 71: ImmutableList.of(). whatever. do it next time

pnepywoda:
+1
We may want to actually downgrade guava in our widely-used libraries to guava 16, or even 14 which is what spark uses (http://mvnrepository.com/artifact/org.apache.spark/spark-core_2.10/1.6.0)

http-clients/src/main/java/com/palantir/remoting/http/GuavaOptionalAwareContract.java
Xenorith, line 71: went with inlining exactly what was in guava

  public static <E> FluentIterable<E> of(E[] elements) {
    return from(Lists.newArrayList(elements));
  }

uschi2000:
can you rebase please, @Xenorith ?

PR #68, opened 2016-02-24
Fix logger output for numbers

nmiyake:
Print numeric value of ThreadLocalInteger
instead of its default Object toString.


nmiyake:
@uschi2000 @markelliot 

Minor bug fix

PR #69, opened 2016-02-25
allow specifying timeout options

pnepywoda:
@markelliot @uschi2000

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -22,6 +22,8 @@ import com.palantir.remoting.http.errors.SerializableErrorErrorDecoder;
 import feign.InputStreamDelegateDecoder;
 import feign.InputStreamDelegateEncoder;
 import feign.OptionalAwareDecoder;
+import feign.Request;
+import feign.Request.Options;
 import feign.TextDelegateDecoder;
 import feign.jackson.JacksonDecoder;
 import feign.jackson.JacksonEncoder;
@@ -34,6 +36,8 @@ import feign.jaxrs.JAXRSContract;
  */
 public final class FeignClients {
 
+    private static final Options DEFAULT_TIMEOUT_OPTIONS = new Request.Options();
+
     private FeignClients() {}
 
     /**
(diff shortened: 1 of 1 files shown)


http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -41,6 +45,14 @@ private FeignClients() {}
...
      */
     public static FeignClientFactory standard() {
+        return standard(DEFAULT_TIMEOUT_OPTIONS);
+    }
+
+    /**
+     * Provides a {@link FeignClientFactory} with an {@link ObjectMapper} configured with {@link
+     * com.fasterxml.jackson.datatype.guava.GuavaModule} and {@link com.fasterxml.jackson.datatype.jdk7.Jdk7Module}.
+     */
+    public static FeignClientFactory standard(Request.Options timeoutOptions) {
```
uschi2000, line 55: seems like we need a BuilderFactoryBuilder :(

uschi2000:
looks OK

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
pnepywoda, line 55: > Yuck

.....or call it Warehouse instead ;)

markelliot:
What if we did FeignClients.withTimeout(...).standard()?

Could keep things internal to FeignClients?

Am worried at this point we're just recreating the Feign builder and ought to pick a different wrapping point.

Sent from my iPhone

On Feb 25, 2016, at 7:48 AM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

looks OK

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
jlzhao27, line 55: this only sets read and connection right, what if we want write timeout?

uschi2000, line 55: there's currently no good support for write timeout in feign. should
probably fix it there first.

pnepywoda, line 55: Ok yes many things can be improved in the future, but this is specifically
so Foundry can stop copy/pasting the method with timeoutOptions and simply
use http-remoting. Are there objections to this commit as-is?

markelliot, line 55: It's probably fine, we need to make some real investment in this library in the near future.

Sent from my iPhone

pnepywoda, line 55: Ok. I'll wait patiently for a +1 ;)

PR #71, opened 2016-02-26
Simplify FeignClients methods

nmiyake:
Use methods to get default contract,
encoder and decoder and use a declarative
framework for decorators.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -19,12 +19,14 @@ package com.palantir.remoting.http;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Optional;
 import com.palantir.remoting.http.errors.SerializableErrorErrorDecoder;
-import feign.InputStreamDelegateDecoder;
+import feign.Chain;
+import feign.InputStreamDecoder;
 import feign.InputStreamDelegateEncoder;
-import feign.OptionalAwareDecoder;
+import feign.OptionalDecoder;
 import feign.Request;
 import feign.Request.Options;
-import feign.TextDelegateDecoder;
+import feign.TextDecoder;
+import feign.codec.Decoder;
 import feign.jackson.JacksonDecoder;
 import feign.jackson.JacksonEncoder;
 import feign.jaxrs.JAXRSContract;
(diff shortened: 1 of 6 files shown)


nmiyake:
@uschi2000 @markelliot 

This may not totally solve the root cause of the pain, but I do think it's an incremental improvement. Although the `Decorators` stuff may be a bit overkill, I like that it allows the decoration to be done in a more declarative manner, and the ordering is also more natural -- it makes the param ordering go from deepest object to shallowest object rather than the other way around, which is what happens with the constructor syntax.

Let me know what you think.

nmiyake:
Ping? Low-pri, but curious to hear feedback.

uschi2000:
will look on tuesday

http-clients/src/main/java/com/palantir/remoting/http/Decorators.java
```diff
@@ -0,0 +1,112 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.reflect.TypeToken;
+import java.lang.reflect.Constructor;
+import java.util.Arrays;
+import java.util.Set;
+
+/**
+ * Methods that allow classes that use the decorator pattern to be instantiated in a declarative manner.
+ */
+final class Decorators {
```
uschi2000: This is a little too much magic for my liking, TBH. When composition is true delegation (i.e., the response of the delegator always depends on the response of the delegate), then IMO it's fine to use something like `new Delegator(new Delegate())`. Things got a little ugly where we are composing `new A(new B(new C()))` although A or B usually short-circuit if they can. In that case a better pattern is probably to make explicit which handler (decoder, encoder, etc) can handle which object, either by return value (Optional<Foo>) or by an explicit Boolean canHandle(...). Then a generic `return Chain.firstHandler(new A(), new B(), new C())` will return the return value of the first handler than `canHandle()` the request. This makes very explicit that the behaviour of A does not depend on the behaviour of B or C.

uschi2000:
btw, gitcolony.com is enabled for this repository, feel free to follow the PR there.

http-clients/src/main/java/com/palantir/remoting/http/Decorators.java
nmiyake: OK cool, thanks for the feedback, totally fair. May take a look at this simplification when I have some spare cycles, but not urgent.

Hmm, went to GitColony but didn't see this PR there?

uschi2000: Strange. I don't see your user in gitcolony.

nmiyake:
@uschi2000 , updated PR with rough sketch of the "handler" approach (not merge-ready, but wanted to submit rough sketch for feedback before touching up).

The cleanest approach would just be to have the handlers implement `feign.codec.Decoder` and add a `canHandle` method. Unfortunately, this isn't enough for the `OptionalAwareDecoder` because even when it can handle the decoding, it still needs a `Decoder` to delegate to in order to decode the value that's inside of the optional (and it needs to wrap that before returning).

My approach for this is to have a chain have a notion of the "base" Decoder (the one that will be used if none of the handlers can handle the decode request), and to also pass that to the "decode" request of each handler in case the handler wants to make use of them.

Open to feedback on this. It may also be the case that any attempt to simplify what's here adds sufficient complexity such that it's not really worth changing what's already there...

nmiyake:
There are some subtleties here and probably not worth pursuing this for now -- closing out.

PR #72, opened 2016-02-26
Feature/query map annotation

nmiyake:


nmiyake:
Preview of what QueryMap annotation support looks like in http-remoting.

Once Feign releases the next release that has my QueryMap PR that merged there, this commit will add `@QueryMap` support for JAX-RS annotated interfaces (whereas my PR on Feign only adds the annotation for the default Feign contract)

nmiyake:
@uschi2000 @markelliot now that the Feign upgrade has merged, this works as expected. Feedback welcome. Will also look at updating #71 with a simpler approach to simplify some of the decoration stuff, but that would be somewhat independent.

nmiyake:
@uschi2000 ping for feedback. Not urgent, but can't use `@QueryMap` on JAX-RS format interfaces in standard http-remoting without adding something like this.

Updated the PR to just be the commit to add this functionality (separated from PR for simplifying some of the decorator methods).

http-clients/src/main/java/com/palantir/remoting/http/GuavaOptionalAwareContract.java
```diff
@@ -39,50 +35,32 @@
...
-                            md.indexToExpanderClass().put(i, GuavaEmptyOptionalExpander.class);
-                        } else if (paramAnnotations.contains(QueryParam.class)) {
-                            md.indexToExpanderClass().put(i, GuavaNullOptionalExpander.class);
-                        } else if (paramAnnotations.contains(PathParam.class)) {
-                            throw new RuntimeException(String.format(
-                                    "Cannot use Guava Optionals with PathParams. (Class: %s, Method: %s, Param: arg%d)",
-                                    targetType.getName(), method.getName(), i));
-                        }
-                    }
+    protected void processMetadata(Class<?> targetType, Method method, MethodMetadata metadata) {
```
uschi2000, line 45: is this just an equivalent rewriting?

nmiyake, line 45: Yes, pushes up common structure/functionality to `AbstractDelegatingContract` and then just re-writes to take advantage of that and reduce duplicated code -- no functionality change.

uschi2000:
code-wise, this looks cool

uschi2000:
I wonder though if QueryMap is going to be a standard use-case and if it should thus be included in the standard FeignClients factories by default?

nmiyake:
Fair -- I don't feel strongly about including it by default. However, I would like it to be in http-remoting and have the ability to include it in generated clients so that if we publish an interface that uses it, people can still use fairly standard http-remoting practices to instantiate a client from the interface.

uschi2000:
Agreed. @melliot, any concerns?

nmiyake:
Only thing on my end is that if we don't include as standard, then improving the "create clients with custom settings" story does become more important. The basic pattern would be something like:

```
FeignClientFactory standardFactory = FeignClients.standard();
Builder = FeignClientFactory.builder(factory);
builder.contract(new QueryMapAwareContract(standardFactory.getContract()));
return builder.build();
```

Ideally would like to combine with the commit/work to make the over-all process easier, but just want to avoid a situation where clients have to manually call `com.palantir.remoting.http.FeignClientFactory.of(Contract, Encoder, Decoder, ErrorDecoder, Function<Optional<SSLSocketFactory>, Client>)` with parameters that just happen to match the defaults by looking at source code.

markelliot:
It's probably fine to include in the standard setup...we should probably revisit whether we're getting mileage out of Feign here, though, increasingly we seem to be working around its setup.

nmiyake:
Cool. Would be interested to be part of the broader discussion as well, but in the mean time sounds like this PR is good as-is then?

nmiyake:
Thanks @uschi2000 ! Would it be possible to get a new minor release for this?

uschi2000:
Done, 0.8.1

PR #74, opened 2016-03-01
basic logging to help debug 404 client errors

pnepywoda:
@uschi2000

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
@@ -22,11 +22,13 @@ import com.google.common.collect.Lists;
 import feign.Client;
 import feign.Contract;
 import feign.Feign;
+import feign.Logger.Level;
 import feign.Request;
 import feign.codec.Decoder;
 import feign.codec.Encoder;
 import feign.codec.ErrorDecoder;
 import feign.okhttp.OkHttpClient;
+import feign.slf4j.Slf4jLogger;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


pnepywoda:
logs to DEBUG on slf4j logger for FeignClientFactory so clients can activate/deactivate via yml

pnepywoda:
@markelliot

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -104,6 +106,8 @@ public static FeignClientFactory of(
                 .errorDecoder(errorDecoder)
                 .client(clientSupplier.apply(sslSocketFactory))
                 .options(options)
+                .logger(new Slf4jLogger(FeignClientFactory.class))
+                .logLevel(Level.BASIC)
```
uschi2000, line 110: I assume you can override this using the standard mechanism provided by your slf4j implementation?

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -124,6 +128,8 @@ public static FeignClientFactory of(
                 .client(client)
                 .retryer(target)
                 .options(options)
+                .logger(new Slf4jLogger(FeignClientFactory.class))
```
markelliot: I'd prefer we use a different string than FeignClientFactory, and should probably encode it as a constant.

pnepywoda: Ok, using FeignClients

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
pnepywoda, line 110: Yeah, by default it logs nothing cause it logs on slf4j's DEBUG by default

uschi2000:
:+1: 
anything left, @markelliot ?

markelliot:
lgtm

PR #75, opened 2016-03-02
Rename http-clients to feign-clients

uschi2000:


markelliot:
Hm, I wonder if it's reasonable to keep this as 'http-clients', ideally we're pushing a particular strategy.

uschi2000:
where do you put retrofit things then? i'd like to avoid picking up
retrifit dependencies if all I want is a feign&okhttp client.

markelliot:
One option: keep http-clients, add retrofit-clients.

Alternatively, I suppose we can state a strong preference for feign-clients in the README.

markelliot:
Also, are we changing the published artifact name as a result of this?

uschi2000:
Yep, all or nothing. I am happy to keep http-clients and add
retrofit-clients though.

PR #76, opened 2016-03-03
Upgrade feign to version 8.14.4

leeavital:


schlosna:
Looking at https://github.com/Netflix/feign/blob/master/CHANGELOG.md this bumps OkHttp and Gson:

Version 8.14
- Add support for RxJava Observable and Single return types via the HystrixFeign builder.
- Adds fallback implementation configuration to the HystrixFeign builder
- Bumps dependency versions, most notably Gson 2.5 and OkHttp 2.7

Version 8.13
- Never expands >8kb responses into memory

Version 8.12
- Adds Feign.Builder.decode404() to reduce boilerplate for empty semantics.

Version 8.11
- Adds support for Hystrix via a HystrixFeign builder.

nmiyake:
Don't think this should impact anything, but the next Feign release should be soon (the maintainer had actually told me it would be yesterday, but looks like they're a bit behind) and contains a PR that I merged that I'd like for http-remoting. Should be a minor bump so don't imagine it will be hard to bump again after this, but just wanted to mention it in case that affects when we want to do the bump in this library.

leeavital:
@nmiyake  that's fine by us. This is easily patchable with gradle exclusions.

uschi2000:
ok, let's wait for the next feign release.

nmiyake:
I've opened PR #77 to upgrade to 8.15.0 (and fixes other dependencies and unit tests)

nmiyake:
@uschi2000 or @markelliot , this can be closed since #77 merged.

uschi2000:
Closing in lieu of #77

PR #79, opened 2016-03-06
Add documentation and tests for OptionalAwareDecoder

nmiyake:
Explicitly document that decoder only applies to the
outer-most Optional object on a response body that is
declared as an Optional and add test to demonstrate.


nmiyake:
Fairly minor documentation change, but this behavior wasn't immediately obviously to me when I was reading through the library (encountered this while working on the chaining stuff), so figured it wouldn't hurt to add explicit explanation and test.

nmiyake:
Hmm, actually, never mind... After thinking about this a bit more I think it's not necessary -- all decoders generally just act on top-level objects (multi-level decoding is the exception rather than the norm), so on second thought don't think this is necessary. Closing out.

PR #80, opened 2016-03-09
Allow JAX-RS 1 clients to consume JAX-RS 2 services

punya:
1. Some `javax.*` packages are in the JDK (e.g. `javax.swing`) but others aren't (`javax.ws.rs.*`).
2. Jersey 1 has `javax.ws.rs` packages in a jar with a different name than Jersey 2 does.
3. `ClientErrorException` and `ServerErrorException` exist in Jersey 2 but not Jersey 1. Note, there are Jersey 1 classes in a different _package_ that have the same names, but sufficiently different behavior that I didn't deem it worth translating types.
4. `WebApplicationException` has new constructors available in Jersey 2, that it didn't have in Jersey 1.

### DIFF
--- a/error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
+++ b/error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
@@ -16,6 +16,8 @@
 
 package com.palantir.remoting.http.errors;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.List;
@@ -35,13 +37,28 @@ public abstract class SerializableError {
 
     public abstract String getMessage();
 
-    public abstract Class<? extends Exception> getExceptionClass();
+    @JsonProperty("exceptionClass")
+    public abstract String getExceptionClassName();
+
+    /**
+     * @deprecated Use {@link #getExceptionClassName()} instead.
+     */
+    @Deprecated
+    @JsonIgnore
+    @SuppressWarnings("unchecked")
+    public final Class<? extends Exception> getExceptionClass() {
+        try {
+            return (Class<? extends Exception>) Class.forName(getExceptionClassName());
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
 
     @Nullable
     public abstract List<StackTraceElement> getStackTrace();
 
     public static SerializableError of(String message, Class<? extends Exception> exceptionClass) {
-        return ImmutableSerializableError.builder().message(message).exceptionClass(exceptionClass)
+        return ImmutableSerializableError.builder().message(message).exceptionClassName(exceptionClass.getName())
                 .build();
     }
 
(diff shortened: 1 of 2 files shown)


punya:
@uschi2000 @markelliot @akroy FYSA.

TheBeruriahIncident:
For my own understanding, why is some some subset of {ServerError, ClientError, BadRequest....} not on the classpath if using Jersey 1? Those classes appear to be javax, which I thought was a standard part of Java?

uschi2000:
@akroy , BadRequestException, for example, is in package `javax.ws`, but comes in through the javax-ws jar, ie, is not on the classpath in every default JVM.

error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
```diff
@@ -35,13 +37,28 @@
 
     public abstract String getMessage();
 
-    public abstract Class<? extends Exception> getExceptionClass();
```
uschi2000, line 38: ah, this was an annoying oversight :(

error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
```diff
@@ -35,13 +37,28 @@
...
+    @JsonProperty("exceptionClass")
+    public abstract String getExceptionClassName();
```
uschi2000, line 41: I suspect the server code should use the String-based mechanism, see http-servers:JsonExceptionMapper.java

punya:
@akroy there are multiple layers of inconsistency :sob: 
1. Some `javax.*` packages are in the JDK (e.g. `javax.swing`) but others aren't (`javax.ws.rs.*`).
2. Jersey 1 has `javax.ws.rs` packages in a jar with a different name than Jersey 2 does.
3. `ClientErrorException` and `ServerErrorException` exist in Jersey 2 but not Jersey 1. Note, there are Jersey 1 classes in a different _package_ that have the same names, but sufficiently different behavior that I didn't deem it worth translating types.
4. `WebApplicationException` has new constructors available in Jersey 2, that it didn't have in Jersey 1.

error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
punya, line 41: I'm not 100% sold on this. The current proposal explicitly avoids breaking any existing APIs so that (for example), people can freely mix and match different versions of http-remoting in their app. If my server relies on old behavior and I pull in a client that relies on new behavior, I don't want my server to break at runtime due to link errors.

On the other hand, the API break gets people using the "right" APIs, so... if you think it's worth it I can make the change.

punya:
@robert3005 also.

error-handling/src/main/java/com/palantir/remoting/http/errors/SerializableError.java
uschi2000, line 41: I think it's fine to not support server:0.9.0 with client:0.8.0.

uschi2000, line 41: nevermind.

markelliot:
Squash?

punya:
@markelliot @uschi2000 squashed.

punya:
Do we publish snapshots anywhere? Or should I wait for the next release?

markelliot:
Cutting a release now.

Issue #81, opened 2016-03-09
Think about behavior when exceptions are passed through


punya:
Say I have the following chain of calls, all made using http-remoting:

Client A ----> Server B ----> Server C

If Server C produces a 404 (encoded as a `NotFoundException`) and server B doesn't handle and wrap it explicitly, then this results in a `NotFoundException` getting propagated to Client A. However, from Client A's perspective, the error really should be some sort of 5xx, because the not-found condition is an implementation detail.

We could handle this a couple of ways in the http-remoting API:
1. Always throw a `WrappedException` type instead of one of the exception types that has a special meaning to Jersey.
2. Allow people to specify exception type mappings when they make their `FeignClient`.
3. Document this oddity and hope that people explicitly handle 4XX errors to convert them into 5XX errors.

My preference is (1) since it's explicit, but it would certainly be an API break.

Tagging @robert3005, @markelliot, @ashwinr and @derekcicerone for thoughts on this.

uschi2000:
The current behaviour mimics what happens in Java: as long as a particular
method doesn't catch/rethrow, you'll see the original exception (together
with the original HTTP error code).
I don't quite understand how wrapping everything would make your life (as a
user) better, can you elaborate? Specifically, why is NotFound/4xx not the
correct response for the user? After all, something wasn't found somewhere.
If B wants to surface this error condition to its users as a 5xx, then it
can catch/rethrow with appropriate error code.

punya:
> I don't quite understand how wrapping everything would make your life (as a
> user) better, can you elaborate? Specifically, why is NotFound/4xx not the
> correct response for the user? After all, something wasn't found somewhere.

A 404 response code means that the thing I (the client) was looking for is absent, not that an arbitrary thing in the stack wasn't found. For example, say the client is trying to create a remote file, and the intermediate server needs to (1) check permissions by consulting a policy server and (2) actually write the file. If I've misconfigured my policy server somehow, the first step (check permissions) might fail because no policy was found (404). That should be relayed to the user as either 401 (unauthorized) or 503 (proxy error). From the user's perspective, a 404 response is bizarre because I never asked the server to find anything for me!

By wrapping, the user always gets some kind of 5XX error, which means that they should contact the intermediate server to improve their error handling rather than looking at their own request to understand how it was malformed.

uschi2000:
In this particular example, I'd expect that either the policy server throws
401/403 directly, or otherwise that the application catches the policy
[...]

PR #82, opened 2016-03-09
[WIP] Support more exceptions

punya:
Depends on #80. May be worth resolving #81 conceptually before deciding what to do with this PR.


punya:
Turns out this isn't necessary since the other `javax.ws` exceptions have 2-argument `(String, Throwable)` constructors (which are already handled).

PR #87, opened 2016-03-19
Replace QueryMapAwareContract with JAXRSWithQueryMapContract

nmiyake:
Delegating model made it impossible to support @QueryMap
annotation on methods that already had body parameters.
This changes creates a new JAXRSWithQueryMapContract contract
that replaces the JAXRSContract more directly to allow full
QueryMap support.


nmiyake:
@uschi2000 @markelliot 

This change is a bit more aggressive, so understand if there's pushback. However, we have an API project that communicates with a server that takes query parameters for `PUT` and `POST` methods, so this change is required to use `QueryMap` on those kinds of methods.

If we decide that this change shouldn't be made in the core project, then we'll include this in our API project instead and require the clients instantiate their `FeignClient` objects using this contract.

uschi2000:
I'm fine fine this. @markelliot ?

http-clients/src/main/java/feign/jaxrs/JAXRSWithQueryMapContract.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Contract that provides the functionality of {@link JAXRSContract} with full support for
+ * the {@link QueryMap} annotation. Allows the QueryMap annotation to be used in methods
+ * that also take a body parameter (such as {@link javax.ws.rs.POST or {@link javax.ws.rs.PUT}}).
+ */
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public final class JAXRSWithQueryMapContract extends Contract.BaseContract {
+
+    private final JAXRSContract delegate = new JAXRSContract();
```
markelliot, line 34: Why does the delegate have to be a JAXRSContract?

nmiyake, line 34: Technically it doesn't have to, but the `DefaultContract` already handles QueryMaps, so this class was designed specifically as "a JAXRS contract that adds QueryMap support".

I could make this part pluggable, but don't particularly see the value that it would add in this case.

nmiyake:
Still want this, but now that my change for HeaderMap has merged to Feign (https://github.com/Netflix/feign/pull/362), would like to include that as part of this change as well -- closing out the current PR until Feign cuts the next release.

PR #89, opened 2016-03-21
Ssl config provider

nmiyake:


nmiyake:
@uschi2000 @markelliot 

Interested in feedback on this approach. This change makes it trivial to create a connection with trust based on a single certificate file in PEM/DER format:

```
SslConfigurationProvider clientConfig = CertificateTrustConfiguration.of(TestConstants.CA_PEM_CERT_PATH);
SSLSocketFactory sslSocketFactory = SslSocketFactories.createSslSocketFactory(clientConfig);
```

It would also be easy to change this implementation (or add a separate implementation) that could accept a collection of certificates or point to a directory with certificates to add all of them.

The current approach does conflate the logic with the representation (`SslConfiguration` and `CertificateTrustConfiguration` are both `Immutables` definitions that also implement the `SslConfigurationProvider` interface and implement the behavior), but I didn't think it made much sense to have a separate class that implemented the interface and just wrapped the representation (since you'd end up with 1:1 between representation and interface).

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -20,15 +20,29 @@
...
+/**
+ * Configuration that provides a {@link TrustManagerFactory} based on a trust store and optionally provides
+ * a {@link KeyManagerFactory} based on a key store. The stores can be in JKS or PKCS12 format.
+ */
 @JsonDeserialize(as = ImmutableSslConfiguration.class)
 @JsonSerialize(as = ImmutableSslConfiguration.class)
 @Value.Immutable
 @Value.Style(visibility = ImplementationVisibility.PACKAGE)
-public abstract class SslConfiguration {
+public abstract class SslConfiguration implements SslConfigurationProvider {
```
uschi2000: I think I'd prefer keeping this a value type without any logic.

nmiyake:
Example of what it would look like to separate out logic. In this example, the JSON objects to define a `provider()` method that handles the instantiation (and if we go this route, I'd likely define an interface with the `provider()` method and have the JSON objects implement it).

If we want to keep behavior and storage completely separate, then we would remove the `provider()` method and replace all current usages where `obj.provider()` is called to be `new Wrapper(obj)` instead.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
markelliot: Yeah, I'd also prefer data objects with a factory that knows how to multiplex. I'd also see the file or directory based configurations as naturally extending using the "type" field for trust and key stores.

nmiyake: Yeah, I did consider that, and that would probably work fine for this case. I thought it might be a bit awkward since it changes how `type` is interpreted (right now we use it as the store type, but that's not strictly necessary so we could change it to signal "type of config" instead and branch logic based on that), and it's possible that there could be config that either doesn't fit the fields described in this particular object or require more fields.

However, doing this based on type would keep the storage simpler (moves the branching of logic to the class that handles the object instead), and for the current file/directory proposal, the API should work.

I'll push up a PR with this approach to compare.

markelliot: I guess I was thinking that the file/directory based config would just be a puppet-style SSL config directory, and we'd set type to "puppet" and path to the base dir. 

Shouldn't actually need more config options?

nmiyake: Yup, Puppet-style directory makes sense, and am fine with that.

The one other use case that I imagined/find myself using locally sometimes is wanting to point to a single PKCS#12 cert (in PEM or DER format) and using that as the trust source. To support this in the Puppet scenario, I would need to create `certs` directory inside the base directory and place the PEM file there.

This isn't too onerous, but still requires some movement/modification of files locally (versus being able to have a config that points to a specific file that's known to exist in a location).

markelliot: Got it. What if we had a "pem" type where you could provide a single file which contained several PEM format certs for trust? Still only needs a type and a path.

nmiyake: Yup, that sounds good to me -- that would mean we support `jks` (current), `puppet` (puppet directory structure), and `pem` (single file/directory with pem files for trust). That should cover pretty much all the bases but only require the current set of settings, so no need to design a more generic/pluggable thing for hypothetical cases.

markelliot: Yep :)

nmiyake:
Initial implementation of single-config approach that implements PEM for trust store only. If this general approach seems OK, the following extensions/follow-ups are possible:
- Support case for PEM trust where path is a directory -- in this case, all PEM/DER certificates in the directory are added to trust store.
- Support keys in PEM/DER format (must be single file with both certificate chain and key). Support directory case for this as well.
- Add "puppet" type and support for trust and keys
  - For trust: all certificates in "certs" directory are trusted
  - For keys: for each key in "private_keys", add using the key and the corresponding cert in "certs"

May want to validate case for private keys -- not sure if there's actually demand for that. Also, directory-read is implemented for `PEM` for trust, that can be used for Puppet trust trivially as well (would just point cert to `$confdir/ssl/certs`).

If the current approach looks good, I would advocate for merging this and then implementing the "directory" feature for `PEM` trust and stopping there unless there are concrete requests on the key side.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
```diff
@@ -30,13 +30,19 @@
 @Value.Style(visibility = ImplementationVisibility.PACKAGE)
 public abstract class SslConfiguration {
 
-    private static final String DEFAULT_STORE_TYPE = "JKS";
+    public enum StoreType {
```
nmiyake, line 33: Typing as enum seems better/more elegant than using String.

Back-compat break for Java code, but not for JSON/serialization (since "JKS" and "PKCS12" enum constants have the same String representation as constants that were used before, objects are still serialization-compatible)

ssl-config/src/test/java/com/palantir/ssl/SerializationTests.java
```diff
@@ -41,4 +38,25 @@ public void testJsonSerDe() throws JsonParseException, JsonMappingException, Jso
...
+        ObjectMapper mapper = ObjectMappers.guavaJdk7();
+        SslConfiguration sslConfig = SslConfiguration.of(
+                TestConstants.CA_TRUST_STORE_PATH,
+                TestConstants.SERVER_KEY_STORE_JKS_PATH,
+                TestConstants.SERVER_KEY_STORE_JKS_PASSWORD);
+
+        assertThat(mapper.readValue(JSON_STRING, SslConfiguration.class), is(sslConfig));
+    }
+
+    private static final String JSON_STRING =
```
nmiyake, line 51: Test shows that serialization format has remained constant (`"JKS"` String is properly read from JSON and converted to enum)

markelliot:
Yeah, this is great.

If the switch statements get larger we should consider breaking the case blocks into methods for readability.

nmiyake:
Yup, makes sense -- wanted to get agreement on the general approach before spending too much time touching up :). Will clean up this commit a bit and then add the directory stuff as a follow-up.

nmiyake:
@markelliot @uschi2000 

Ready for review. Adds support for certificate-based trust -- if `trustStoreType` is set to `PEM`, then `trustStorePath` can be a path to a single X.509 certificate in PEM or DER format, or a directory that contains X.509 certificates in PEM or DER format. If the path is a directory, the directory must only contain certificate files (excluding hidden files, which are ignored).

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
```diff
@@ -0,0 +1,154 @@
...
+    private static Certificate readX509Certificate(Path certificateFilePath) {
+        try {
+            CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
+            Certificate certificate;
+            try (InputStream stream = Files.newInputStream(certificateFilePath)) {
+                certificate = certFactory.generateCertificate(stream);
+            }
+            return certificate;
+        } catch (GeneralSecurityException | IOException e) {
+            throw Throwables.propagate(e);
```
markelliot: Does this contain the file name?

markelliot:
Generally looks good. Only concern is if we produce good errors when files aren't valid certs (need to make sure errors have enough detail for users to debug the problem ... Ie the filename)

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -68,18 +63,30 @@ public static SSLContext createSslContext(SslConfiguration config) {
         }
     }
 
-    private static TrustManagerFactory createTrustManagerFactory(Path trustStorePath, String trustStoreType) {
+    private static TrustManagerFactory createTrustManagerFactory(
+            Path path,
+            SslConfiguration.StoreType trustStoreType) {
```
uschi2000, line 68: names should be symmetric

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -68,18 +63,30 @@ public static SSLContext createSslContext(SslConfiguration config) {
...
+        switch (trustStoreType) {
+            case JKS:
+            case PKCS12:
+                keyStore = KeyStores.loadKeyStore(trustStoreType.name(), path, Optional.<String>absent());
+                break;
+            case PEM:
+                keyStore = KeyStores.createTrustStoreFromCertificates(path);
+                break;
+            default:
+                throw new IllegalStateException("Unrecognized case: " + trustStoreType);
```
uschi2000: Unsupported trust store type: ...

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
```diff
@@ -0,0 +1,154 @@
...
+
+final class KeyStores {
+
+    private KeyStores() {
+    }
+
+    /**
+     * Returns a {@link KeyStore} created by loading the certificate or certificates specified by the provided path.
+     *
+     * @param path a path to a X.509 certificate in PEM or DER format, or to a directory containing certificate files.
```
uschi2000: can you fix the continuation indentation here?

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
```diff
@@ -0,0 +1,154 @@
...
+     * provided path. The returned store will not have any password.
+     */
+    static KeyStore createTrustStoreFromCertificates(Path path) {
+        try {
+            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
+            keyStore.load(null, null);
+
+            File pathFile = path.toFile();
+            if (pathFile.isDirectory()) {
+                File[] files = pathFile.listFiles(VISIBLE_FILE_FILTER);
```
uschi2000, line 55: does the order matter in which certs are added?

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
uschi2000: (similar to newKeyStoreWithEntry)

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
```diff
@@ -0,0 +1,154 @@
...
+     */
+    static KeyStore loadKeyStore(String storeType, Path path, Optional<String> password) {
+        try {
+            KeyStore keyStore = KeyStore.getInstance(storeType);
+            try (InputStream stream = Files.newInputStream(path)) {
+                char[] passwordChars = password.isPresent() ? password.get().toCharArray() : null;
+                keyStore.load(stream, passwordChars);
+            }
+            return keyStore;
+        } catch (GeneralSecurityException | IOException e) {
```
uschi2000, line 94: fine for this method to declare the exceptions since it's intended to be used inside this class only?

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static Certificate readX509Certificate(Path certificateFilePath) {
+        try {
+            CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
+            Certificate certificate;
+            try (InputStream stream = Files.newInputStream(certificateFilePath)) {
+                certificate = certFactory.generateCertificate(stream);
+            }
+            return certificate;
+        } catch (GeneralSecurityException | IOException e) {
```
uschi2000, line 146: can declare exception

ssl-config/src/main/java/com/palantir/remoting/ssl/SslConfiguration.java
uschi2000, line 33: cool

ssl-config/src/test/java/com/palantir/remoting/ssl/KeyStoresTests.java
```diff
@@ -0,0 +1,118 @@
...
+
+        KeyStore trustStore = KeyStores.createTrustStoreFromCertificates(certFolder.toPath());
+
+        assertThat(trustStore.size(), is(0));
+    }
+
+    @Test
+    public void testCreateTrustStoreFromDirectoryIgnoresHiddenFiles() throws IOException, GeneralSecurityException {
+        File certFolder = tempFolder.newFolder();
+
```
uschi2000: nit: can remove some space here and above

uschi2000:
looks good modulo a few nitpicks

uschi2000:
thanks Nick!

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
nmiyake: Good call -- fixed to include more descriptive exception message that includes file path.

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
nmiyake: Is there a good way to do this in IntelliJ? Ended up opening the project in Eclipse and formatting there :(

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
nmiyake, line 55: Don't think it should matter -- they're named entries, and fairly sure that for trust stores the provider will iterate through all of the certificates in the store when doing validation

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
nmiyake, line 94: This is called from `SslSocketFactories`, so code ends up being a bit cleaner by doing the try/catch here (could make exceptions checked and make the entire calling block in `SslSocketFactories` catch exception, but don't see much of an upside to making the change)

ssl-config/src/main/java/com/palantir/remoting/ssl/KeyStores.java
nmiyake, line 146: Keeping in this case to address Mark's suggestion -- allows throwing a more specific exception for error reading certificate specified by the path

ssl-config/src/test/java/com/palantir/remoting/ssl/KeyStoresTests.java
nmiyake: Done -- tightened up spacing in tests

nmiyake:
@uschi2000 @markelliot 

Addressed PR feedback

uschi2000:
+1 looks good, thanks Nick!

nmiyake:
Great, will rebase and squash

Issue #90, opened 2016-03-21
Run `certSetup.sh` as part of the build/test task


nmiyake:
The script generates all of the encryption material already, so it makes sense to remove the current encryption material from version control and just run the script as part of the build. Will serve as enforcement for the script continuing to work, and will make it such that change sets aren't polluted with change in security materials when the script is changed.

nmiyake:
Fixed by #91.

PR #91, opened 2016-03-21
Fix cert setup

nmiyake:


nmiyake:
@uschi2000 @markelliot 

Fix for #90

PR #94, opened 2016-04-04
Update Feign to 8.16.0

nmiyake:


nmiyake:
No breaking changes according to change log: https://github.com/Netflix/feign/blob/master/CHANGELOG.md

Main motivation is that 8.16 includes the `@HeaderMap` annotation I added, which will be useful in one of my client projects :)

PR #95, opened 2016-04-04
Replace QueryMapAwareContract with JAXRSWithHeaderAndQueryMapContract

nmiyake:
Delegating model made it impossible to support @HeaderMap and
@QueryMap annotation on methods that already had body parameters.
This changes creates a new JAXRSWithHeaderAndQueryMapContract
that replaces the JAXRSContract more directly to allow full
HeaderMap and QueryMap support for clients.


nmiyake:
Updated version of #87 now that `@HeaderMap` merged into Feign.

nmiyake:
@uschi2000 @markelliot 

Open to feedback/pushback on this -- can implement this on consumer client if needed (just means longer factory method call to remoting), but can see a case for having it as part of core as well.

pnepywoda:
ping - this PR seems reasonable. Nick: it's just a straight-up refactor using feign's new class right? (no added/removed functionality from what http-remoting already has)

nmiyake:
Yup, correct. And would be great if we could merge this -- would alleviate need to implement this same logic in every client library (see https://github.palantir.build/deployability/axiom-api/blob/develop/axiom-api/src/main/java/com/palantir/axiom/client/ProxyFactory.java)

pnepywoda:
+1

nmiyake:
@uschi2000 @markelliot

http-clients/src/main/java/feign/jaxrs/JAXRSWithHeaderAndQueryMapContract.java
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.reflect.Method;
+import java.util.Map;
+
+/**
+ * Contract that provides the functionality of {@link JAXRSContract} with full support for
+ * the {@link QueryMap} annotation. Allows the QueryMap annotation to be used in methods
+ * that also take a body parameter (such as {@link javax.ws.rs.POST or {@link javax.ws.rs.PUT}}).
+ */
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public final class JAXRSWithHeaderAndQueryMapContract extends Contract.BaseContract {
```
uschi2000: what's wrong with JaxRsWithHeaderAndQueryMapContract?

http-clients/src/test/java/feign/jaxrs/JAXRSWithHeaderAndQueryMapContractTests.java
```diff
@@ -0,0 +1,182 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+import org.junit.Rule;
+import org.junit.Test;
+
+@SuppressWarnings("checkstyle:abbreviationaswordinname")
+public final class JAXRSWithHeaderAndQueryMapContractTests {
```
uschi2000: as above

uschi2000:
let's stick with our naming convention (JaxRs vs JAXRS). otherwise: 👍

uschi2000:
(needs rebase, I believe)

http-clients/src/main/java/feign/jaxrs/JAXRSWithHeaderAndQueryMapContract.java
nmiyake: OK will go with that -- was doing `JAXRS` to match what Feign does, but since this is one of our classes fine with using our conventions

nmiyake:
Thanks, will fix naming, rebase and re-push

nmiyake:
@uschi2000 updated commit

uschi2000:
cheers, nick

nmiyake:
Thanks Rob! Will this be in a new RC for the current release, or in a new release?

PR #96, opened 2016-04-05
add service config

qinfchen:
https://palantir.quip.com/Ws4jA8hEmDTV


service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+public abstract class ServiceConfiguration {
+
+    public abstract Optional<BearerToken> apiToken();
+
+    public abstract Optional<SslConfiguration> security();
+
+    public abstract List<String> uris();
+
+    @Check
+    protected final void check() {
```
sixinli: hmmm you might want to allow empty because some service relies on empty = disable

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+@Immutable
+@JsonDeserialize(as = ImmutableServiceDiscoveryConfiguration.class)
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+public abstract class ServiceDiscoveryConfiguration {
+
+    @JsonProperty("apiToken")
+    public abstract Optional<BearerToken> defaultApiToken();
+
+    @JsonProperty("security")
+    public abstract Optional<SslConfiguration> defaultSecurity();
```
sixinli: should it match the name instead? defaultSecurity()->security()?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @param serviceName
+     *        the name of the service to be used to retrieve the respective {@link ServiceConfiguration}
+     * @returns the API token for a service based on the input service name
+     */
+    public final Optional<BearerToken> getApiToken(String serviceName) {
+        checkArgument(!Strings.isNullOrEmpty(serviceName));
+
+        Optional<ServiceConfiguration> maybeServiceConfig = getServiceConfiguration(serviceName);
+        if (!maybeServiceConfig.isPresent()) {
```
sixinli: if service is not present it should throw error? same question to the other methods

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Uses the service name as the key to retrieve the respective {@link ServiceConfiguration}.
+     *
+     * @param serviceName
+     *        the name of the service.
+     * @returns the {@link ServiceConfiguration} for a service based on the input service name
+     */
+    public final Optional<ServiceConfiguration> getServiceConfiguration(String serviceName) {
```
sixinli: probably easier to return the nullable thing here, and let it throw RuntimeException when you try to use it later?

qinfchen: you still need to check for null vs Optional.isPresent, prefer using latter to be consistent with `getSslConfiguration` and other functions.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: Prefer `defaultSecurity` as it is more explicit to the caller that this is the default security configuration

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: I don't have strong preference here. i can see it either way 1) since apiToken is optional,  there is no need to throw exception vs 2) the configuration doesn't exist for that service, doesn't make sense to return any api token

sixinli: agreeing more with 2) @_@

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
sixinli: serviceConfiguration is not optional though? i see it different from sslConfiguration which is optional because there's a default hmm

qinfchen: right, but Optional type doesn't equate to "optional", it just says that the service configuration can be absent for the given service name. https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: fixed.

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen: fixed

qinfchen:
@uschi2000 @markelliot ?

gradle/libs.gradle
```diff
@@ -23,6 +23,7 @@ project.ext.libs = [
     'mockwebserver': 'com.squareup.okhttp:mockwebserver:2.7.0',
     'okhttp': 'com.squareup.okhttp:okhttp:2.5.0',
     'retrofit': 'com.squareup.retrofit:retrofit:1.9.0',
+    'tokens': 'com.palantir.tokens:auth-tokens:0.2.3',
```
nmiyake: nit, but imports like these should be alphabetized ('tokens' after 'slf4j')

service-config/src/test/java/com/palantir/config/service/example/AuthConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+import org.eclipse.jetty.util.security.Password;
+import org.hibernate.validator.constraints.NotEmpty;
+import org.immutables.value.Value;
+import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Style;
+
+/**
+ * Configuration for the Authentication.
+ */
+@Value.Immutable
```
nmiyake: The other classes in this commit use `@Immutable` directly (omitting the "`Value`"). Should stick with a single style for consistency.

gradle/libs.gradle
qinfchen: yes, somehow missed "slf4j" :)

service-config/src/test/resources/configs/example.yml
```diff
@@ -0,0 +1,74 @@
...
+      port: 8443
+      keyStorePassword: keystore
+      keyStorePath: src/test/resources/keyStore.jks
+      excludedProtocols: [SSLv3]
+      supportedProtocols: [TLSv1, TLSv1.1, TLSv1.2]
+      validateCerts: false
+  adminConnectors: []
+  applicationContextPath: /example
+  rootPath: /api/*
+
```
nmiyake: Remove trailing whitespace (keep newline, but it should be blank -- there's currently a space there)

service-config/src/test/java/com/palantir/config/service/example/AuthConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public abstract ServiceConfiguration getService();
+
+    @JsonProperty("oauth2")
+    public abstract OAuth2Configuration getOAuth2();
+
+    @Immutable
+    @JsonDeserialize(as = ImmutableOAuth2Configuration.class)
+    @Style(visibility = Style.ImplementationVisibility.PACKAGE)
+    abstract static class OAuth2Configuration {
```
nmiyake: Is there a reason that this is declared as an inner static class instead of as its own file?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Uses the provided service name as the key to retrieve the API token for the respective
+     * {@link ServiceConfiguration}. If the API token is not defined in the {@link ServiceConfiguration}, the default
+     * API token will be returned.
+     *
+     * @param serviceName
+     *        the name of the service to be used to retrieve the respective {@link ServiceConfiguration}
+     * @returns the API token for a service based on the input service name
+     */
+    public final Optional<BearerToken> getApiToken(String serviceName) {
```
nmiyake, line 86: This is more of a philosophical question/issue, but this methods and the ones that follow it are all for logic for this class. There's a question as to whether Immutables classes should just represent JSON/POJOs and not provide any extra logic (extra logic provided by other helper classes/classes that accept the objects as input), or whether bundling logic like this makes sense. @uschi2000 may have thoughts on this.

service-config/src/test/java/com/palantir/config/service/example/AuthConfiguration.java
qinfchen: No particular reason other than this is used only by this class.

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+import com.palantir.tokens.auth.BearerToken;
+import java.util.List;
+import java.util.Set;
+import javax.net.ssl.SSLSocketFactory;
+import org.immutables.value.Value.Check;
+import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Style;
+
+@Immutable
+@JsonDeserialize(as = ImmutableServiceConfiguration.class)
```
nmiyake, line 28: Do the classes need to be annotated with `@JsonSerialize` as well, or is that not needed?

nmiyake:
I don't see an issue for this -- is there a Quip or some other documentation that provides a BLUF of what this change is meant to do/address?

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen, line 28: This is mainly used for to deserialize the yml file, i don't see any use cases for serializing it.

qinfchen:
here it is https://palantir.quip.com/Ws4jA8hEmDTV

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen, line 86: It was discussed in the quip doc linked below that we want to have service specific token if they are present, but default to the root level configurations if not. Hence bundling the logic here to make sure the behavior is consistent across consuming apps

nmiyake:
Awesome, thanks!

service-config/build.gradle
```diff
@@ -0,0 +1,13 @@
+apply from: "$rootDir/gradle/libs.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+dependencies {
+    compile project(':http-clients')
```
uschi2000: I am surprised that the dependency goes this way. I suspect http-clients should depend on on service-config instead?

qinfchen: updated. Original thought was that consuming apps just need to pull in service config and not to worry about feign client or how to build the it.

qinfchen:
need to remove http-client dependency from ssl-config, or IDE will complain about circular dependencies even if it is test dependency

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,42 @@ public static FeignClientFactory of(
...
+            ServiceDiscoveryConfiguration discoveryConfig, String serviceName, Class<T> serviceClass) {
+
+        Optional<ServiceConfiguration> maybeServiceConfig = discoveryConfig.getServiceConfiguration(serviceName);
+        if (!maybeServiceConfig.isPresent()) {
+            throw new RuntimeException("Unable to find the configuration for " + serviceName + ".");
+        }
+
+        return createProxies(maybeServiceConfig.get(), discoveryConfig.defaultSecurity(), serviceClass);
+    }
+
```
sixinli, line 164: dont need this extra line here

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,42 @@ public static FeignClientFactory of(
...
+        Set<String> uris = Sets.newHashSet(serviceConfig.uris());
+        return FeignClients.standard().createProxy(socketFactory, uris, serviceClass);
+    }
+
+    /**
+     * Constructs a dynamic proxy for the specified type using the {@link ServiceConfiguration} obtained from the
+     * supplied {@link ServiceDiscoveryConfiguration} for the specified service name.
+     */
+    public <T> T createProxies(
+            ServiceDiscoveryConfiguration discoveryConfig, String serviceName, Class<T> serviceClass) {
```
sixinli: each param in its own line?

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,42 @@ public static FeignClientFactory of(
...
+
+        Set<String> uris = Sets.newHashSet(serviceConfig.uris());
+        return FeignClients.standard().createProxy(socketFactory, uris, serviceClass);
+    }
+
+    /**
+     * Constructs a dynamic proxy for the specified type using the {@link ServiceConfiguration} obtained from the
+     * supplied {@link ServiceDiscoveryConfiguration} for the specified service name.
+     */
+    public <T> T createProxies(
```
sixinli: why do we prefer it this way? that means i have 2 dependency instead of 1 when i need to create client?

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
qinfchen: see line 182. If it fits in separate line, prefer keeping it to be consistent

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
qinfchen: I had a similar opinion for the previous approach, it is really convenient, but we would have a  strong coupling between feign and service config.  Say we decide to use other client instead of feign,  it is harder to decouple them. Oh, take Retrofit client for example, we also need to be able to create retrofit client using the service config.  Also Service Config should not be changed often in the future.

sixinli: Cool! Thanks for explaining :)

uschi2000:
ready for another review, @qinfchen ?

sixinli:
:+1:

qinfchen:
Yeah,:)

Sent from my iPhone

On Apr 6, 2016, at 11:22 PM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

ready for another review, @qinfchenhttps://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_qinfchen&d=CwMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=2LK6Jot8d42Knk04ei1_FPh1f_GvosvCO-13ihGNoi8&m=jRqkQPWOiN4kUMolG8K243sfynIsbnN1plN9r4ztmFM&s=ApvPcq9ffdvnyCJzI_Kyt0BNLTyb_P8Rd9Dd3clPajg&e= ?

uschi2000:
OK, will look tomorrow.

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+public abstract class ServiceConfiguration {
+
+    public abstract Optional<BearerToken> apiToken();
+
+    public abstract Optional<SslConfiguration> security();
+
+    public abstract List<String> uris();
```
uschi2000, line 45: why is this a List rather than a Set? you seem to turn it into a Set as soon as you see it.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param serviceName
+     *        the name of the service to be used to retrieve the respective {@link ServiceConfiguration}
+     * @return the {@link ServiceConfiguration} for a service based on the input service name
+     */
+    public final Optional<SslConfiguration> getSslConfiguration(String serviceName) {
+        checkArgument(!Strings.isNullOrEmpty(serviceName));
+
+        Optional<ServiceConfiguration> maybeServiceConfig = getServiceConfiguration(serviceName);
+        if (!maybeServiceConfig.isPresent()) {
```
uschi2000: why is this not symmetric with getSslConfiguration?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Uses the service name as the key to retrieve the respective {@link ServiceConfiguration}.
+     *
+     * @param serviceName
+     *        the name of the service.
+     * @return the {@link ServiceConfiguration} for a service based on the input service name
+     */
+    public final Optional<ServiceConfiguration> getServiceConfiguration(String serviceName) {
+        checkArgument(!Strings.isNullOrEmpty(serviceName));
```
uschi2000: redundant check

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Uses the service name as the key to retrieve the respective {@link ServiceConfiguration}.
+     *
+     * @param serviceName
+     *        the name of the service.
+     * @return the {@link ServiceConfiguration} for a service based on the input service name
+     */
+    public final Optional<ServiceConfiguration> getServiceConfiguration(String serviceName) {
```
uschi2000: private

http-clients/src/test/java/com/palantir/remoting/http/serviceconfig/ExampleApp.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.palantir.config.service.ServiceConfiguration;
+import com.palantir.config.service.ServiceDiscoveryConfiguration;
+import io.dropwizard.Application;
+import io.dropwizard.setup.Environment;
+
+/**
+ * Used to test {@link ServiceConfiguration} and {@link ServiceDiscoveryConfiguration}.
+ */
+public final class ExampleApp extends Application<ExampleAppConfiguration> {
```
uschi2000: the test code is pretty heavy here. can we re-use some code from other projects by making a shared test-only project?

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,41 @@ public static FeignClientFactory of(
...
+            ServiceConfiguration serviceConfig,
+            Optional<SslConfiguration> fallbackSslConfiguration,
+            Class<T> serviceClass) {
+
+        Optional<SSLSocketFactory> socketFactory = Optional.absent();
+        Optional<SslConfiguration> sslConfig = serviceConfig.security().or(fallbackSslConfiguration);
+
+        if (sslConfig.isPresent()) {
```
uschi2000, line 156: can probably use the functional primitives of Optional (transform/map)

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,41 @@ public static FeignClientFactory of(
     }
 
     /**
+     * Constructs a dynamic proxy for the specified type using the supplied {@link ServiceConfiguration}. If the SSL
+     * configuration is not present in the {@link ServiceConfiguration}, the fallback SSL configuration will be used.
+     */
+    public <T> T createProxies(
+            ServiceConfiguration serviceConfig,
+            Optional<SslConfiguration> fallbackSslConfiguration,
```
uschi2000: do you really need this? there is already a fallback/default in the config.

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,41 @@ public static FeignClientFactory of(
...
+
+        Optional<SSLSocketFactory> socketFactory = Optional.absent();
+        Optional<SslConfiguration> sslConfig = serviceConfig.security().or(fallbackSslConfiguration);
+
+        if (sslConfig.isPresent()) {
+            socketFactory = Optional.of(SslSocketFactories.createSslSocketFactory(sslConfig.get()));
+        }
+
+        Set<String> uris = Sets.newHashSet(serviceConfig.uris());
+        return FeignClients.standard().createProxy(socketFactory, uris, serviceClass);
```
uschi2000: should use methods from this FeignClientFactory instance rather than FeignClients.standard()

http-clients/src/test/java/com/palantir/remoting/http/serviceconfig/AuthConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value.Style;
+
+/**
+ * Configuration for the Authentication.
+ */
+@Immutable
+@JsonDeserialize(as = ImmutableAuthConfiguration.class)
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+public abstract class AuthConfiguration {
```
uschi2000: this is very MP-specific, right? at the minimum, this needs a lot of comments. I wonder if http-remoting is even the right place for this?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: what do you mean? having the default?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: it could be useful to obtain the configuration for the specific service for the consuming app.

http-clients/src/test/java/com/palantir/remoting/http/serviceconfig/AuthConfiguration.java
qinfchen: this is for test only in order to demonstrate a real world use case scenario presenting what Mark, Punya, and Si, discussed in the linked Quip doc. and the yaml file has a lot of comments.

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
qinfchen: there is no defaut config for ServiceConfiguration. Default config is from ServiceDiscoveryConfiguration

http-clients/src/test/java/com/palantir/remoting/http/serviceconfig/ExampleApp.java
qinfchen: will look into it. not much code can be shared. But will look into shared test-only project as a separate task

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen, line 45: Using Set due to FeignClientFactory and that is impl detail. We want use list so we can specify priorities in which the urls will be used.

qinfchen:
Friendly ping :)

uschi2000:
there are a few open/outstanding comments, @qinfchen . could you look at those first?

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +139,42 @@ public static FeignClientFactory of(
...
+    }
+
+    /**
+     * Constructs a dynamic proxy for the specified type using the {@link ServiceConfiguration} obtained from the
+     * supplied {@link ServiceDiscoveryConfiguration} for the specified service name.
+     */
+    public <T> T createProxies(
+            ServiceDiscoveryConfiguration discoveryConfig, String serviceName, Class<T> serviceClass) {
+
+        Optional<ServiceConfiguration> maybeServiceConfig = discoveryConfig.getServiceConfiguration(serviceName);
```
uschi2000: this seems roundabout. how about:

```
ServiceConfiguration config = Preconditions.checkNotNull(discoveryConfig.services().get(serviceName), "...");
```

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
```diff
@@ -0,0 +1,102 @@
...
+    abstract static class AuthConfiguration {
+
+        public abstract Optional<String> getFailureEndpoint();
+
+        @SuppressWarnings("checkstyle:designforextension")
+        public Duration getTokenCache() {
+            return Duration.seconds(30L);
+        }
+
+        public abstract boolean legacyModeEnabled();
```
uschi2000: what does this mean?

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
```diff
@@ -0,0 +1,102 @@
...
+    public ServiceDiscoveryConfiguration getServiceDiscoveryConfiguration() {
+        return this.serviceDiscoveryConfig;
+    }
+
+    @Immutable
+    @JsonDeserialize(as = ImmutableAuthConfiguration.class)
+    @Style(visibility = Style.ImplementationVisibility.PACKAGE)
+    abstract static class AuthConfiguration {
+
+        public abstract Optional<String> getFailureEndpoint();
```
uschi2000: what does this mean?

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
```diff
@@ -0,0 +1,102 @@
...
+        @JsonProperty("oauth2")
+        public abstract OAuth2Configuration getOAuth2();
+    }
+
+    @Immutable
+    @JsonDeserialize(as = ImmutableOAuth2Configuration.class)
+    @Style(visibility = Style.ImplementationVisibility.PACKAGE)
+    abstract static class OAuth2Configuration {
+
+        @NotEmpty
```
uschi2000: why do some fields have validations attached and others do not?

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
uschi2000: ah, this is test code, nevermind

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
qinfchen: micmic multipass auth config. same as the one below.

http-clients/src/test/java/com/palantir/remoting/http/ServiceConfigTestAppConfig.java
qinfchen: will remove them. not needed for tests

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+public abstract class ServiceDiscoveryConfiguration {
+
+    @JsonProperty("apiToken")
+    public abstract Optional<BearerToken> defaultApiToken();
+
+    @JsonProperty("security")
+    public abstract Optional<SslConfiguration> defaultSecurity();
+
+    public abstract Map<String, ServiceConfiguration> services();
```
uschi2000: needs docs

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
qinfchen: hmm. not sure i like this style. that is a lot of stuff happening in one line.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Uses the provided service name as the key to retrieve the API token for the respective
+     * {@link ServiceConfiguration}. If the API token is not defined in the {@link ServiceConfiguration}, the default
+     * API token will be returned.
+     *
+     * @param serviceName
+     *        the name of the service to be used to retrieve the respective {@link ServiceConfiguration}
+     * @return the API token for a service based on the input service name
+     */
+    public final Optional<BearerToken> getApiToken(String serviceName) {
```
uschi2000, line 86: this API seems broken, since: `config.getApiToken("foo") != config.getServiceConfiguration("foo").getApiToken()`

http-clients/src/test/resources/service-config-example.yml
```diff
@@ -0,0 +1,73 @@
+auth:
+  service: # just a standard ServiceConfig block
+    security:
+      trustStorePath: src/test/resources/trustStore.jks
+    uris:
+      - https://some.internal.url:8643/service/api
+  oauth2:
```
uschi2000: why do we need this?

http-clients/src/test/resources/service-config-example.yml
```diff
@@ -0,0 +1,73 @@
...
+        keyStoreType: JKS                                 # optional (defaults to JKS)
+
+    goodbye: # just a standard ServiceConfig block
+      security:
+        trustStorePath: src/test/resources/trustStore.jks
+      uris:
+        - https://localhost:8443/example/api
+
+# information for the front-end
+clients:
```
uschi2000: why do we need this?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen, line 86: hmm... good point, not sure what going to do about this. Ideally config.getServiceConfiguration("foo").getApiToken() should never be called. 

so don't expose getServiceConfiguration?

http-clients/src/test/resources/service-config-example.yml
qinfchen: just an example for the test and shows a working version of the config similar to the real world config

qinfchen:
updated for latest Rob's comments.

http-clients/src/test/resources/service-config-example.yml
uschi2000: ping. why do we need MP configuration for these tests?

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
uschi2000, line 28: is there a use-case for sending configuration over the wire?

http-clients/src/test/resources/service-config-example.yml
qinfchen: we don't necessary need them, but i want include a complete set of configuration to demonstrate it is working as expected and also gives people a idea how service discovery configuration should look like

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
uschi2000, line 86: can you instead "complete" or "resolve" the per-service config by filling in the absent pieces from the top-level defaults?

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen, line 28: i don't think so, you would have to send the ssl configuration over the wire and that doesn't seem correct

http-clients/src/test/resources/service-config-example.yml
uschi2000: `getOAuth2()` isn't called anywhere, so this is redundant. the whole `auth` block is dangling mid-air without purpose.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen: I made some changes, can you take a look at it?

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
uschi2000, line 28: ok

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
qinfchen, line 86: Updated

http-clients/src/test/resources/service-config-example.yml
uschi2000: ping. please remove these unused configuration items.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @JsonProperty("services")
+    abstract Map<String, ServiceConfiguration> originalServices();
+
+    /**
+     * A map of {@code serviceName}s as the keys and their respective {@link ServiceConfiguration}s as the values.
+     */
+    @Lazy
+    @SuppressWarnings("checkstyle:designforextension")
+    public Map<String, ServiceConfiguration> getServices() {
```
uschi2000, line 63: last request: can you add a test verifying that `getServices().get("foo").getApiToken()` respects fallbacks?

qinfchen, line 63: There should be a test in the serviceDiscoveryConfigurationTests already

Sent from my iPhone

On Apr 14, 2016, at 7:46 PM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

In service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.javahttps://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_http-2Dremoting_pull_96-23discussion-5Fr59809953&d=CwMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=2LK6Jot8d42Knk04ei1_FPh1f_GvosvCO-13ihGNoi8&m=kiU3Oy4nbMYyCIIqb3n-SWDN0IdlynMFSU4Xb9-Q0vk&s=aAIblTFesqfago_sC_KIP4KF_pUhomjxfyJDrXZ5_V0&e=:

> -    /**
> -     \* Fallback SSL Configuration to be used if the service specific SSL configuration is not defined in the
> -     \* {@link ServiceConfiguration}.
> -     */
> -    @JsonProperty("security")
> -    public abstract Optional<SslConfiguration> getDefaultSecurity();
>   +
> -    @JsonProperty("services")
> -    abstract Map<String, ServiceConfiguration> originalServices();
>   +
> -    /**
> -     \* A map of {@code serviceName}s as the keys and their respective {@link ServiceConfiguration}s as the values.
> -     */
> -    @Lazy
> -    @SuppressWarnings("checkstyle:designforextension")
> -    public Map<String, ServiceConfiguration> getServices() {

last request: can you add a test verifying that getServices().get("foo").getApiToken() respects fallbacks?

uschi2000, line 63: ah, missed that. thanks.

uschi2000:
@markelliot , anything from your side?

sixinli:
ping, is this ready?

uschi2000:
Thanks, @qinfchen !

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -134,6 +140,29 @@ public static FeignClientFactory of(
     }
 
     /**
+     * Constructs a dynamic proxy for the specified type using the {@link ServiceConfiguration} obtained from the
+     * supplied {@link ServiceDiscoveryConfiguration} for the specified service name.
+     */
+    public <T> T createProxies(
```
pnepywoda, line 146: createProxy (singular)

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+import org.immutables.value.Value.Style;
+
+/**
+ * Configuration class that contains a map of {@code serviceName}s and their respective {@link ServiceConfiguration}s
+ * and provides utility methods for apps to build service clients.
+ */
+@Immutable
+@JsonDeserialize(as = ImmutableServiceDiscoveryConfiguration.class)
+@Style(get = "*", visibility = Style.ImplementationVisibility.PACKAGE)
+public abstract class ServiceDiscoveryConfiguration {
```
pnepywoda, line 39: Because of how we customize the discovery block in https://github.palantir.build/elements/foundry/blob/develop/foundry-build/var/conf/foundry-build.yml#L41 I need a way to construct a SDC programmatically, like a .of(): https://github.palantir.build/elements/foundry/blob/develop/foundry-services/src/main/java/com/palantir/foundry/services/config/ServiceDiscoveryConfiguration.java#L24

pnepywoda, line 39: I also use .of() in a bunch of foundry tests....could you add one to this?

qinfchen:
will have PR soon to address Paul's comments

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Uses the service name as the key to retrieve the {@link SslConfiguration} for the respective
+     * {@link ServiceConfiguration}. If the {@link SslConfiguration} is not defined in the {@link ServiceConfiguration},
+     * the default {@link SslConfiguration} will be returned.
+     *
+     * @param serviceName
+     *        the name of the service to be used to retrieve the respective {@link ServiceConfiguration}
+     * @return the {@link ServiceConfiguration} for a service based on the input service name
+     */
+    public final Optional<SslConfiguration> getSslConfiguration(String serviceName) {
```
ellisjoe, line 103: any reason not to include `Optional<List<String>> getUris(String serviceName)`?

qinfchen, line 103: not really, the original impl has logic to use default apiToken and security if the service specific ones aren't present. i am fine with adding that for uris.

uschi2000, line 103: +1

ellisjoe, line 103: cool. would be useful for us to have it

qinfchen, line 103: Here is the PR that includes the change, https://github.com/palantir/http-remoting/pull/106

Issue #98, opened 2016-04-11
Javadoc error in WebPreconditions


nmiyake:
When I try to build the latest version locally using `./gradlew build`, it fails with the following exception:

```
/Volumes/git/palantir/http-remoting/http-servers/src/main/java/com/palantir/remoting/http/server/WebPreconditions.java:44: warning - Tag @link: can't find checkNotEmpty(String) in com.palantir.remoting.http.server.WebPreconditions
1 error
22 warnings
:http-servers:javadoc FAILED

FAILURE: Build failed with an exception.
```

Interestingly, in Circle it seems that this same issue is a warning rather than an error. However, the reference does seem to be incorrect. Should figure out why it's not failing in Circle as a separate thing, but underlying issue should be fixed in the mean time.

nmiyake:
@jtamayo

nmiyake:
Opened PR for fix, but tagged you on the ticket for SA. Does `./gradlew build` fail for you locally with this issue, or just give a warning? Curious as to why it seems to be a failure for me locally but only a warning in Circle.

nmiyake:
Fixed by #99

nmiyake:
FWIW, figured out why this was failing locally but succeeding in Circle -- the Circle environment for this project is Java 7, and this Javadoc issue is just a warning in Java 7. However, most of us run Java 8 locally, and since this is an error in Java 8 we would get a failure.

Think this is an edge case we'll just have to live with -- although `gradle-docker-test-runner` has the capability to test this, it wouldn't catch this with its standard configuration since it shares the artifacts that were built externally for perf purposes. Think this should be a pretty rare case, though.

PR #99, opened 2016-04-11
Fix Javadoc so that it is valid

nmiyake:
Update method reference to be a method that exists.
Fixes issue #98.


nmiyake:
@uschi2000 @markelliot

PR #102, opened 2016-04-12
Update Baseline version to 0.3.0

nmiyake:
Fixes issue where working directory for unit tests
is not properly set in IDEA.


nmiyake:
@uschi2000 @markelliot 

Fixes super annoying unit test working directory issue and also fixes some illegal imports that were found

PR #105, opened 2016-04-16
add option to specify timeout when overridding objectmapper

robert3005:
Seems odd that we don't expose that. Need it to integrate with contour.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -118,6 +118,13 @@ public final class FeignClients {
      * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
      */
     public static FeignClientFactory withMapper(ObjectMapper mapper) {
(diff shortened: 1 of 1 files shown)


http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -118,6 +118,13 @@ public static FeignClientFactory vanilla(Request.Options timeoutOptions) {
      * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
      */
     public static FeignClientFactory withMapper(ObjectMapper mapper) {
+        return withMapper(DEFAULT_TIMEOUT_OPTIONS, mapper);
+    }
+
+    /**
+     * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
+     */
+    public static FeignClientFactory withMapper(Request.Options timeoutOptions, ObjectMapper mapper) {
```
uschi2000: can you reverse the arguments?

robert3005: Sure, here you go.

uschi2000:
cheers Robert.

PR #106, opened 2016-04-19
add builder to service config and improve APIs

qinfchen:
- added builder to both ServiceConfiguration and ServiceDiscoveryConfiguration classes for tests
- renamed createProxies to createProxy for building feign client with service config
- removed get prefix for variables be consistent with the other class
- renamed getSslConfiguration to getSecurity for consistency
- removed @Value.Parameter in favor of Builder since `of()` requires all class variables

### DIFF
--- a/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
@@ -26,7 +26,7 @@ import org.immutables.value.Value.Style;
 
 @Immutable
 @JsonDeserialize(as = ImmutableServiceConfiguration.class)
-@Style(allParameters = true, visibility = Style.ImplementationVisibility.PACKAGE)
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
 public abstract class ServiceConfiguration {
 
     /**
@@ -43,4 +43,23 @@ public abstract class ServiceConfiguration {
      * A list of service URIs.
      */
     public abstract List<String> uris();
+
+    // hides implementation details
+    public static Builder builder() {
+        return ImmutableServiceConfiguration.builder();
+    }
+
+    // hides implementation details
+    public interface Builder {
+
+        Builder apiToken(BearerToken apiToken);
+
+        Builder security(SslConfiguration security);
+
+        Builder uris(Iterable<String> uris);
+
+        Builder from(ServiceConfiguration otherConfig);
+
+        ServiceConfiguration build();
+    }
 }
--- a/service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
+++ b/service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
@@ -16,6 +16,7 @@
 
 package com.palantir.config.service;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.common.base.Optional;
@@ -24,6 +25,7 @@ import com.palantir.remoting.ssl.SslConfiguration;
 import com.palantir.tokens.auth.BearerToken;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import org.immutables.value.Value.Immutable;
 import org.immutables.value.Value.Lazy;
@@ -35,7 +37,7 @@ import org.immutables.value.Value.Style;
  */
 @Immutable
 @JsonDeserialize(as = ImmutableServiceDiscoveryConfiguration.class)
-@Style(get = "*", visibility = Style.ImplementationVisibility.PACKAGE)
+@Style(visibility = Style.ImplementationVisibility.PACKAGE)
 public abstract class ServiceDiscoveryConfiguration {
 
     /**
@@ -43,14 +45,14 @@ public abstract class ServiceDiscoveryConfiguration {
      * {@link ServiceConfiguration}.
      */
     @JsonProperty("apiToken")
-    public abstract Optional<BearerToken> getDefaultApiToken();
+    public abstract Optional<BearerToken> defaultApiToken();
 
     /**
      * Fallback SSL Configuration to be used if the service specific SSL configuration is not defined in the
      * {@link ServiceConfiguration}.
      */
     @JsonProperty("security")
-    public abstract Optional<SslConfiguration> getDefaultSecurity();
+    public abstract Optional<SslConfiguration> defaultSecurity();
 
     @JsonProperty("services")
     abstract Map<String, ServiceConfiguration> originalServices();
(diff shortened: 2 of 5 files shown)


qinfchen:
@uschi2000

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -43,4 +43,23 @@
...
      */
     public abstract List<String> uris();
+
+    // hides implementation details
+    public static Builder builder() {
+        return ImmutableServiceConfiguration.builder();
+    }
+
+    // hides implementation details
+    public interface Builder {
```
uschi2000, line 53: can you try using this pattern instead:

```
public static Builder builder() {
    return new Builder();
}

public static final class Builder extends ImmutableFoo.Builder {}
```

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -100,11 +104,30 @@
...
         return serviceConfig.security();
     }
+
+    // hides implementation details
+    public static Builder builder() {
+        return ImmutableServiceDiscoveryConfiguration.builder();
+    }
+
+    // hides implementation details
+    public interface Builder {
```
uschi2000, line 134: as above

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -60,15 +61,18 @@
...
         Map<String, ServiceConfiguration> intializedServices = new HashMap<String, ServiceConfiguration>();
 
         for (Map.Entry<String, ServiceConfiguration> entry : originalServices().entrySet()) {
-            ServiceConfiguration serviceConfig = entry.getValue();
-            Optional<BearerToken> apiToken = serviceConfig.apiToken().or(getDefaultApiToken());
-            Optional<SslConfiguration> security = serviceConfig.security().or(getDefaultSecurity());
-            intializedServices.put(entry.getKey(),
-                    ImmutableServiceConfiguration.of(apiToken, security, serviceConfig.uris()));
+            ServiceConfiguration originalConfig = entry.getValue();
+            Optional<BearerToken> apiToken = originalConfig.apiToken().or(defaultApiToken());
```
uschi2000: I think this would be more clear as follows:

```
ServiceConfiguration configWithDefaults = ServiceConfiguration.builder()
    .apiToken(serviceConfig.apiToken().or(defaultApiToken())
    .security(serviceConfig.security().or(defaultSecurity())
    .uris(serviceConfig.uris())
    .build();
```

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen, line 53: Unfortunately, the patten you suggested doesn't hide the implementation details.  We hit this when we were implementing the web-security bundle. see more details here 
https://github.com/palantir/dropwizard-web-security/pull/29

nmiyake, line 53: +1 to what's currently in the commit -- I use the same style in the axiom-api project as well since it seemed like that's what was recommended by the Immutables project

qinfchen, line 53: Would love to collect some best practices on immutables and add them in baseline. I created a quip doc here, please feel free to update it. https://palantir.quip.com/dDUfABPSL7eP

uschi2000, line 53: What exactly do you mean by "implementation details"?

qinfchen, line 53: the immutables library ends up being exposed through the API if you do not define the Builder interface for a specific class. this means you are requiring consuming applications to include the library or they'll throw an exception at build time.

```
error: incompatible types: Optional<ImmutableSomeClass> cannot be converted to Optional<SomeClass>
```

uschi2000, line 53: I don't understand why Foundry doesn't hit this problem. @pnepywoda ?

qinfchen, line 53: we ran into this when the service didn't have immutables yet

uschi2000:
I'm fine with this, although I find this immutables pattern a bit silly.

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
pnepywoda, line 53: Uh, yeah i don't understand the reason for "hiding impl details" either. You still need to publish the Immutable\* classes since that's the actual implementation, the only difference is that Immutable*.Builder is public....so what?

pnepywoda, line 53: Foundry has used this patten forever: 

```
    public static Builder builder() {
        return new Builder();
    }

    public static final class Builder extends ImmutableJobResult.Builder {}
```

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
pnepywoda, line 134: ditto

pnepywoda:
yeah I'm ok with it too, but also agree the immutable pattern is silly and unnecessary

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -43,4 +43,23 @@
...
+
+        Builder apiToken(BearerToken apiToken);
+
+        Builder security(SslConfiguration security);
```
pnepywoda, line 57: I didn't notice this until I tried using it - this should also have
Builder security(Optional<SslConfiguration> security)
so we can directly pass one variable to another. Immutables does this for you but by making your own interface you bypass the auto-generation.

Can we please just not have this "hide implementation interface" class?

pnepywoda, line 57: @qinfchen

qinfchen, line 57: for immutable builders, you can omit the optional fields if they are absent.  I don't see the need to pass in a Optional typed value when you only need to set the value when it exists. 

Sure, lets use the old pattern if this is not convinient for other users. I will investigate/try to understand the problem more and update back with results.

pnepywoda, line 57: with how things currently are I have to do silly things like https://github.palantir.build/elements/foundry/pull/771/files#diff-f18eac1da998e57c44fff75ab5bc4fdaR67 , also, the method is called "originalServices" instead of "services" which doesn't match. All the more reason to NOT have this overriding/hiding interface

qinfchen, line 57: I am traveling atm, so I don't have the access to link. :(. Want to clarify on the "originalServices" name that the services you set via the builder will not be the same as the services you get from "getServices" because we inject default tokens and security config for service config that don't have them defined. Hence different names. Open to suggestions

pnepywoda, line 57: Link Summary: it's messy ;)

So id recommend less magic. Have getservices return exactly setservices.
And of course do the special default thing in the wrapping class

uschi2000:
Any chance we can sort this out today? Would like to make a 0.11.0 release asap

qinfchen:
What Paul suggested was the original impl, we changed it so that the api is consistent for the following cases:

DiscoveryConfig.getServices().get("service1").getApiToken()
DiscoveryConfig.getApiToken("service1")

Sent from my iPhone

On Apr 22, 2016, at 4:24 PM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:

Any chance we can sort this out today? Would like to make a 0.11.0 release asap

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
nmiyake, line 53: Just played around with this more locally, and can confirm what @uschi2000 and @pnepywoda said -- basically, the `extends Immutable*.Builder {}` approach should work fine. @qinfchen , what was the scenario where this didn't work? I tried using the approach described and consuming from an external project without Immutables and everything still compiled/worked.

Philosophically, the re-defining approach is cleaner because it does hide the impl details. For example, if you change the templating so that the generated immutable classes have a different naming scheme, the `extends Immutable*.Builder {}` approach will break in source, while the former approach will not. Also, if there's a situation where the source is published for the interface/abstract class but not for the generated class, then defining the interface in the declaring class makes more sense.

Ultimately it seems like either approach works fine in almost all cases, so I think this largely comes down to a matter of taste.

nmiyake, line 53: Sorry to keep harping on this thread, but adding some notes based on what I just figured out myself.

The approach that Rob and Paul suggested never occurred to me because originally the generated `Builder` class is final, so never thought that subclassing it was a possibility. It seems that you need to include the `Builder builder()` method and `public static final class Builder extends Immutable*.Builder {}` code in the original in order for the generated code to work properly. However, once that's done, everything does seem to work for me.

This definitely seems more elegant from a code standpoint, so I'm probably going to move to this approach for my own code. Thanks for the tip!

uschi2000, line 53: Any volunteer for making this change here?

nmiyake, line 53: Sure, PR at https://github.com/palantir/http-remoting/pull/109.

-Nick

PR #108, opened 2016-04-26
add configurable timeouts to serviceconfiguration

robert3005:
Tests are incoming. Wanted to throw it out here since there were multiple parties asking for it.

### DIFF
--- a/service-config/build.gradle
+++ b/service-config/build.gradle
@@ -4,6 +4,7 @@ apply from: "$rootDir/gradle/publish.gradle"
 dependencies {
     compile project(':ssl-config')
     compile libs.tokens
+    compile libs.dropwizard.utils
 
     processor libs.immutables
 
--- a/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.common.base.Optional;
 import com.palantir.remoting.ssl.SslConfiguration;
 import com.palantir.tokens.auth.BearerToken;
+import io.dropwizard.util.Duration;
 import java.util.List;
 import org.immutables.value.Value.Immutable;
 import org.immutables.value.Value.Style;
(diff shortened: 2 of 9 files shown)


service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -40,26 +40,23 @@
     public abstract Optional<SslConfiguration> security();
 
     /**
+     * Connect timeout for requests in ms.
+     */
+    public abstract Optional<Integer> connectTimeoutMs();
```
markelliot: can we use a Duration here instead?

uschi2000: Duration has a pretty funky serialization format; are you proposing we ask users to specify timeouts as something like "PT1M" (one minute), or that they specify milliseconds and the java class exposes Durations?

markelliot: I guess I was thinking of the Dropwizard Duration format, which might be useful to extract to a more general library for things like this, which allows encoding things like "100 milliseconds" and "100 ms" as well as "10 seconds" or "2d" (for 2 days).

uschi2000: I see. Yeah, that makes sense. ISO durations are too cryptic for most
people.

robert3005:
Just pushed version with Duration changes and tests. Thanks for the suggestion @markelliot @uschi2000 .

service-config/build.gradle
```diff
@@ -4,6 +4,7 @@ apply from: "$rootDir/gradle/publish.gradle"
 dependencies {
     compile project(':ssl-config')
     compile libs.tokens
+    compile libs.dropwizard.utils
```
uschi2000, line 7: what are the (transitive) dependencies of dropwizard-utils?

uschi2000:
looks OK modulo the transitive dependencies brought in by dropwizard. are we bringing in anything new with this change?

ash211:
Piggybacking a little on this PR -- the JDBC spec requires that the timeout on the underlying DB connection be configurable.  For that scenario should we be trying to change the timeout on an already created and active service config, or should we be instantiating a new service config with the modified timeout every time the timeout changes?

uschi2000:
Service configuration objects are technically immutable, so you'd have to
make a new one anyway. More importantly, you'd have to create new feign
clients given a new configuration (whether mutated are copied&changed).

robert3005:
@uschi2000 it's adding findbugs annotations and joda-time. Anything we don't want to include?

```
compile - Compile classpath for source set 'main'.
+--- project :ssl-config
|    +--- com.google.guava:guava:18.0
|    \--- com.fasterxml.jackson.core:jackson-databind:2.6.1
|         +--- com.fasterxml.jackson.core:jackson-annotations:2.6.0
|         \--- com.fasterxml.jackson.core:jackson-core:2.6.1
+--- com.palantir.tokens:auth-tokens:0.2.3
|    +--- com.fasterxml.jackson.core:jackson-databind:2.4.4 -> 2.6.1 (*)
|    +--- com.google.guava:guava:16.0.1 -> 18.0
|    +--- org.slf4j:jcl-over-slf4j:1.7.12
|    |    \--- org.slf4j:slf4j-api:1.7.12
|    \--- org.slf4j:slf4j-api:1.7.12
\--- io.dropwizard:dropwizard-util:0.8.2
     +--- com.fasterxml.jackson.core:jackson-annotations:2.5.0 -> 2.6.0
     +--- com.google.guava:guava:18.0
     +--- com.google.code.findbugs:jsr305:3.0.0
     \--- joda-time:joda-time:2.7
```

uschi2000:
ok, that should be fine

PR #109, opened 2016-04-26
Simplify builder definition

nmiyake:


pnepywoda:
+1

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
```diff
@@ -44,22 +44,10 @@
...
-        Builder apiToken(BearerToken apiToken);
-
-        Builder security(SslConfiguration security);
-
-        Builder uris(Iterable<String> uris);
-
-        Builder from(ServiceConfiguration otherConfig);
-
-        ServiceConfiguration build();
+    public static final class Builder extends ImmutableServiceConfiguration.Builder {
```
uschi2000, line 51: need the same change for ServiceDiscoveryConfiguration, right?

uschi2000:
actually, #109 already has has this change, let's go with that.

service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
qinfchen, line 51: Sorry to keep bringing back the thread. 

This patten does not hide the impl details that
- `Builder.build` method will return ImmutableServiceConfiguration typed configure
- `apiToken` method will return ImmutableServiceConfiguration.Builder typed builder
- and also for the rest of the setters

uschi2000, line 51: sure. but does that matter?

pnepywoda, line 51: We've been using this pattern on Foundry for a while and it hasn't mattered to us or our users so far...

nmiyake, line 51: Only case where I think it matters is compile errors it can produce -- see https://github.com/palantir/http-remoting/pull/111/files#diff-bd6f45eabbff99720347cb52c6b81eb1R39.

I don't have a strong feeling either way -- I think it is ideal for consumers to not have to worry about how they use the API (which would argue for the "hide the implementation" approach), but at the same time am also not a fan of all the boilerplate code required for the interface definitions (whole point of Immutables is to try to move away from boilerplate).

qinfchen, line 51: I don't think either solution is good enough :(. 
- the current one exposes both the abstract class and the generated class, and we still have some boilerplate for builders
- the other hides the implementations, but more boilerplate for the builder. 

There seems to be a better way (from Javier, Mark, and Punya) that, instead of using the abstract class and hides the generated class, hide the abstract class and use the generated class.
- even less boilerplate code and more functionality from the generated class
- it is cleaner in that only generated class is used
- always `final` vs exposing abstract class where there is a risk of users extending it. 

```
@Value.Immutable
@Value.Style(
    get = {"is*", "get*"}, // Detect 'get' and 'is' prefixes in accessor methods
    init = "set*", // Builder initialization methods will have 'set' prefix
    typeAbstract = {"Abstract*"}, // 'Abstract' prefix will be detected and trimmed
    typeImmutable = "*", // No prefix or suffix for generated immutable type
    builder = "new", // construct builder using 'new' instead of factory method
    build = "create", // rename 'build' method on builder to 'create'
    visibility = ImplementationVisibility.PUBLIC, // Generated class will be always public
    defaults = @Value.Immutable(copy = false)) // Disable copy methods by default
abstract class AbstractItem {
  abstract String getName();
  abstract Set<String> getTags();
  abstract Optional<String> getDescription();
}

// Use generated value object
Item namelessItem = Item.builder()
    .setName("Nameless")
    .addTags("important", "relevant")
    .setDescription("Description provided")
    .build();

Item namedValue = namelessItem.withName("Named");
```

pnepywoda, line 51: This means a lot of the generated class methods will be accessible/viewable
to the calling user (like fromAttributes, withName, withTags) and will NOT
include convenience .of() methods that we've specifically written in many
places in foundry

qinfchen, line 51: `of` methods should be available as long as you annotate the variables with `@Value.Parameter` in the abstract class, no? We are experimenting with the new pattern atm. Given the benefits above, I am fine with more generated methods be accessible to the calling user as long as they provide some values to the users.

nmiyake, line 51: Is there an email list or something like that where we can discuss more in depth? Although this method of using the generated class is interesting, philosophically it doesn't seem correct -- the codebase won't compile at all without the generated classes, and it breaks the idea that the code that we write specifies the API, and the generator classes just fill them out. Any generated method that happens to be public will "leak" into the API, and if `Immutables` later changes and either generates more public methods or changes the public methods that are generated in the dependent class, that will introduce external API breaks without any author action. I think the readability will suffer as well -- this isn't the standard way that frameworks like `Immutables` or `AutoValue` are used.

uschi2000, line 51: @chrisalice thoughts?

alicederyn, line 51: - I dislike referring directly to generated types because you can no longer find them in github, etc. Generated methods are fine as one once you've found the type, you can see the code generation annotations and look up the docs.
- Exposing final types in your API makes downstream tests fragile, as Mockito can no longer create mocks. [Smart nulls](http://docs.mockito.googlecode.com/hg/1.9.5/org/mockito/Answers.html#RETURNS_SMART_NULLS) are super helpful in tracking down test failures, and being able to pass trivial mocks through wiring code means simpler, clearer tests.

FWIW, you can keep your static `.of` methods, as static methods are inherited in Java, though IDEs aren't always cooperative.

qinfchen, line 51: @nmiyake we are still experimenting with the new pattern, will send out an email to discuss this more in depth. 
@chrisalice all immutable classes are value classes, so builders can be used instead of mocking them in tests which isn't a huge compromise i think :)

alicederyn, line 51: @quinfchen I'm guessing you didn't read about smart nulls.

PR #111, opened 2016-04-28
Immutables builder example

nmiyake:
NOSUBMIT: test build only


pnepywoda:
should we keep this PR open or close it?

nmiyake:
We can close it -- it was for demonstration only.

PR #112, opened 2016-04-28
add `isServiceEnabled` to ServiceDiscoveryConfiguration

### LINKED ISSUE #110: add `isServiceEnabled` for service discovery
can we have `isServiceEnabled` = there's config for the service and uris is not empty?

sixinli:
fixes #110

### DIFF
--- a/service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
+++ b/service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
@@ -136,6 +136,20 @@ public abstract class ServiceDiscoveryConfiguration {
         return getServiceConfig(serviceName).readTimeout();
     }
 
+    /**
+     * Checks if a service is enabled.
+     * <p>
+     * A service is enabled iff the configuration for it exists and its list of URIs is non-empty.
+     *
+     * @param serviceName the name of the service
+     * @return whether or not the service is enabled
+     */
+    public final boolean isServiceEnabled(String serviceName) {
+        ServiceConfiguration serviceConfig = getServices().get(serviceName);
+
+        return serviceConfig != null && !serviceConfig.uris().isEmpty();
+    }
+
     private ServiceConfiguration getServiceConfig(String serviceName) {
         return Preconditions.checkNotNull(getServices().get(serviceName),
                 "Unable to find the configuration for " + serviceName + ".");
--- a/service-config/src/test/java/com/palantir/config/service/ServiceDiscoveryConfigurationTests.java
+++ b/service-config/src/test/java/com/palantir/config/service/ServiceDiscoveryConfigurationTests.java
@@ -99,6 +99,17 @@ public final class ServiceDiscoveryConfigurationTests {
     }
 
     @Test
+    public void testIsServiceEnabled() throws IOException {
+        URL resource = Resources.getResource("configs/discovery-config-with-empty-uri.yml");
+        ServiceDiscoveryConfiguration discoveryConfig =
+                mapper.readValue(resource.openStream(), ServiceDiscoveryConfiguration.class);
+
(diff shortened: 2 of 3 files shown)


service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -136,6 +136,12 @@
         return getServiceConfig(serviceName).readTimeout();
     }
 
+    public final boolean isServiceEnabled(String serviceName) {
```
uschi2000, line 147: how about: `hasService()`?

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -136,6 +136,12 @@
...
+        ServiceConfiguration serviceConfig = getServices().get(serviceName);
+
+        return serviceConfig != null && !serviceConfig.uris().isEmpty();
```
uschi2000, line 150: Would it make sense to force uris() to be non-empty in the constructor instead?

sixinli, line 150: The main reason I want to add this function was because the none-empty check. We need to support services with empty list of Uris - this happens when service discovery tries to discover the service and finds nothing. So technically those services aren't enabled but will have an entry in the map. 

We could also find a way to filter services - ignore and remove those from the services list somehow (probably from the service discovery constructor). This seems a bit easier to understand but the code might be messier.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
sixinli, line 147: See the other reply :)

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
uschi2000, line 150: understood.

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
uschi2000, line 147: Deal. Please leave javadoc explaining what "enabled means": `/** Returns true iff the configuration for the given service name exists and its list of URIs is non-empty. */`

uschi2000, line 147: Deal. Please leave javadoc explaining what "enabled means": `/** Returns true iff the configuration for the given service name exists and its list of URIs is non-empty. */`

uschi2000, line 147: Deal. Please leave javadoc explaining what "enabled means": `/** Returns true iff the configuration for the given service name exists and its list of URIs is non-empty. */`

sixinli:
updated docs, also going to rebase branch now

sixinli:
friendly ping :D

service-config/src/main/java/com/palantir/config/service/ServiceDiscoveryConfiguration.java
```diff
@@ -136,6 +136,20 @@
...
 
+    /**
+     * Checks if a service is enabled.
+     * <p>
+     * A service is enabled iff the configuration for it exists and its list of URIs is non-empty.
+     *
+     * @param serviceName the name of the service
+     * @return whether or not the service is enabled
+     */
+    public final boolean isServiceEnabled(String serviceName) {
```
markelliot, line 147: should this and other attributes here be marked as `@Value.Derived`? http://immutables.github.io/immutable.html#derived-attributes

PR #115, opened 2016-05-02
Add call tracing

markelliot:
All remoting calls will now emit standard Zipkin-style tracing headers for traceId and spanId.

Additionally, applications including the tracing library can start and stop traces
inside the application using `Traces.deriveTrace(String)` and a completion call
to emit a span `Traces.complete()`. (Traces must happen within a single
thread, work dispatched to executors may not track correctly given current
implementation details.)

Servers can interpret, load, and will automatically continue passing
the same trace identifier by installing the `TraceInheritingFilter`
Jersey resource. Calls that did not include a trace will emit an identified trace
as a result of installing the filter.


tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * A container object used to hold information about call tracing.
+ */
+@JsonDeserialize(as = ImmutableTraceState.class)
+@JsonSerialize(as = ImmutableTraceState.class)
+@Value.Immutable
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+public abstract class TraceState {
+
+    private static final Random RANDOM = new Random();
```
schlosna: May want to consider [ThreadLocalRandom](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html) as:

> When applicable, use of ThreadLocalRandom rather than shared Random objects in concurrent programs will typically encounter much less overhead and contention.

http-clients/src/main/java/com/palantir/remoting/http/TraceRequestInterceptor.java
```diff
@@ -0,0 +1,34 @@
...
+import feign.RequestInterceptor;
+import feign.RequestTemplate;
+
+public final class TraceRequestInterceptor implements RequestInterceptor {
+
+    @Override
+    public void apply(RequestTemplate template) {
+        TraceState traceState = Traces.getTrace();
+        if (traceState != null) {
+            template.header(Traces.TRACE_HEADER, traceState.getTraceId());
```
schlosna: Do we want a span ID as well or add on later?

tracing/src/main/java/com/palantir/tracing/Traces.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.tracing;
+
+import com.google.common.base.Strings;
+
+public final class Traces {
+
+    public static final String TRACE_HEADER = "trace-id";
```
schlosna: I think we want something like `X-B3-TraceId` which is used by both [zipkin](https://twitter.github.io/zipkin/Instrumenting.html) and [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java#L52)

See also:
- http://opentracing.io/spec/
- https://spring.io/blog/2016/02/15/distributed-tracing-with-spring-cloud-sleuth-and-spring-cloud-zipkin
- http://cloud.spring.io/spring-cloud-sleuth/spring-cloud-sleuth.html#_feign

tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -0,0 +1,73 @@
...
+    @Value.Default
+    public String getTraceId() {
+        return randomId();
+    }
+
+    public static final Builder builder() {
+        return new Builder();
+    }
+
+    private static String randomId() {
```
splittingfield: What would you opinion be as to introducing a snowflake ID as opposed to random?  https://github.com/rholder/fauxflake

splittingfield: In particular, https://github.com/rholder/fauxflake/blob/master/fauxflake-core/src/main/java/com/github/rholder/fauxflake/provider/boundary/FlakeEncodingProvider.java

tracing/src/main/java/com/palantir/tracing/Traces.java
markelliot: Ah, ok. Happy to switch headers. It looks like those take 64bit identifiers instead of arbitrary strings, which makes one of the things here less ideal -- we wanted the ability to name a trace explicitly (rather than use a generated identifier), so perhaps we should pass that separately?

pnepywoda:
do we still need this now that https://github.com/palantir/http-remoting/pull/74 is merged?

markelliot:
Yes, this provides a per-invocation tracking system which will supply the same id to code that wants it traversing possibly many system boundaries. (See Zipkin, for instance, as a viewer of this kind of data.)

pnepywoda:
Ah I see. Ok cool. Though are there existing libraries that do this for us? Took a quick look around and would something like https://github.com/openzipkin/brave work for us? (maybe it's too much overhead and that's why we're doing it ourselves here?)

markelliot:
Latest changeset moves to Zipkin/OpenTracing semantics, things left to do here:
- generate 64bit IDs instead of UUIDs
- figure out how to complete spans on Feign/Retrofit returns (which may or may not matter since we now automatically complete spans when the server responds)
- figure out where to send completion events

http-clients/src/main/java/feign/TraceResponseDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+    public TraceResponseDecoder(Decoder delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Object decode(Response response, Type type) throws IOException, DecodeException, FeignException {
+        String traceId = safeGetOnlyElement(response.headers().get(Traces.Headers.TRACE_ID), null);
+        String spanId = safeGetOnlyElement(response.headers().get(Traces.Headers.SPAN_ID), null);
+        Optional<TraceState> trace = Traces.getTrace();
+        if (traceId != null && spanId != null && trace.isPresent()) {
```
markelliot, line 41: we could consider always popping the top of the trace here, in theory every single request should be matched by a decode call.

tracing/src/main/java/com/palantir/tracing/Traces.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static final Set<Subscriber> SUBSCRIBERS = Sets.newConcurrentHashSet();
+
+    public static Optional<TraceState> getTrace() {
+        Deque<TraceState> stack = STATE.get();
+        return (!stack.isEmpty()) ? Optional.of(stack.peek()) : Optional.<TraceState>absent();
+    }
+
+    public static void setTrace(TraceState state) {
+        STATE.get().clear();
```
schlosna: We may want to `STATE.remove()` rather than clear if these deques are going to grow significantly beyond their default size, to allow shrinking.

schlosna:
:+1:

markelliot:
Dave, meant to ask, do you know if the IDs need to be 64bit or if that’s just common practice? If it’s common practice, how do others typically get them to be unique enough?

splittingfield:
Every implementation I have seen uses a variation of twitters snowflake algorithm.

marc

On May 18, 2016, at 11:20 AM, Mark Elliot <notifications@github.com<mailto:notifications@github.com>> wrote:

Dave, meant to ask, do you know if the IDs need to be 64bit or if that’s just common practice? If it’s common practice, how do others typically get them to be unique enough?

markelliot:
Interesting, it looks like that's a central service rather than an algorithm. With the fauxflake thing you linked I'm not sure what to seed as the service identifier.

splittingfield:
It can be anything really or a config value specified at startup time.

marc

On May 18, 2016, at 12:20 PM, Mark Elliot <notifications@github.com<mailto:notifications@github.com>> wrote:

Interesting, it looks like that's a central service rather than an algorithm. With the fauxflake thing you linked I'm not sure what to seed as the service identifier.

markelliot:
That's not super satisfying, I'll look at what we can use.

splittingfield:
I like time stamp: MAC address: service specific id : counter (from

marc

On May 18, 2016, at 12:45 PM, Mark Elliot <notifications@github.com<mailto:notifications@github.com>> wrote:

That's not super satisfying, I'll look at what we can use.

splittingfield:
Maybe some variation of service name+version:MAC address: time stamp?

Assuming that MAC address is accessible and that we don't collide multiple copies services on the same host and we can trust outer ntp servers.

marc

On May 18, 2016, at 12:45 PM, Mark Elliot <notifications@github.com<mailto:notifications@github.com>> wrote:

That's not super satisfying, I'll look at what we can use.

markelliot:
I think the challenge is turning that into a 64bit number?

The sad thing is the form you're describing what UUIDs are supposed to be anyway.

splittingfield:
Yeah I know, but type 1 uuids are so confusing!

marc

On May 18, 2016, at 1:44 PM, Mark Elliot <notifications@github.com<mailto:notifications@github.com>> wrote:

I think the challenge is turning that into a 64bit number?

The sad thing is the form you're describing what UUIDs are supposed to be anyway.

tracing/src/main/java/com/palantir/tracing/Span.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public abstract Optional<String> getParentSpanId();
+
+    public abstract String getSpanId();
+
+    public abstract String getOperation();
+
+    public abstract long getStartTimeMs();
+
+    public abstract long getDurationNs();
```
uschi2000, line 43: does ns resolution make sense if the start time is in ms?

tracing/src/main/java/com/palantir/tracing/Span.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
uschi2000, line 2: inconsistent header

tracing/src/main/java/com/palantir/tracing/Span.java
markelliot, line 43: Accurate timing on the JVM needs to use the nanosecond counter on the CPU, but wall clock time is only really available in milliseconds. I'm happy to convert the nanoseconds to milliseconds but I figured I'd save the math operation and leave that to a subscriber. (Don't have a strong opinion, just explaining rationale.)

uschi2000, line 43: ok, that sounds fine

tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -0,0 +1,109 @@
...
+                .startTimeMs(System.currentTimeMillis())
+                .startClockNs(System.nanoTime());
+    }
+
+    public static String randomId() {
+        // non-secure random generated UUID because speed is important here and security is not
+        byte[] randomBytes = new byte[16];
+        ThreadLocalRandom.current().nextBytes(randomBytes);
+
+        randomBytes[6]  &= 0x0f;  /* clear version        */
```
uschi2000: I'm curious, why don't you just set it to 0x40?

markelliot: Er yeah, this is just the UUID#randomUUID code with a faster source of random. I need to attribute that somehow, or we need to pick someone else's Id library.

tracing/src/main/java/com/palantir/tracing/Traces.java
```diff
@@ -0,0 +1,116 @@
...
+        protected Deque<TraceState> initialValue() {
+            return new ArrayDeque<>();
+        }
+    };
+
+    private static final Set<Subscriber> SUBSCRIBERS = Sets.newConcurrentHashSet();
+
+    public static Optional<TraceState> getTrace() {
+        Deque<TraceState> stack = STATE.get();
+        return (!stack.isEmpty()) ? Optional.of(stack.peek()) : Optional.<TraceState>absent();
```
uschi2000: avoid negation for readability?

tracing/src/main/java/com/palantir/tracing/Traces.java
```diff
@@ -0,0 +1,116 @@
...
+    public static void setTrace(TraceState state) {
+        STATE.remove();
+        STATE.get().push(state);
+    }
+
+    /**
+     * Derives a new call trace from the currently known call trace labeled with the
+     * provided operation.
+     */
+    public static TraceState deriveTrace(String operation) {
```
uschi2000: the name should convey that this changes the internal state. addToTrace? addOperation?

tracing/src/main/java/com/palantir/tracing/TraceState.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Returns a description of the operation for this event.
+     */
+    public abstract String getOperation();
+
+    /**
+     * Returns the start time in milliseconds of the span represented by this state.
+     * <p>
+     * Users of this class should not set this value manually in the builder, it is configured
```
uschi2000, line 42: that's a little awkward since anyone can do `new Builder()`. but I also don't know a better solution.

tracing/src/main/java/com/palantir/tracing/Span.java
```diff
@@ -0,0 +1,39 @@
...
+import org.immutables.value.Value;
+
+/**
+ * A value class representing a completed Span.
+ */
+@JsonDeserialize(as = ImmutableSpan.class)
+@JsonSerialize(as = ImmutableSpan.class)
+@Value.Immutable
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+public abstract class Span {
```
uschi2000, line 31: I'm a little confused: why do we need both TraceState and Span if they contain mostly the same things?

http-clients/src/main/java/feign/TraceResponseDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+            if (trace.get().getTraceId().equals(traceId)
+                    && trace.get().getSpanId().equals(spanId)) {
+                // this trace is for the traceId and spanId on top of the tracing stack, complete it
+                Traces.complete();
+            }
+        }
+        return delegate.decode(response, type);
+    }
+
+    private static <T> T safeGetOnlyElement(Iterable<T> iterable, T defaultValue) {
```
uschi2000, line 53: add some @Nullables

tracing/src/main/java/com/palantir/tracing/Span.java
uschi2000, line 31: ah, it's for the duration. nevermind.

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/RetrofitClientFactory.java
```diff
@@ -32,6 +33,8 @@
  */
 public final class RetrofitClientFactory {
 
+    private static final Interceptor TRACE_INTERCEPTOR = new TraceInterceptor();
```
uschi2000: enum singleton?

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/TraceInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+                .header(Traces.Headers.SPAN_ID, callState.getSpanId());
+        if (callState.getParentSpanId().isPresent()) {
+            instrumentedRequest.header(Traces.Headers.PARENT_SPAN_ID, callState.getParentSpanId().get());
+        }
+
+        Response response;
+        try {
+            response = chain.proceed(instrumentedRequest.build());
+        } finally {
+            // complete response
```
uschi2000: redundant comment

tracing/src/main/java/com/palantir/tracing/Traces.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public interface Headers {
+        String TRACE_ID = "X-B3-TraceId";
+        String PARENT_SPAN_ID = "X-B3-ParentSpanId";
+        String SPAN_ID = "X-B3-SpanId";
+        String SPAN_NAME = "X-Span-Name";
+    }
+
+    // stack of trace states
+    private static final ThreadLocal<Deque<TraceState>> STATE = new ThreadLocal<Deque<TraceState>>() {
```
uschi2000, line 35: trying to understand the flow here: does the queue ever contain more than two elements?

markelliot, line 35: Yeah, you can add spans within an application using the static derive/complete stuff, which would get you more fine-grained detail of what's happening if you emit stuff to, say, Zipkin.

tracing/src/main/java/com/palantir/tracing/Traces.java
markelliot: How about "startSpan()" and I'll change complete to "completeSpan()"?

tracing/src/main/java/com/palantir/tracing/Traces.java
markelliot: done

tracing/src/main/java/com/palantir/tracing/Traces.java
markelliot: done

tracing/src/main/java/com/palantir/tracing/TraceState.java
markelliot, line 42: going to fix this using a different Immutables pattern.

http-clients/src/main/java/feign/TraceResponseDecoder.java
markelliot, line 53: what lib do those come from, also not totally clear on how to label this

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/TraceInterceptor.java
markelliot: done

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/RetrofitClientFactory.java
markelliot: done

tracing/src/main/java/com/palantir/tracing/TraceState.java
markelliot: switching to use just a single random long -- Zipkin wants a 64bit id, and this is what Brave does anyway: 
https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ClientTracer.java#L131

tracing/src/main/java/com/palantir/tracing/Span.java
markelliot, line 2: would be really nice if something enforced this...but fixing all of them

uschi2000, line 2: have been tracking that for a while, but it's not done:
https://github.com/palantir/gradle-baseline/issues/59

markelliot:
@uschi2000 ready for another look, LMK if you have additional feedbakc

tracing/src/main/java/com/palantir/tracing/Traces.java
uschi2000, line 35: but without additional calls, i.e., just with DW/Feign as per this PR, it would only ever be at most two elements?

markelliot, line 35: Without additional tracing calls should be at most 2 elements, yes.

tracing/src/main/java/com/palantir/tracing/Traces.java
uschi2000: cool

tracing/src/main/java/com/palantir/tracing/Traces.java
uschi2000, line 35: ok, cool

jkozlowski:
For the tracing, how about using https://github.com/openzipkin/brave? I am about to prototype this for Phoenix.

markelliot:
Could you point me to some details on how to make a span internal to a server with Brave? Poking around the readmes and code this seems entirely non-obvious to me, and is an important part of the tracing framework. 

Separately, as long as the HTTP parts of the implementation match the Zipkin APIs, it shouldn't matter how the internals work.

uschi2000:
Like this?
https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java

markelliot:
Maybe I'm misreading the code, but that doesn't seem to allow nesting of recorded local spans?

uschi2000:
Haven't dug in enough detail to answer this either, sorry.

uschi2000:
Generally, the Brave code-base looks very solid (and has no external
dependencies), so I think we should consider it, even if it's currently
lacking that feature. (We could always push for adding it there.)

markelliot:
It looks like it'd be a pretty substantive change to how it stores state (been poking around code for quite a bit now) to make it work, and would be enough of an API break that it'd likely mean a major version bump for it. 

We've found a lot of value from other internal initiatives having this particular feature, though maybe it's less interesting when services are granular enough. 

At any rate, I can make two proposals:
1. You take a stab at an alternative PR that adds Brave everywhere necessary (I'm not sure I have time at this point to retry this with another library)
2. We take this PR and advise against anyone invoking "Traces" directly for internal instrumentation so that we can re-evaluate later

Upside for option #1 is immediate integration with Zipkin. I think higher priority for us than that, though, is a tracing-aware dropwizard logging addition, though, so we can get trace and span ids into every log statement we emit, including in request logs.

jkozlowski:
I am going to be digging into this a bit more over the course of next week, but my understanding was that the definitely had some solution for inter-process tracing (ThreadLocal based) and they had a solution for crossing threadpool boundaries.

jkozlowski:
I am definitely going to be prototyping this for Phoenix with Brave, simply because that is the only thing I can use right now (we have upgrade to http-remoting in the works). I want to see Zipkin style traces coming out of Phoenix: adding trace aware log messages is useful, but right now Phoenix doesn't actually do much that much logging, so it's lower priority.

markelliot:
Crossing thread pool boundaries is simple enough and would be a small addition here, the challenge is maintaining an internal stack of spans so that you can instrument internally in a detailed way. Again, not that familiar with Brave, so definitely open to the possibility I'm missing something in its capability or implementation.

uschi2000:
Should we set a deadline for the Brave experimentation and punt on this PR
until then? Would suggest end of next week.

jkozlowski:
A bit more digging around: I hooked brave into a servlet filter and then invoked LocalTracer twice:

```
    public PhoenixVersion getVersion() {
        brave.localTracer().startNewSpan("this-is-a-test", "text");
        try {
            Thread.sleep(1000);
            secondLevel();
        } catch (InterruptedException e) {
            throw Throwables.propagate(e);
        } finally {
            brave.localTracer().finishSpan();
        }
    }

    private void secondLevel() {
        brave.localTracer().startNewSpan("this-is-a-test1", "text1");
        try {
            Thread.sleep(2000);
        } catch (InterruptedException e) {
            throw Throwables.propagate(e);
        } finally {
            brave.localTracer().finishSpan();
        }
    }
```

Gave me 2 traces:

```
{"traceId":"0000000000000001","name":"text1","id":"7672bbb60ba5b369","parentId":"0000000000000002","timestamp":1464889183460000,"duration":2004831,"annotations":[],"binaryAnnotations":[{"key":"lc","value":"this-is-a-test1","endpoint":{"serviceName":"braveservletinterceptorintegration","ipv4":"10.203.68.159"}}]}
```

```
{"traceId":"0000000000000001","name":"get","id":"0000000000000002","timestamp":1464889182435000,"duration":3150000,"annotations":[{"endpoint":{"serviceName":"braveservletinterceptorintegration","ipv4":"10.203.68.159"},"timestamp":1464889182435000,"value":"sr"},{"endpoint":{"serviceName":"braveservletinterceptorintegration","ipv4":"10.203.68.159"},"timestamp":1464889185585000,"value":"ss"}],"binaryAnnotations":[{"key":"http.status_code","value":"200","endpoint":{"serviceName":"braveservletinterceptorintegration","ipv4":"10.203.68.159"}},{"key":"http.url","value":"/phoenix/api/v0/version","endpoint":{"serviceName":"braveservletinterceptorintegration","ipv4":"10.203.68.159"}}]}
```

So the traces got linked correctly, but it looks to me like it lost the intermediate _this-is-a-test_ and only recorded _this-is-a-test1_ - so no stacking as per what @markelliot is saying. I'll continue poking around tomorrow, maybe I am using it wrong.

jkozlowski:
Looking at it a bit more, all the ServerTracer, ClientTracer and LocalTracer operate on ServerClientAndLocalSpanState which is an interface and can be replaced. Not sure how many things it would break if we replaced it's ThreadLocalServerClientAndLocalSpanState with what you had implemented here. Not sure how valuable it is to use this library if we need to replace this part of it and how constraining it is.

jkozlowski:
Ok I think I have a better understanding of the library:

The core is ServerClientAndLocalSpanState: this is basically the per-thread storage for the current server, local and client spans: it only stores a single span per each category.

Then you have ServerTracer -> LocalTracer -> ClientTracer. This means that whenever LocalTracer wants to start a new span, it will lookup the current server trace from ServerClientAndLocalSpanState and use that as it's parent and set the new span into ServerClientAndLocalSpanState. Similarly, for ClientTracer: it will use the current local trace as it's parent. 

Therefore, if you want to have nesting in local tracing, you just need to have another layer that restores the current local trace after a subtrace returns: the ServerClientAndLocalSpanState has basically methods that set the spans for server, local and client.

I am going to go with Brave for my little experiment (in fact I have almost finished integrating it), since adding this small layer for nested local tracing (which I will not want just yet) should be fairly trivial and I gain a lot of the out of the box integrations that Brave comes with.

Now sure how we'd like to proceed? I suggest:
- Amend the PR to use Brave and hook everything up with a caveat that LocalTracer should not be used for multiple layers of tracing.
- Contribute back the filters you created and integrations for retrofit etc. and then pull in the newest version.
- Once we need the local tracing amended, implement it here first and test drive it on a few projects and see how they feel about contributing this back.

jkozlowski:
Also how does this all compare to Chronicle and what should we use?

markelliot:
Chronicle and the implementation here are nearly identical (though that’s not so surprising since I’m the original author of Chronicle). We would’ve used it except its dependency tree makes it hard to open source – I looked at the impl a bit as I put together this setup.

I think internal spans have been extremely useful in the past, and would highly value continuing to have them.

Replacing the guts of Brave seems pretty daunting – the setup here would give us an API-compatible system though none of the Zipkin connectivity, so the trade is likely a hard one.

jkozlowski:
Looks like there is interest from brave to support this: https://github.com/openzipkin/brave/issues/166

PR #117, opened 2016-05-03
Fix immutable style warning

qinfchen:
Inline style causes the following warning

```
warning: unknown enum constant ImplementationVisibility.PACKAGE
  reason: class file for org.immutables.value.Value$Style$ImplementationVisibility not found
```

See this pr as an example, https://github.com/qinfchen/dropwizard-example-app/pull/3

I created an issue for Immutables, https://github.com/immutables/immutables/issues/291. Here is the summary of the issue:

> I've investigated that in a detail. It's a native javac warning issued when -Xlint is enabled. There is no category for it so it cannot be disabled, unlike other categorized warnings such as "unchecked" etc. Once you're linting, it's always enabled. Of course it would be no issues if we use source only annotation retention. But we use class-file retention for the annotations so incremental compilation and advanced discovery works during annotation processing.

@uschi2000 @nmiyake it seems that ssl-config has the same issue. Is there a plan to create -commons project to include common classes? so we can move the custom style to there and share it across other projects.

### DIFF
--- a/service-config/src/main/java/com/palantir/config/service/PackageVisibilityImmutableStyle.java
+++ b/service-config/src/main/java/com/palantir/config/service/PackageVisibilityImmutableStyle.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package com.palantir.config.service;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.immutables.value.Value.Style;
+
+/**
+ * Sets the style to be used by the Immutables library.
+ */
+@Target({ElementType.PACKAGE, ElementType.TYPE})
+@Style(
+    visibility = Style.ImplementationVisibility.PACKAGE)
+@interface PackageVisibilityImmutableStyle {}
--- a/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
+++ b/service-config/src/main/java/com/palantir/config/service/ServiceConfiguration.java
@@ -23,11 +23,10 @@ import com.palantir.tokens.auth.BearerToken;
 import io.dropwizard.util.Duration;
 import java.util.List;
 import org.immutables.value.Value.Immutable;
-import org.immutables.value.Value.Style;
 
 @Immutable
 @JsonDeserialize(as = ImmutableServiceConfiguration.class)
-@Style(visibility = Style.ImplementationVisibility.PACKAGE)
+@PackageVisibilityImmutableStyle
 public abstract class ServiceConfiguration {
 
     /**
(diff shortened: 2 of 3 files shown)


nmiyake:
So to make this concrete, the suggestion is to create a `commons` or `base` project, define the Immutable style annotation there, and apply it to all Immutables in the project? I'd be fine with that -- seems like a bit of overhead, but don't really see a major downside, and would allow the project to ensure that the Immutable styles/practices are consistent across the projects.

service-config/src/main/java/com/palantir/config/service/CustomImmutableStyle.java
```diff
@@ -0,0 +1,20 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.immutables.value.Value.Style;
+
+/**
+ * Custom style for immutable classes.
+ */
+@Target({ElementType.PACKAGE, ElementType.TYPE})
```
nmiyake: Are the `@Target` and `@Retention` annotations needed?

service-config/src/main/java/com/palantir/config/service/CustomImmutableStyle.java
```diff
@@ -0,0 +1,20 @@
...
+@Retention(RetentionPolicy.CLASS)
+@Style(
+    visibility = Style.ImplementationVisibility.PACKAGE)
+@interface CustomImmutableStyle {}
```
nmiyake: Name is pretty generic -- would prefer it either communicate the functionality (`PackageAccessImmutable`, `PackageVisibilityImmutableStyle`, etc.) or project (`HttpRemotingImmutable`)

qinfchen:
We have some setup for tests that can be shared as well.

service-config/src/main/java/com/palantir/config/service/CustomImmutableStyle.java
qinfchen: +1 on HttpRemotingImmutableStyle if we can move it to `base` or `common`

nmiyake:
From what I recall, test code isn't visible to downstream projects, so only way to share test code is to have it as part of the source of an upstream project, which often isn't ideal.

service-config/src/main/java/com/palantir/config/service/CustomImmutableStyle.java
qinfchen: removed @Retention, but kept @Target since we only want the style to be applied at the package or the class/interface level

qinfchen:
I see. Was thinking about creating some test utils or rules to improve the tests

nmiyake:
@qinfchen @markelliot @uschi2000 

Pretty sure this issue can be closed. These warnings don't appear when using the `./gradlew` build tasks because the Palantir annotation-processor plugin properly adds the annotation classes to the class path used by javac (see https://github.com/palantir/gradle-processors/issues/19). Confirmed locally, where doing a clean build doesn't appear to show any of these warnings even without refactoring the annotation. Qinfeng, are you seeing different behavior? If not, then I don't think this is an issue and we can likely close the PR.

qinfchen:
cool, thanks for the update!

PR #118, opened 2016-05-10
handle retrofit errors similar to feign errors fixes #116

### LINKED ISSUE #116: Retrofit client factory should support propagating exceptions
Currently, any clients constructed through Retrofit will not include any information about the inner exception in the Java RetrofitException that comes back, despite the JSON representation having an extra field that specifically contains such information. This effectively means that certain inner exception information is swallowed.

For Feign clients, it appears we do extra work in FeignClientFactory to support this, likely through our custom SerializableErrorErrorDecoder.java. The request is to do something similar with our Retrofit clients, perhaps by building a Retrofit's custom ErrorHandler.

Examples of exceptions that get swallowed include any errors from Spark, such as exceptions relating to jagged rows in our TextDataFrameReader.

pnepywoda:
@uschi2000

### DIFF
--- a/retrofit-clients/src/main/java/com/palantir/remoting/retrofit/errors/RetrofitSerializableErrorErrorHandler.java
+++ b/retrofit-clients/src/main/java/com/palantir/remoting/retrofit/errors/RetrofitSerializableErrorErrorHandler.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting.retrofit.errors;
+
+import com.google.common.collect.ImmutableSet;
+import com.palantir.remoting.http.errors.SerializableErrorToExceptionConverter;
+import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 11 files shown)


pnepywoda:
@eshlin

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/errors/RetrofitSerializableErrorErrorHandler.java
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        InputStream body;
+        try {
+            body = response.getBody().in();
+        } catch (IOException e) {
+            return new RuntimeException("Cannot get input stream from response: " + e.getMessage(), e);
+        }
+        return SerializableErrorErrorDecoder.INSTANCE.decode(contentTypes, response.getStatus(), response.getReason(),
```
uschi2000: move into try block?

uschi2000:
very nice. just to confirm: this doesn't break any compatibility, right?

eshlin:
awesome - looks great. thanks for implementing. +1

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/errors/RetrofitSerializableErrorErrorHandler.java
pnepywoda: well, the point of the error decoder was to never throw an exception but to always return one....do you still think we should wrap in try/catch though? (the try block above is actually specific to getting the input stream currently....you propose we widen the responsibility of the try/catch?)

pnepywoda:
ping

pnepywoda:
1. made util instead of enum
2. there's no backcompat break

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/errors/RetrofitSerializableErrorErrorHandler.java
uschi2000: ok, let's leave it as it is actually.

uschi2000:
build broken, otherwise +1

pnepywoda:
@uschi2000 build isn't broken? can you merge?

PR #119, opened 2016-05-18
standardize registering jersey components made in http-servers

pnepywoda:
for example, OptionalAsNoContentMessageBodyWriter wasn't being included

@uschi2000 
@leeavital for SA on compass - it should use this new endpoint. Mayank found an issue with compass returning Optional while testing https://github.palantir.build/elements/foundry/pull/928/commits/96bfb75b8d2bfcc7c39b440cd87b69d4c789884c : https://circle.palantir.build/gh/elements/foundry/4334


http-servers/src/main/java/com/palantir/remoting/http/server/JerseyComponents.java
```diff
@@ -0,0 +1,43 @@
...
+    private JerseyComponents() {}
+
+    /**
+     * Registers standard exception mappers and message body writers.
+     *
+     * @param includeStackTrace
+     *        include the stack trace in exceptions written to the BODY of the http response; turn off when clients are
+     *        untrusted and shouldn't know details about the server implementation
+     */
+    public static void registerStandardComponents(Environment env, boolean includeStackTrace) {
```
markelliot: Should we be passing in whatever `env.jersey()` is instead of the `Environment`? We don't seem to be using anything from the top level object, and the class is named specifically for JerseyComponents.

pnepywoda:
@markelliot re-pushed

http-servers/src/main/java/com/palantir/remoting/http/server/JerseyComponents.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Registers standard exception mappers and message body writers.
+     *
+     * @param includeStackTrace
+     *        include the stack trace in exceptions written to the BODY of the http response; turn off when clients are
+     *        untrusted and shouldn't know details about the server implementation
+     */
+    //DEV NOTE: Consumer instead of Jersey's ResourceConfig since jersey 1/2 are different
+    public static void registerStandardComponents(Consumer<Object> componentConsumer, boolean includeStackTrace) {
```
markelliot: should we force `includeStackTrace` to false at this point?

http-servers/src/main/java/com/palantir/remoting/http/server/JerseyComponents.java
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting.http.server;
+
+public final class JerseyComponents {
```
markelliot, line 21: should we call this `RemotingJerseyComponents` to distinguish?

http-servers/src/test/java/com/palantir/remoting/http/server/ExceptionMappersTest.java
```diff
@@ -16,11 +16,10 @@
 
 package com.palantir.remoting.http.server;
```
uschi2000, line 17: need to rename this file

http-servers/src/main/java/com/palantir/remoting/http/server/JerseyComponents.java
uschi2000, line 21: I wonder if we should pull all server setup routines into one call?
`RemotingServers.configure(env, ...);`

markelliot, line 21: Isn't that what this is?

uschi2000, line 21: there is also the brave business

markelliot, line 21: Ah yes, should add that here (this PR predates its addition)

pnepywoda, line 21: brave is for feign instead of jersey I thought? also, JerseyComponents doesn't distinguish from anything......what do you mean?

uschi2000, line 21: Look at DropwizardTracingFilters.

pnepywoda:
So, right now JerseyComponents isn't dropwizard-specific and can be used in places like spark-module-server. If we add DropwizardTracingFilters to it, then it becomes dropwizard-specific. Or are you recommending we have a higher-up type class which is DropwizardComponents that does BOTH registering exception stuff and tracing stuff?

pnepywoda:
Know what, I'll just do that. seems reasonable

uschi2000:
closing in lieu of https://github.com/palantir/http-remoting/pull/201

PR #120, opened 2016-05-18
Newer versions of Jackson no longer ship jdk7 module

Vidit-Gupta:


uschi2000:
I assume this refers to: "Starting with Jackson 2.7, this module will be DEPRECATED, as its handlers will be incorporated directly in core Databind."?

It's not really clear to me what the right way is to handle this. Do we need different http-remoting releases for different jackson versions? Should we shade jackson so we can always use the correct version?

markelliot:
I suspect we should shade the dependency. This is only for the client anyway, I think, and we could eliminate all the Jackson 2.4 mess as a result.

uschi2000:
yep, I'm gravitating towards that solution too

uschi2000:
let's take this as a stop-gap and then do shading as soon as someone has the time. I think I need this for gradle-compatibility anyway.

uschi2000:
two requests before we merge:
- add a TODO in your code, saying that this should go away once we do shading
- add a test showing that serialization of java7 types works. (not sure such a test exists at the moment)

Vidit-Gupta:
@uschi2000 Did both things you suggested, whenever you want to take another look. Thanks!

uschi2000:
Thanks!
FYI: https://github.com/palantir/http-remoting/releases/tag/0.10.1
(need to wait for publish build to run, about 5-6mins)

PR #123, opened 2016-05-26
Use default encoder when body type is text/plain

markelliot:


dotCipher:
👍

http-clients/src/main/java/feign/TextDelegateEncoder.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
```
ash211: 2016

PR #124, opened 2016-05-27
Refactor FeignClients

ellisjoe:
@markelliot wanted to do this before adding yet another Contract to every factory method

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -26,6 +26,8 @@ import feign.Request;
 import feign.Request.Options;
 import feign.TextDelegateDecoder;
 import feign.TextDelegateEncoder;
+import feign.codec.Decoder;
+import feign.codec.Encoder;
 import feign.jackson.JacksonDecoder;
 import feign.jackson.JacksonEncoder;
 import feign.jaxrs.JaxRsWithHeaderAndQueryMapContract;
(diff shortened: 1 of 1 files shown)


http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -132,14 +101,18 @@ public static FeignClientFactory withMapper(ObjectMapper mapper) {
      * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
      */
     public static FeignClientFactory withMapper(ObjectMapper mapper, Request.Options timeoutOptions) {
+        return withEncoderAndDecoder(new JacksonEncoder(mapper), new JacksonDecoder(mapper), timeoutOptions);
+    }
+
+    /**
+     * Provides a {@link FeignClientFactory} with the specified {@link Encoder} and {@link Decoder}.
+     */
+    public static FeignClientFactory withEncoderAndDecoder(Encoder encoder, Decoder decoder,
```
markelliot: should we make this private?

ellisjoe: Not sure I see the benefit of hiding this?

markelliot: Freedom to change/refactor it later, because FeignClients is really part of the public part of this library.

ellisjoe: fair. making it private

PR #125, opened 2016-05-27
Jellis/slash encoding

ellisjoe:
@markelliot one open question is if we should make encoding slashes an opt in feature rather than the default behavior. I haven't found a good reason why we shouldn't default to encoding besides Feign not encoding by default.

### DIFF
--- a/http-clients/src/test/java/com/palantir/remoting/http/SlashEncodingContractTest.java
+++ b/http-clients/src/test/java/com/palantir/remoting/http/SlashEncodingContractTest.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


http-clients/src/test/java/com/palantir/remoting/http/SlashEncodingContractTest.java
```diff
@@ -0,0 +1,64 @@
...
+        @GET
+        @Path("/foo/{foo}")
+        String encoded(@PathParam("foo") String foo, @QueryParam("bar") String bar);
+    }
+
+    @Test
+    public void testEncodePathParam() throws InterruptedException {
+        proxy.encoded("slash/path", "slash/query");
+        RecordedRequest request = server.takeRequest();
+        assertThat(request.getPath(), is("/foo/slash%2Fpath?bar=slash%2Fquery"));
```
markelliot: hm, we might want to maintain the behavior that queryparam slashes are decoded, while pathparam slashes are not

ellisjoe: thought the same thing except theres no easy way around this. Feign encodes and replaces query and path params then does the slash decoding to the entire resolved url: https://github.com/Netflix/feign/blob/master/core/src/main/java/feign/RequestTemplate.java#L216

My understanding is that Jersey shouldn't have a problem decoding query params though.

markelliot:
I think default encoding is fine, will mark this as a breaking change though

http-clients/src/test/java/com/palantir/remoting/http/SlashEncodingContractTest.java
markelliot: should we add the tests that show this behaves well with our jersey setup?

ellisjoe:
@uschi2000 for SA. We can make decoding opt in later if we find a reason.

markelliot:
lgtm, just fix the withEncoderAmdDecoder method to be private and we're good to go.

ellisjoe:
rebased

PR #126, opened 2016-06-01
Go errors

nmiyake:


nmiyake:
@markelliot @uschi2000 

This PR adds an ErrorDecoder that allows Go exceptions to be decoded as Java exceptions. This decoder supports native/standard Go errors, Go errors formatted using our https://github.com/palantir/stacktrace library and Go panics (although panics are not processed quite as nicely as the first two).

The first commit consists of the implementation. The second commit adds an `http-clients-integration` project that runs integration tests against a local server to verify the Go error processing (in CI, a Docker container that runs a Go web server that serves up errors is run).

Current implementation preserves back-compat, but deprecates `FeignSerializableErrorErrorDecoder` in favor of `ErrorDecoders.SERIALIZABLE_ERROR` (the `ErrorDecoders` provides implementations for different error decoders, including the Go ones). If we're okay with breaking back-compat, would be cleaner to remove the class entirely.

Let me know what you guys think.

uschi2000:
I will look tomorrow, thanks Nick!

markelliot:
I think we may want to consider not taking this, and moving to a stronger world of non-optional exception suppression. With distributed tracing and centralized logging I don't think we get much benefit from pushing more than the type of exception and a descriptive message, at this point....and for security reasons it may be better to default to suppression always.

nmiyake:
Is there an intermediate world where you will take this (or something similar) until we have a tracing solution more widely deployed and move towards that? Fine if not, but in that case I'll probably use this same approach client-side for now until our producer applications make the bigger change of pushing back tracing IDs instead of logs. Basically, right now we're sending back stack traces that are poorly formatted/unreadable. Correct long-term solution is don't send traces at all and use IDs to track back instead. I basically don't want to completely punt on the current pain of "traces that we do hand back suck to read" just because the long-term solution is to not send them (especially when I have the implementation mostly written already).

uschi2000:
Long-term, I could see this go in one of two directions:
1. we have a language-independent way of forwarding exceptions across system boundaries
2. we never forward exceptions across system boundaries and rely on call tracing only to re-assemble error states across systems

Since we need call tracing in non-error cases anyway (for instance for performance measuring, etc), option 2 seems more appealing.

Short-term, this PR is a sort-of half-baked version of 1.: it adds support for Go, but keeps a Java-centric serialization format. While I'm not a big fan of this approach long-term, I would still support it as a stop-gap until we have 2., as Nick suggests. It should be clear though, that this facility will eventually get removed.

nmiyake:
Agree with this characterization. Given that (2) is the likely long-term approach for remoting, I'm also fine with remoting itself not accepting this PR and just providing this ourselves on the client side for our Go stuff until remoting has full tracing support.

markelliot:
@nmiyake going to close this for now, please let us know if you'd like to revisit later/after current round of refactoring

PR #127, opened 2016-06-01
Use configuration-resolver plugin to resolve dependencies

nmiyake:


nmiyake:
@markelliot @uschi2000

PR #129, opened 2016-06-05
Update gradle-processors and refactor Javadoc

nmiyake:


nmiyake:
@uschi2000 @markelliot 

Version bump and minor Gradle refactor

PR #130, opened 2016-06-05
Add delegating decoder that retries on specified status codes

nmiyake:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/130)

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/errors/StatusCodeRetryableErrorDelegatingDecoder.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/errors/StatusCodeRetryableErrorDelegatingDecoder.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting.http.errors;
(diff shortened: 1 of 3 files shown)


nmiyake:
@markelliot @uschi2000 interested in hearing your feedback on this approach/idea.

General concept is that there are some error codes (particularly 404, 409 and 412) where we want to apply the failover/retry logic. This PR adds an `ErrorDecoder` implementation that takes a set of error codes that it should consider "Retryable", along with another `ErrorDecoder` that it wraps. If the error it processes has a response code that matches the provided one, it wraps exception returned by the delegate in a `RetryableException` so that the rest of the system will treat it as a retryable error.

Possible that there's a better/more canonical way to do this, so was interested in feedback.

uschi2000:
Generally, this makes sense to me. Few things to discuss:
- when the call fails eventually, does feign rethrow the original exception or a wrapped RetryableException?
- should retryability be based on http code or exception type? or both?
- is the whole idea of retrying much too coarse-grained to be useful? this is outside the scope of this PR, but long-term I believe we may want a combination of:
  - resources (server-side) specify idempotence
  - clients make an explicit per-call choice for retry/failover, possibly with a default, e.g.

```
MyService s = FeignClients.withRetries(5 /*times? seconds? */).createProxy(...);
s.foo(); // retries as per above
s.withoutRetries().bar(); // override retry configuration
```

nmiyake:
- This implementation always throws a wrapped one. I think the most Feign-native answer is that exceptions that should be retryable should directly implement `RetryableException` itself -- that way, no wrapping is needed. However, for my particular purposes I care more about the message/trace/causes than the actual exception type, so this particular implementation does wrap the inner exception (in many cases I think it also doesn't make sense for the original exception to implement `RetryableException`, which is very http-specific). One improvement to my implementation would be to check if the type of the exception produced by the delegate `ErrorDecoder` is already a `RetryableException` and to re-throw directly in that case.
- My idea for this library is that there are times when the client may want to specify retryability based on response code. If there's a desire to do it by exception type, the server or decoder can just return an exception that subclasses `RetryableException` and Feign will just handle it.
- Definitely possible. In this specific case, Axiom has a very specific contract that it will respond with 412 errors in some circumstances, but our suggested approach for clients is to re-try on those error codes. I'm encoding this logic in the client to handle it, but didn't know if this was something that might be true more generally.

PR #132, opened 2016-06-08
Update build.gradle to quiet Javadoc warnings in Java 8+

nmiyake:


build.gradle
```diff
@@ -41,6 +41,12 @@ subprojects {
 
     sourceCompatibility = 1.7
 
+    tasks.withType(Javadoc) {
+        // suppress Javadoc doclint warnings in Java 8+
+        if (!System.getProperty("java.version").startsWith("1.7")) {
```
markelliot, line 46: we know that the java version this compiles with is always Java 8, right?

nmiyake, line 46: How so? Right now source compatibility is set to be Java 1.7 and CircleCI runs the build using Java 7.

markelliot, line 46: Oh, I see, thanks.

PR #133, opened 2016-06-10
Clients must specify a user agent

jamesthomp:
This change will intentionally be breaking for clients when they update.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
@@ -56,6 +56,7 @@ public final class FeignClientFactory {
     private final Function<Optional<SSLSocketFactory>, Client> clientSupplier;
     private final BackoffStrategy backoffStrategy;
     private final Request.Options options;
+    private final UserAgentInterceptor userAgentInterceptor;
 
     private FeignClientFactory(
             Contract contract,
@@ -64,7 +65,8 @@ public final class FeignClientFactory {
             ErrorDecoder errorDecoder,
             Function<Optional<SSLSocketFactory>, Client> clientSupplier,
             BackoffStrategy backoffStrategy,
-            Request.Options options) {
+            Request.Options options,
+            String userAgent) {
         this.contract = contract;
         this.encoder = encoder;
         this.decoder = decoder;
@@ -72,6 +74,7 @@ public final class FeignClientFactory {
         this.clientSupplier = clientSupplier;
         this.backoffStrategy = backoffStrategy;
         this.options = options;
+        this.userAgentInterceptor = UserAgentInterceptor.of(userAgent);
     }
 
     /**
@@ -84,9 +87,10 @@ public final class FeignClientFactory {
             Encoder encoder,
             Decoder decoder,
             ErrorDecoder errorDecoder,
-            Function<Optional<SSLSocketFactory>, Client> clientSupplier) {
+            Function<Optional<SSLSocketFactory>, Client> clientSupplier,
+            String userAgent) {
         return new FeignClientFactory(contract, encoder, decoder, errorDecoder, clientSupplier,
-                NeverRetryingBackoffStrategy.INSTANCE, new Request.Options());
+                NeverRetryingBackoffStrategy.INSTANCE, new Request.Options(), userAgent);
     }
 
     /**
@@ -99,9 +103,10 @@ public final class FeignClientFactory {
             ErrorDecoder errorDecoder,
             Function<Optional<SSLSocketFactory>, Client> clientSupplier,
             BackoffStrategy backoffStrategy,
-            Request.Options options) {
+            Request.Options options,
+            String userAgent) {
         return new FeignClientFactory(
-                contract, encoder, decoder, errorDecoder, clientSupplier, backoffStrategy, options);
+                contract, encoder, decoder, errorDecoder, clientSupplier, backoffStrategy, options, userAgent);
     }
 
     /**
@@ -119,6 +124,7 @@ public final class FeignClientFactory {
                 .options(requestOptions)
                 .logger(new Slf4jLogger(FeignClients.class))
                 .logLevel(Level.BASIC)
+                .requestInterceptor(userAgentInterceptor)
                 .target(type, uri);
     }
 
@@ -151,6 +157,7 @@ public final class FeignClientFactory {
                 .options(requestOptions)
                 .logger(new Slf4jLogger(FeignClients.class))
                 .logLevel(Level.BASIC)
+                .requestInterceptor(userAgentInterceptor)
                 .target(target);
     }
 
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -47,68 +47,69 @@ public final class FeignClients {
      * Provides a {@link FeignClientFactory} with an {@link ObjectMapper} configured with {@link
      * com.fasterxml.jackson.datatype.guava.GuavaModule} and {@link com.fasterxml.jackson.datatype.jdk7.Jdk7Module}.
      */
-    public static FeignClientFactory standard() {
-        return standard(DEFAULT_TIMEOUT_OPTIONS);
+    public static FeignClientFactory standard(String userAgent) {
+        return standard(DEFAULT_TIMEOUT_OPTIONS, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} with an {@link ObjectMapper} configured with {@link
      * com.fasterxml.jackson.datatype.guava.GuavaModule} and {@link com.fasterxml.jackson.datatype.jdk7.Jdk7Module}.
      */
-    public static FeignClientFactory standard(Request.Options timeoutOptions) {
-        return withMapper(ObjectMappers.guavaJdk7(), timeoutOptions);
+    public static FeignClientFactory standard(Request.Options timeoutOptions, String userAgent) {
+        return withMapper(ObjectMappers.guavaJdk7(), timeoutOptions, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} compatible with jackson 2.4.
      */
-    public static FeignClientFactory standardJackson24() {
-        return standardJackson24(DEFAULT_TIMEOUT_OPTIONS);
+    public static FeignClientFactory standardJackson24(String userAgent) {
+        return standardJackson24(DEFAULT_TIMEOUT_OPTIONS, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} compatible with jackson 2.4.
      */
-    public static FeignClientFactory standardJackson24(Request.Options timeoutOptions) {
+    public static FeignClientFactory standardJackson24(Request.Options timeoutOptions, String userAgent) {
         return withEncoderAndDecoder(
                 new Jackson24Encoder(ObjectMappers.guavaJdk7()),
                 new JacksonDecoder(ObjectMappers.guavaJdk7()),
-                timeoutOptions);
+                timeoutOptions,
+                userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} with an unmodified {@link ObjectMapper}.
      */
-    public static FeignClientFactory vanilla() {
-        return vanilla(DEFAULT_TIMEOUT_OPTIONS);
+    public static FeignClientFactory vanilla(String userAgent) {
+        return vanilla(DEFAULT_TIMEOUT_OPTIONS, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} with an unmodified {@link ObjectMapper}.
      */
-    public static FeignClientFactory vanilla(Request.Options timeoutOptions) {
-        return withMapper(ObjectMappers.vanilla(), timeoutOptions);
+    public static FeignClientFactory vanilla(Request.Options timeoutOptions, String userAgent) {
+        return withMapper(ObjectMappers.vanilla(), timeoutOptions, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
      */
-    public static FeignClientFactory withMapper(ObjectMapper mapper) {
-        return withMapper(mapper, DEFAULT_TIMEOUT_OPTIONS);
+    public static FeignClientFactory withMapper(ObjectMapper mapper, String userAgent) {
+        return withMapper(mapper, DEFAULT_TIMEOUT_OPTIONS, userAgent);
     }
 
     /**
      * Provides a {@link FeignClientFactory} with the specified {@link ObjectMapper}.
      */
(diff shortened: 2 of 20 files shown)


jamesthomp:
@ellisjoe @markelliot @uschi2000

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
```diff
@@ -0,0 +1,40 @@
...
+
+import feign.RequestInterceptor;
+import feign.RequestTemplate;
+
+public final class UserAgentInterceptor implements RequestInterceptor {
+
+    private final String userAgent;
+
+    public UserAgentInterceptor(String userAgent) {
+        this.userAgent = userAgent;
```
markelliot, line 34: I think we probably want to validate that `userAgent` is a valid form

markelliot:
Can you do this for the retrofit stuff as well?

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
markelliot, line 34: http://stackoverflow.com/a/19028585/21441 says:

> HTTP header values may contain any ISO-8859-1 character, though many of them needed to be transmitted within quotes.

markelliot, line 34: ...or maybe it's not worthwhile. meh, this is probably fine.

ash211, line 34: Much better to do now than debug later.  Please validate with a tight character whitelist now and we can open up later if needed?

markelliot, line 34: Ok sure, I think a baseline could be “[A-Za-z0-9/.,_\s]+”

jamesthomp:
I've added a validation check and done this for retrofit also.

retrofit-clients/src/test/java/com/palantir/remoting/retrofit/RetryInterceptorTest.java
```diff
@@ -54,7 +54,7 @@ public void before() throws IOException {
     }
 
     @Test
-    public void testNoRetries() throws IOException {
+    public void testUserAgentSet() throws IOException {
```
markelliot: Think this might be an erroneous refactor

retrofit-clients/src/test/java/com/palantir/remoting/retrofit/UserAgentTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.IOException;
+import javax.net.ssl.SSLSocketFactory;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Matchers;
+
+public final class UserAgentTest {
```
markelliot, line 44: Would be good to see this with a MockWebServer to verify it actually sends the header in the request. (Basically same test as with Feign)

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+
+public final class UserAgentInterceptor implements RequestInterceptor {
+
+    private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9/\\.,_\\s]+");
+    private final String userAgent;
+
+    public UserAgentInterceptor(String userAgent) {
+        if (!VALID_USER_AGENT.matcher(userAgent).matches()) {
+            throw new IllegalArgumentException("User Agent " + userAgent + " is not valid.");
+        }
```
markelliot: Nit: a more common pattern for us is to use a statically imported Preconditions.checkArgument call, which gives you string substitution as well. We also typically write error messages of the form: "Invalid user agent: %s". 

(Can take this as is, but if you're going to correct other comments you might consider matching style here.)

@uschi2000: I wonder if we should have a baseline check that alerts on certain kinds of usage like this.

uschi2000: hmm, not sure we can write a reasonable checkstyle pattern for this, unfortunately.

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -84,9 +87,10 @@ public static FeignClientFactory of(
             Encoder encoder,
             Decoder decoder,
             ErrorDecoder errorDecoder,
-            Function<Optional<SSLSocketFactory>, Client> clientSupplier) {
+            Function<Optional<SSLSocketFactory>, Client> clientSupplier,
+            String userAgent) {
```
uschi2000, line 91: I think we need to release a few versions with both forms and `@Deprecated` the old one. otherwise upgrading may be very tedious, in particular with diamond-dependencies.

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -47,76 +47,78 @@ private FeignClients() {}
      * Provides a {@link FeignClientFactory} with an {@link ObjectMapper} configured with {@link
      * com.fasterxml.jackson.datatype.guava.GuavaModule} and {@link com.fasterxml.jackson.datatype.jdk7.Jdk7Module}.
      */
-    public static FeignClientFactory standard() {
-        return standard(DEFAULT_TIMEOUT_OPTIONS);
+    public static FeignClientFactory standard(String userAgent) {
```
uschi2000, line 55: needs javadoc for the user agent. please also give a example user agent string.

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
uschi2000, line 91: ^^ this comment really applies to FeignClients rather than this class, I guess.

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.regex.Pattern;
+
+public final class UserAgentInterceptor implements RequestInterceptor {
+
+    private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9/\\.,_\\s]+");
+    private final String userAgent;
+
+    public UserAgentInterceptor(String userAgent) {
+        if (!VALID_USER_AGENT.matcher(userAgent).matches()) {
+            throw new IllegalArgumentException("User Agent " + userAgent + " is not valid.");
```
uschi2000: need to say why it's not valid:

```
Preconditions.checkArgument(VALID_USER_AGENT.matcher(userAgent).matches(), 
        "User Agent must match pattern '%s': %s", USER_AGENT_PATTERN, userAgent);
```

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/RetrofitClientFactory.java
```diff
@@ -48,7 +49,10 @@ private static Client newHttpClient(Optional<SSLSocketFactory> sslSocketFactory,
         RetryInterceptor retryInterceptor = options.getMaxNumberRetries().isPresent()
                 ? new RetryInterceptor(options.getMaxNumberRetries().get())
                 : new RetryInterceptor();
+
+        UserAgentInterceptor userAgentInterceptor = UserAgentInterceptor.of(userAgent);
```
uschi2000: inline

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/UserAgentInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.squareup.okhttp.Interceptor;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.Response;
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+public final class UserAgentInterceptor implements Interceptor {
+
+    private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9/\\.,_\\s]+");
```
uschi2000: add comment:

```
// keep in sync with corresponding pattern in UserAgentInterceptor.java in http-remoting project
```

(and add reciprocal comment there)

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/UserAgentInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+    private final String userAgent;
+
+    public UserAgentInterceptor(String userAgent) {
+        if (!VALID_USER_AGENT.matcher(userAgent).matches()) {
```
uschi2000: dito

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/UserAgentInterceptor.java
```diff
@@ -0,0 +1,49 @@
...
+            throw new IllegalArgumentException("User Agent " + userAgent + " is not valid.");
+        }
+        this.userAgent = userAgent;
+    }
+
+    public static UserAgentInterceptor of(String userAgent) {
```
uschi2000, line 39: why do have this and the constructor public? choose one.

retrofit-clients/src/test/java/com/palantir/remoting/retrofit/UserAgentTest.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+
+    @Test
+    public void testUserAgent_invalidUserAgentThrows() throws InterruptedException {
+        expectedException.expect(IllegalArgumentException.class);
+
+        OkHttpClientOptions okHttpClientOptions = OkHttpClientOptions.builder().build();
+
+        RetrofitClientFactory.createProxy(Optional.<SSLSocketFactory>absent(), "", String.class, okHttpClientOptions,
```
uschi2000, line 114: should check the message because all sorts of things can throw in there.

uschi2000:
couple of nitpicks + the question whether to deprecate or remove the old method signatures? @markelliot

markelliot:
I think we can release an RC with the break. No projects have a shared remoting dependency, and this doesn't actually affect behavior besides fixing logs on receiving services.

uschi2000:
Deal.

jamesthomp:
Thanks for the comments! I believe they have now all been addressed.

http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
```diff
@@ -36,6 +36,11 @@
  * Static factory methods for producing common configurations of {@link FeignClientFactory}, which in turn may be used
  * to create HTTP proxies for HTTP remoting clients. The returned instances serialize server-side exceptions as JSON and
  * decode any 204 response as an {@link Optional#absent} in case the proxied interface is of type {@link Optional}.
+ * <p>
+ *     All factories take a User Agent and this will be embedded as the User Agent header for all requests.
+ *     For services, recommended user agents are of the form: ServiceName (Version), e.g. MyServer (1.2.3)
+ *     For services that run multiple instances, recommended user agents are of the form:
+ *       ServiceName/InstanceId (Version), e.g. MyServer/12 (1.2.3)
```
markelliot: The indents here seem odd, can we left align?

markelliot: Also might want to use `{@code ...}` to differentiate the recommendations stylistically...e.g. `{@code ServiceName (Version)}`

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
```diff
@@ -0,0 +1,47 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import feign.RequestInterceptor;
+import feign.RequestTemplate;
+import java.util.regex.Pattern;
+
+public final class UserAgentInterceptor implements RequestInterceptor {
+
+    // keep in sync with corresponding pattern in UserAgentInterceptor.java in retrofit-clients project
+    private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9/\\.,_\\s]+");
```
markelliot: doh, this should include parens per recommended format described above

http-clients/src/test/java/com/palantir/remoting/http/UserAgentTest.java
```diff
@@ -0,0 +1,80 @@
...
+        RecordedRequest request = server.takeRequest();
+        assertThat(request.getHeader("User-Agent"), is(USER_AGENT));
+    }
+
+    @Test
+    public void testUserAgent_invalidUserAgentThrows() throws InterruptedException {
+        expectedException.expect(IllegalArgumentException.class);
+        expectedException.expectMessage(is("User Agent must match pattern '[A-Za-z0-9/\\.,_\\s]+': ("));
+
+        FeignClients.standard("(");
```
markelliot: and oddly enough this should pass (per comments above)

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/RetrofitClientFactory.java
```diff
@@ -29,12 +29,18 @@
  * Utilities to help create Retrofit proxies. Feign clients should be preferred except in cases where proxies must
  * support file upload and download. Read and write timeouts are customizable in order to allow arbitrary sized file
  * uploads/downloads.
+ * <p>
+ *     All proxies take a User Agent and this will be embedded as the User Agent header for all requests.
+ *     For services, recommended user agents are of the form: ServiceName (Version), e.g. MyServer (1.2.3)
+ *     For services that run multiple instances, recommended user agents are of the form:
+ *       ServiceName/InstanceId (Version), e.g. MyServer/12 (1.2.3)
```
markelliot: same deal with indents here

retrofit-clients/src/main/java/com/palantir/remoting/retrofit/UserAgentInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+import com.squareup.okhttp.Interceptor;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.Response;
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+public final class UserAgentInterceptor implements Interceptor {
+
+    // keep in sync with corresponding pattern in UserAgentInterceptor.java in http-clients project
+    private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9/\\.,_\\s]+");
```
markelliot: and parens here.

markelliot:
few more things, sorry

markelliot:
thanks!

PR #135, opened 2016-06-14
User agent not forced

jamesthomp:
Bring back the old methods (marked as deprecated) to allow for diamond dependencies to get updated.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClients.java
@@ -49,11 +49,31 @@ public final class FeignClients {
     private FeignClients() {}
 
     /**
+     * @deprecated Clients should specify a user agent. This method will be removed when clients have updated.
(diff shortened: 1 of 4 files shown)


markelliot:
what's an example of a diamond dependency where this hits us?

uschi2000:
foundryfs & multipass-bundle

markelliot:
I'm okay with this, but maybe as an alternative we should just make the FeignClients deprecated and move to FeignClients2 (or some less-grossly named thing?)

jamesthomp:
I agree that it would be nice to depreciate the class and not have a bunch of methods be deprecated, but I also think that FeignClients2 (or any name) is pretty gross. I don't feel strongly (so I'm happy to make the change to FeignClients2) but I think the current set up is nicer.

markelliot:
Either way, can live with this as is, too.

uschi2000:
Yep, let's leave the deprecated methods.

PR #137, opened 2016-06-21
Jellis/user agent help

ellisjoe:


ellisjoe:
@uschi2000 @markelliot

http-clients/build.gradle
```diff
@@ -21,6 +21,7 @@ dependencies {
     compile libs.wsRsApi
 
     testCompile project(':http-servers')
+    testCompile files('lib/VersionTest.jar')
```
uschi2000, line 24: can you not use one of the jars that are on the classpath anyway?

ellisjoe, line 24: I'd have to update the test any time the version changes.

uschi2000, line 24: or alternatively generate this thing somehow?

uschi2000, line 24: regex for the version?

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
```diff
@@ -0,0 +1,43 @@
...
+    public static String getUserAgent(String userAgent, String version) {
+        return String.format("%s (%s)", userAgent, version);
+    }
+
+    /**
+     * Constructs a user agent from the {@link java.util.jar.Attributes.Name#IMPLEMENTATION_TITLE} and {@link
+     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class. The default value for
+     * both properties is "{@value DEFAULT_VALUE}".
+     */
+    public static String getUserAgent(Class<?> clazz) {
```
markelliot: maybe `getForClass`?

markelliot: or `fromClass`

http-clients/src/test/java/com/palantir/remoting/http/UserAgentsTest.java
```diff
@@ -0,0 +1,38 @@
...
+public final class UserAgentsTest {
+
+    @Test
+    public void testGetUserAgent_format() {
+        assertThat(UserAgents.getUserAgent("name", "version"), is("name (version)"));
+    }
+
+    @Test
+    public void testGetUserAgent_fromExternalVersionTestJar() throws IOException {
+        assertThat(UserAgents.getUserAgent(VersionTest.class), is(String.format("test-name (test-version)")));
```
markelliot: `String.format` is superfluous

http-clients/build.gradle
ellisjoe, line 24: Not a fan of generating since it makes running tests a multistep process. regexing the user agent generated using something like Feign could work, just seems brittle since we're depending on the format a non-public non-static string

markelliot, line 24: You could make this an unpublished project with a fixed version…but it'd be yet one more project.

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
ellisjoe: done

http-clients/src/test/java/com/palantir/remoting/http/UserAgentsTest.java
ellisjoe: done

http-clients/build.gradle
ellisjoe, line 24: and to get it working with unit tests id have to build it before running tests

markelliot, line 24: You could depend on it as project(':version-test')…

ellisjoe, line 24: tried that with SerializableError but since it's not packaged before the test runs you end up getting null back for the version and name

PR #138, opened 2016-06-21
Upgrade to Feign 8.17.0 (and, transitively, OkHttp 3.2.0)

uschi2000:


uschi2000:
@robert3005 @mingyukim @markelliot Any concerns with this upgrade, in particular in Spark-land?

robert3005:
not really concerned. Those libraries are pretty self contained. Only major thing would be jackson dependency changes but spark 2.0 is just around the corner. I went through okhttp and feign dependencies and briefly crosschecked against spark and couldn't find anything.

PR #140, opened 2016-06-22
Throw when UserAgents can't find title or version

jared2501:
Modifies UserAgents in http-remoting to throw an exception if it can’t find a project name + version instead of the current behaviour of saying “unknown”. The reasoning behind this change is that in the world in which UserAgents returns unknown I have to write an integration test to capture the case in which my build stops putting the information in the manifest file. If, instead, UserAgents throws an exception, then any integration test will capture this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # Gradle
 *.class
-*.jar
 .gradle
 build/
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # Gradle
 *.class
+*.jar
+!gradle/wrapper/gradle-wrapper.jar
+!http-clients/lib/VersionTest.jar
 .gradle
 build/
 
--- a/http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
@@ -16,11 +16,9 @@
 
 package com.palantir.remoting.http;
 
-import com.google.common.base.Optional;
+import static com.google.common.base.Preconditions.checkArgument;
 
 public final class UserAgents {
-
-    public static final String DEFAULT_VALUE = "unknown";
 
     private UserAgents() {}
 
@@ -30,13 +28,20 @@ public final class UserAgents {
 
     /**
      * Constructs a user agent from the {@link java.util.jar.Attributes.Name#IMPLEMENTATION_TITLE} and {@link
-     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class. The default value for
-     * both properties is "{@value DEFAULT_VALUE}".
+     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class.
+     * <p>
+     * Typically, the title and version are extracted from {@code MANIFEST.MF} entries of the Jar package containing the
+     * given class.
+     *
+     * @throws IllegalArgumentException if the implementation title or version are not provided in the class's {@link
+     * Package}
      */
     public static String fromClass(Class<?> clazz) {
         Package classPackage = clazz.getPackage();
-        String userAgent = Optional.fromNullable(classPackage.getImplementationTitle()).or(DEFAULT_VALUE);
-        String version = Optional.fromNullable(classPackage.getImplementationVersion()).or(DEFAULT_VALUE);
+        String userAgent = classPackage.getImplementationTitle();
+        String version = classPackage.getImplementationVersion();
+        checkArgument(userAgent != null, "Cannot load user agent from implementation title for class: %s", clazz);
+        checkArgument(version != null, "Cannot load version from implementation version for class: %s", clazz);
         return getUserAgent(userAgent, version);
     }
 
--- a/http-clients/src/test/java/com/palantir/remoting/http/UserAgentsTest.java
+++ b/http-clients/src/test/java/com/palantir/remoting/http/UserAgentsTest.java
@@ -21,9 +21,14 @@ import static org.junit.Assert.assertThat;
 
 import com.palantir.VersionTest;
 import java.io.IOException;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 public final class UserAgentsTest {
+
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
 
     @Test
     public void testGetUserAgent_format() {
@@ -33,6 +38,13 @@ public final class UserAgentsTest {
     @Test
     public void testGetUserAgent_fromExternalVersionTestJar() throws IOException {
         assertThat(UserAgents.fromClass(VersionTest.class), is("test-name (test-version)"));
+    }
+
+    @Test
+    public void testGetUserAgent_fromExternalVersionWithoutManifestTestJar() throws IOException {
+        expectedException.expect(IllegalArgumentException.class);
+        expectedException.expectMessage("Cannot load user agent from implementation title");
+        UserAgents.fromClass(UserAgentsTest.class);
     }
 
 }


jared2501:
Also, I just realized that this won't add the InstanceId as suggested: `ServiceName/InstanceId (Version)` where InstanceId identifies the instance when a service is run in clustered mode....

uschi2000:
will take a look once it builds ;)

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
```diff
@@ -30,13 +28,16 @@ public static String getUserAgent(String userAgent, String version) {
 
     /**
      * Constructs a user agent from the {@link java.util.jar.Attributes.Name#IMPLEMENTATION_TITLE} and {@link
-     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class. The default value for
-     * both properties is "{@value DEFAULT_VALUE}".
+     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class.
+     *
+     * @throws IllegalArgumentException if the implementation title or version are not provided
```
uschi2000: "are not provided in the class's {@link Package}"

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
```diff
@@ -30,13 +28,16 @@ public static String getUserAgent(String userAgent, String version) {
...
      */
     public static String fromClass(Class<?> clazz) {
         Package classPackage = clazz.getPackage();
-        String userAgent = Optional.fromNullable(classPackage.getImplementationTitle()).or(DEFAULT_VALUE);
-        String version = Optional.fromNullable(classPackage.getImplementationVersion()).or(DEFAULT_VALUE);
+        String userAgent = classPackage.getImplementationTitle();
```
uschi2000, line 41: I personally like this style, you?

```
String userAgent = checkNotNull(classPackage.getImplementationTitle(), "Cannot determine user agent from package of class " + clazz);
```

version-test/src/main/java/com/palantir/version/test/without/versions/VersionTestWithoutVersions.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.version.test.without.versions;
+
+public final class VersionTestWithoutVersions {
```
uschi2000: just use the test class itself?

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
```diff
@@ -30,13 +28,16 @@ public static String getUserAgent(String userAgent, String version) {
 
     /**
      * Constructs a user agent from the {@link java.util.jar.Attributes.Name#IMPLEMENTATION_TITLE} and {@link
-     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class. The default value for
-     * both properties is "{@value DEFAULT_VALUE}".
+     * java.util.jar.Attributes.Name#IMPLEMENTATION_VERSION} of the package of the provided class.
```
uschi2000, line 31: can you add a little hint on how to get the correct information into that package?
"Typically, the title and version are extracted from {@code MANIFEST.MF} entries of the Jar package containing the given class."

version-test/src/main/java/com/palantir/version/test/without/versions/VersionTestWithoutVersions.java
ellisjoe: Yeah, you should at least be able to use a class from a another project

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
jared2501, line 41: trouble is that you throw an NPE, when instead i think you want to throw `IllegalArgumentException` and i think this depends on the argument that you pass in. I.e. the class you passed in is illegal because it doesnt have anything defined in the MANIFEST

version-test/src/main/java/com/palantir/version/test/without/versions/VersionTestWithoutVersions.java
jared2501: i decided not to since this might cause flakes if someone changes the build.gradle to include the name/version in the manifest (lol silly I know).. will change back though

uschi2000: Does Gradle run the tests against the jars or against the classes?

jared2501: Not actually sure..

ellisjoe: quick test looks like both intellij and gradle run against the classes and not the jars, even though the test task apparently depends on the jar task...

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
jared2501: done

version-test/src/main/java/com/palantir/version/test/without/versions/VersionTestWithoutVersions.java
jared2501: done

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
jared2501, line 31: done

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
```diff
@@ -30,13 +28,20 @@ public static String getUserAgent(String userAgent, String version) {
...
+     * @throws IllegalArgumentException if the implementation title or version are not provided in the class's {@link
+     * Package}
      */
     public static String fromClass(Class<?> clazz) {
         Package classPackage = clazz.getPackage();
-        String userAgent = Optional.fromNullable(classPackage.getImplementationTitle()).or(DEFAULT_VALUE);
-        String version = Optional.fromNullable(classPackage.getImplementationVersion()).or(DEFAULT_VALUE);
+        String userAgent = classPackage.getImplementationTitle();
+        String version = classPackage.getImplementationVersion();
+        checkArgument(userAgent != null, "Cannot load user agent from implementation title");
```
markelliot: sorry to be a pain, but I think these should say `...from implementation title for class: %s", clazz)`

jared2501: done

.gitignore
```diff
@@ -1,6 +1,5 @@
 # Gradle
 *.class
-*.jar
```
uschi2000, line 3: think we can add this back?

http-clients/src/main/java/com/palantir/remoting/http/UserAgents.java
uschi2000, line 41: ok

uschi2000:
one nitpick (.gitignore), otherwise +1

.gitignore
jared2501, line 3: > Jared Newman [8:03 PM]  
> hey rob - so adding back *.jar on that PR. I think it's probably best to remove it cause it technically means that the jars thats already in there is gitignored (edited)
> 
> Robert Fink [8:04 PM]  
> oh, hmm
> [8:04]  
> ok
> [8:04]  
> sure

jared2501:
i think this is ready to merge?

.gitignore
markelliot, line 3: I think we could ignore *.jar but use an exception for the two we need

jared2501, line 3: not a huge fan of this pattern, but done :)

ellisjoe:
slight issue I realized tonight: this change makes it so you'll only be able to start up services if they've been packaged into jars, which means no starting from within your IDE or from gradle. I don't think we want to force that and I don't have a better solution than sticking in default values like we were previously

markelliot:
Ah, yeah. We can do the thing with having a system property override for debugging like we do with EFS for Circle.

PR #141, opened 2016-06-22
Introduce ClientSupplier in FeignClientFactory

uschi2000:
In preparation for Brave PR

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/ClientSupplier.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/ClientSupplier.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


http-clients/src/main/java/com/palantir/remoting/http/ClientSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.common.base.Optional;
+import feign.Client;
+import javax.net.ssl.SSLSocketFactory;
+
+/**
+ * Given an optional {@link javax.net.ssl.SSLSocketFactory} and a user agent, creates and returns a {@link feign.Client
+ * Feign client}.
+ */
+public interface ClientSupplier {
+    Client createClient(Optional<SSLSocketFactory> sslSocketFactory, String userAgent);
```
markelliot, line 28: this should probably take an object rather than args so we don't have to break the API every time we need to add more args

http-clients/src/main/java/com/palantir/remoting/http/ClientSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.google.common.base.Optional;
+import feign.Client;
+import javax.net.ssl.SSLSocketFactory;
+
+/**
+ * Given an optional {@link javax.net.ssl.SSLSocketFactory} and a user agent, creates and returns a {@link feign.Client
+ * Feign client}.
+ */
+public interface ClientSupplier {
```
markelliot, line 27: it's a bit of a nit, but this isn't actually a `Supplier` (the function it replaces wasn't either =/)

http-clients/src/main/java/com/palantir/remoting/http/ClientSupplier.java
uschi2000, line 28: spoke offline, will find an alternative solution mid-term

PR #143, opened 2016-06-23
Upgrade to OkHttp3 and Retrofit2 [breaking change]

uschi2000:


uschi2000:
This is a precondition for adding Brave tracing support to retrofit-clients. Before this merges, I'd like to make one more release of http-remoting with tracing support for Feign clients, but without the breaking change in retrofit-clients.

@markelliot @ellisjoe @pnepywoda

retrofit-clients/build.gradle
```diff
@@ -8,7 +8,8 @@ dependencies {
     // ssl-config is not used in this project, but we depend on it here in order to provide
     // it transitively to users of retrofit-clients as a convenience.
     compile project(':ssl-config')
-    compile libs.retrofit
+    compile libs.retrofit.retrofit
+    compile libs.retrofit.gson
```
markelliot, line 12: we really should convert this to the nebula dependency thing

uschi2000, line 12: yep, will do

pnepywoda:
pushing down the error handler to okhttp instead of forking for retrofit/feign? sounds great to me!!

markelliot:
👍

leeavital:
Maybe a silly question, but some projects publish a client which depends on http-remoting because their client does something more than simply instrumenting a JAX-RS interface. Would depending on okhttp3 _and_ one of those project clients which depend on okhttp2 cause problems?

uschi2000:
The way I understand it at this point (although I haven't tried it): you
should be able to depend on okhttp2 and okhttp3 at the same time.

markelliot:
good to take this change now?

uschi2000:
Yep. Note that this is a pretty big break for most users of the
http-retrofit package: they'd need to change their retrofit API
definitions/annotations.

markelliot:
In that case, maybe we need to add a retrofit2-clients project to preserve upgradability?

uschi2000:
Yep. I really do want people to upgrade soon though to support tracing.

markelliot:
Ok, I assume the complex thing here is how people release new APIs that work for both old retrofit and new retrofit; so we can release with a new project + deprecate the old one, then for 1.0 blow away the deprecated stuff?

uschi2000:
Yep, correct. There may not be a way around duplicating the APIs for
retrofit1 and retrofit2 clients. Feign/JAX-RS clients don't have the same
problem because JAX-RS annotations are versioned independently of Feign.

markelliot:
Is it possible to annotate the same interface with both Retrofit1 and Retrofit2 annotations?

uschi2000:
Unclear, have to check. Either way, we need to publish both
retrofit-clients and retrofit2-clients here.

uschi2000:
Will split up in two more obviously separated PRs.

PR #146, opened 2016-06-27
Add public `createTrustManagers` factory

markelliot:


markelliot:
@uschi2000 I think we need this to play nicely with new okhttp

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -63,7 +63,7 @@ public static SSLContext createSslContext(SslConfiguration config) {
         }
     }
 
-    private static TrustManagerFactory createTrustManagerFactory(
+    public static TrustManagerFactory createTrustManagerFactory(
```
uschi2000: the trustStorePath parameter is non-trivial (can be file or directory, depending on the type), so this will need docs, unfortunately.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
```diff
@@ -63,7 +61,17 @@ public static SSLContext createSslContext(SslConfiguration config) {
         }
     }
 
-    public static TrustManagerFactory createTrustManagerFactory(
+    /**
+     * Create a {@link TrustManager[]} initialized from the provided configuration.
+     *
+     * @param config an {@link SslConfiguration} describing at least the trust store configuration
+     * @return an {@link TrustManager[]} according to the input configuration
```
uschi2000, line 68: ok, so I guess I'll have to add docs to SslConfiguration now ;)

uschi2000:
👍

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
markelliot: Refactored to make this match the other methods in the class in style and docs.

ssl-config/src/main/java/com/palantir/remoting/ssl/SslSocketFactories.java
markelliot, line 68: yep, but we're already doing this elsewhere, so we can leave that for another PR

PR #152, opened 2016-06-29
Make ObjectMappers methods return copy of mappers

### LINKED ISSUE #151: ObjectMappers methods return shared mutable ObjectMapper
`ObjectMappers.vanilla()` and `ObjectMappers.guavaJdk7()` both directly return a static variable that is a mutable object, which means that if a caller changes the configuration of the returned object, the change is shared across all instances and for all future callers.

These methods should be changed to hand back a copy of the mappers instead so that callers are free to mutate the returned values without impacting other instances or future callers.

nmiyake:
Fixes #151

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/ObjectMappers.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/ObjectMappers.java
@@ -22,15 +22,31 @@ import com.fasterxml.jackson.datatype.guava.GuavaModule;
 
 public final class ObjectMappers {
 
+    /**
+     * Default {@link ObjectMapper}. This is a single shared instance that can be accessed directly to minimize
+     * allocations and memory use and should not be mutated.
+     */
     static final ObjectMapper VANILLA_MAPPER = vanilla();
+
+    /**
+     * {@link ObjectMapper} that is configured with the Guava module and the JDK 7 module (if present). This is a single
+     * shared instance that can be accessed directly to minimize allocations and memory use and should not be mutated.
+     */
     static final ObjectMapper GUAVA_JDK7_MAPPER = guavaJdk7();
 
     private ObjectMappers() {}
 
+    /**
+     * Returns a newly allocated default {@link ObjectMapper}.
+     */
     public static ObjectMapper vanilla() {
         return new ObjectMapper();
     }
 
+    /**
+     * Returns a newly allocated {@link ObjectMapper} that is configured with the Guava module and the JDK 7 module
+     * (if present).
+     */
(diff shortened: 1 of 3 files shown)


nmiyake:
@uschi2000 @markelliot

markelliot:
urgh, I think we need to find a way to make these immutable rather than pass around copies; we were trying to control memory load a bit by sharing objects.

nmiyake:
Hmm, in that case, maybe it would make sense to make the shared instances package-protected and keep public methods that hand back copies? Just seems pretty dangerous/error-prone to have a key object open to modification from anywhere within a JVM.

`ObjectMapper` is non-final so it's technically possible to make an unmodifiable wrapper, but that would be a bigger change with more possible implications (especially if someone does actually want a copy that they can mutate/configure for their own use)

uschi2000:
I like this idea: "sense to make the shared instances package-protected and keep public methods that hand back copies"

nmiyake:
Implemented suggested variation. Made it so that static variables are package-protected and the internal methods use the variables directly -- I think this makes it much more apparent when reading code that these are shared instances.

uschi2000:
Will look tomorrow

http-clients/src/main/java/com/palantir/remoting/http/ObjectMappers.java
```diff
@@ -22,32 +22,28 @@
...
 public final class ObjectMappers {
 
-    private static final ObjectMapper VANILLA_MAPPER = new ObjectMapper();
+    static final ObjectMapper VANILLA_MAPPER = vanilla();
+    static final ObjectMapper GUAVA_JDK7_MAPPER = guavaJdk7();
 
-    private static final ObjectMapper GUAVA_JDK7_MAPPER;
+    private ObjectMappers() {}
+
+    public static ObjectMapper vanilla() {
```
markelliot, line 42: - some Java doc?

http-clients/src/main/java/com/palantir/remoting/http/ObjectMappers.java
```diff
@@ -22,32 +22,28 @@
 
 public final class ObjectMappers {
 
-    private static final ObjectMapper VANILLA_MAPPER = new ObjectMapper();
+    static final ObjectMapper VANILLA_MAPPER = vanilla();
+    static final ObjectMapper GUAVA_JDK7_MAPPER = guavaJdk7();
```
markelliot, line 35: Worth adding some comments here

nmiyake:
Added comments

PR #156, opened 2016-07-08
Use Case insensitive header comparison for newer OkHttp compatibility

a-k-g:


a-k-g:
This is needed because from OkHttp 3.3.0 onwards, `Headers.toMultimap()` lowers the case for header names. This will be fixed in 3.4.0 as `Headers.toMultimap` will return a `TreeMap` which does case insensitive get. However, Feign again wraps the response headers in https://github.com/Netflix/feign/blob/master/core/src/main/java/feign/Response.java#L50-L51.

In general, `http-remoting` shouldn't be making assumptions about the header's case sensitivity.

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
```diff
@@ -40,8 +41,8 @@ public static UserAgentInterceptor of(String userAgent) {
 
     @Override
     public void apply(RequestTemplate template) {
-        if (!template.headers().containsKey("User-Agent")) {
-            template.header("User-Agent", userAgent);
+        if (!HeaderAccessUtils.caseInsensitiveContains(template.headers(), HttpHeaders.USER_AGENT)) {
```
uschi2000, line 44: needs test

http-clients/src/main/java/com/palantir/remoting/http/HeaderAccessUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+package com.palantir.remoting.http;
+
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+public final class HeaderAccessUtils {
```
uschi2000, line 30: add javadoc explaining why we need it and when we can get rid of it again.

http-clients/src/main/java/com/palantir/remoting/http/HeaderAccessUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static Collection<String> caseInsensitiveGet(Map<String, Collection<String>> headers, String header) {
+        return caseInsensitiveMultimapOf(headers).get(header);
+    }
+
+    private static Map<String, List<String>> caseInsensitiveMultimapOf(Map<String, Collection<String>> headers) {
+        Map<String, List<String>> result = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+        for (Map.Entry<String, Collection<String>> entry : headers.entrySet()) {
+            String headerName = entry.getKey();
+            List<String> values = result.get(headerName);
```
uschi2000: logic looks right, but somewhat unintuitive. how about something like:

```
if (result.contains(headerName)) {
  result.get(headerName).addAll(entry.getValue());
} else {
  result.put(Lists.newLinkedList(entry.getValue()));
}
```

small performance sacrifice, but I think I'd bias towards legibility here.

http-clients/src/main/java/com/palantir/remoting/http/HeaderAccessUtils.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static boolean caseInsensitiveContains(Map<String, Collection<String>> headers, String header) {
+        return caseInsensitiveGet(headers, header) != null;
+    }
+
+    public static Collection<String> caseInsensitiveGet(Map<String, Collection<String>> headers, String header) {
+        return caseInsensitiveMultimapOf(headers).get(header);
+    }
+
+    private static Map<String, List<String>> caseInsensitiveMultimapOf(Map<String, Collection<String>> headers) {
```
uschi2000:
```
// returns a case-insensitive TreeMap in which the values for all keys equal modulo case are merged.
```

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
markelliot, line 44: This is a hot code path, I'm concerned that we're making a map every time we traverse this path.

uschi2000:
few nitpicks, otherwise this is fine. thanks Aakash!

http-clients/src/main/java/com/palantir/remoting/http/UserAgentInterceptor.java
markelliot, line 44: to expand on that, the test we could do for `caseInsensitiveContains` should do the loop but not the map creation.

uschi2000, line 44: +1

http-clients/src/main/java/com/palantir/remoting/http/HeaderAccessUtils.java
```diff
@@ -22,27 +22,38 @@
...
+        for (String headerName : headers.keySet()) {
+            if (headerName.equalsIgnoreCase(header)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     public static Collection<String> caseInsensitiveGet(Map<String, Collection<String>> headers, String header) {
         return caseInsensitiveMultimapOf(headers).get(header);
```
markelliot: sorry, didn't catch this earlier, but I think making this also short-circuit as soon as it finds the match is probably better given where these things get called.

a-k-g: Maybe I'm missing something, but I'm not sure how we can short-circuit this.

a-k-g: We can't return as soon as a match is found, we'd still have to go through the entire entrySet to look for multiple values. We could however avoid creation of a TreeMap completely.

markelliot: Got it; yeah, I'd avoid making the TreeMap.

markelliot:
lgtm; thanks for iteration

uschi2000:
Thanks Aakash!

PR #158, opened 2016-07-11
Proxy support

iamthomaspowell:
Not ready for merging yet but aiming to get feedback on this a solution to providing proxy support to http-remoting. The main motivation for switching ProxyConfiguration to ProxyConfigurationProvider is so that we can easily setup a provider chain. If this is used in standalone libraries it should make it to use environment variables or JVM args to configure the proxy and minimize the amount of custom code required.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
+++ b/http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
@@ -32,9 +32,12 @@ import com.google.common.collect.Sets;
 import com.google.common.net.HttpHeaders;
 import com.google.common.net.InetAddresses;
 import com.palantir.config.service.BasicCredentials;
-import com.palantir.config.service.ProxyConfiguration;
 import com.palantir.config.service.ServiceConfiguration;
 import com.palantir.config.service.ServiceDiscoveryConfiguration;
+import com.palantir.config.service.proxy.EmptyProxyConfigurationProvider;
+import com.palantir.config.service.proxy.ProxyConfiguration;
+import com.palantir.config.service.proxy.ProxyConfigurationProvider;
+import com.palantir.config.service.proxy.WrapperProxyConfigurationProvider;
 import com.palantir.ext.brave.SlfLoggingSpanCollector;
 import com.palantir.remoting.ssl.SslConfiguration;
 import com.palantir.remoting.ssl.SslSocketFactories;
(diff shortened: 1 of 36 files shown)


pnepywoda:
If I followed the discussion correctly - this PR is only for supporting older versions of okhttp with proxy support and it's fixed in newer versions of okhttp? (this is a pretty huge PR, wouldn't it be easier to workaround/fix the issues with upgrading okhttp?)

iamthomaspowell:
My understanding was the the change to okhttp was to fix an underlying bug when authenticating with a proxy. However, this PR is more about actually exposing the ability to configure the proxy.

http-clients/src/main/java/com/palantir/remoting/http/FeignClientFactory.java
```diff
@@ -117,68 +135,129 @@ public static FeignClientFactory of(
...
-        return Feign.builder()
+        return createProxy(sslSocketFactory, uris, type, defaultProxyConfigurationProvider, requestOptions);
+    }
+
+    /**
+     * Constructs a dynamic proxy for the specified type, using the supplied SSL factory if is present, proxy
+     * configuration provider, and {@link okhttp3.OkHttpClient} wrapped in {@link feign.okhttp.OkHttpClient}.
+     * A {@link FailoverFeignTarget} is used to cycle through the given uris on failure if there's more than one uri.
+     */
+    public <T> T createProxy(Optional<SSLSocketFactory> sslSocketFactory, Set<String> uris, Class<T> type,
```
pnepywoda, line 219: there's a ton of createProxy methods now - think we should move to a builder pattern?

markelliot, line 219: Yeah, we need to fix the API before this change, it's no good as is. 

Also need to seriously consider what we're really doing with Feign at this point, and likely hide its use as implementation detail.

pnepywoda, line 219: Is this alluring to discussions around Conjure and switching from feign to retrofit all-together?

markelliot, line 219: Rob and I have been discussing at least hiding Feign from the API in this regardless. 

Conjure gives us optionality for change, but isn't a golden ticket. Certainly things are simpler if we hide the impl details, though.

uschi2000, line 219: I will propose a new API in the next couple of hours, shooting for an RC
release with proxy support by EOD.

uschi2000:
closing in lieu of #165 #166

PR #165, opened 2016-07-18
Rfink/jaxrs clients proxy

uschi2000:


uschi2000:
@iamthomaspowell

service-config/src/main/java/com/palantir/config/service/BasicCredentials.java
```diff
@@ -0,0 +1,46 @@
...
+    @Value.Parameter
+    public abstract String username();
+
+    @Value.Parameter
+    public abstract String password();
+
+    @Check
+    protected final void check() {
+        Preconditions.checkArgument(!username().isEmpty(), "Username must not be empty");
+        Preconditions.checkArgument(!password().isEmpty(), "Password must not be empty");
```
markelliot: I'm not sure we should enforce non-empty constraints

markelliot:
lgtm besides nit on username/password: my guess is that at least one of these we run into has just a username or just a password, some day.

uschi2000:
I guess that ":", "foo:", and ":bar" are all valid basic auth creds?

uschi2000:
rebasing onto develop.

PR #166, opened 2016-07-18
Add proxy support to retrofit2-clients

uschi2000:
Preliminary PR against feature branch. fyi @iamthomaspowell


markelliot:
don't want to also refactor this first?

markelliot:
doubt anyone is using this yet, so easier to break

uschi2000:
yep, can do.

PR #167, opened 2016-07-19
Refactor jaxrs client API

uschi2000:


uschi2000:
Alternative API for jaxrs clients. Note that JaxRsClientConfig is (1) completely independent from jaxrs (and could be reused for retrofit2 clients), (2) an adapter for ServiceConfiguration that makes fewer assumptions on things being located on disk. Thoughts, @markelliot

markelliot:
This is fine; I was thinking you could also just mirror the Client/ClientBuilder class setup for retrofit2.

uschi2000:
Yep, that was my plan. Do you think the config/clientbuilder will diverge
between jaxrs-type and retrofit-type clients?
  R.

markelliot:
Nope, should be the same, I think. It's desirable for it to be the same, at least.

uschi2000:
Agree.

uschi2000:
Circle down?

PR #168, opened 2016-07-19
Custom GsonConverterFactory in retrofit2-clients

iamthomaspowell:
Allow a custom GsonConverterFactory to be provided when constructing a retrofit2 proxy. This is useful when configuring custom type adapters

### DIFF
--- a/error-handling/build.gradle
+++ b/error-handling/build.gradle
@@ -14,5 +14,5 @@ dependencies {
     testCompile libs.hamcrest
     testCompile libs.junit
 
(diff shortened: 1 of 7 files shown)


markelliot:
Hey Tom, we generally want to get away from custom converters as a practice, can you comment here (or ping me offline if needed) with some examples of what you're trying to do and we can maybe help you through the Jackson annotations.

uschi2000:
Will move to Jackson instead of GSON.

PR #169, opened 2016-07-19
Switch retrofit2-clients from using gson to jackson.

iamthomaspowell:


markelliot:
@uschi2000, did you intentionally pick Gson over Jackson? Think we should've taken Jackson.

uschi2000:
I don't actually remember who made the first retrofit client. If it was me, then it was not intentional.

uschi2000:
Thomas, which Jackson version does this bring in?

uschi2000:
In related news, I really need to change to nebula for dependencies so we have some accountability around library versions.

iamthomaspowell:
This brings in Jackson 2.7.2. Seems that we explicitly specify 2.6.1 so I can force this version

uschi2000:
No, that's fine. Thanks Thomas.

PR #170, opened 2016-07-20
Circle.yml: JVM memory<=3.5GB, Gradle parallel+daemon builds

uschi2000:


circle.yml
```diff
@@ -3,22 +3,15 @@ machine:
...
   environment:
     TERM: dumb
-    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx2048m -XX:+HeapDumpOnOutOfMemoryError"'
-
-general:
-  artifacts:
-    - error-handling/build/libs
-    - http-clients/build/libs
-    - http-servers/build/libs
-    - ssl-config/build/libs
```
markelliot, line 13: No desire to keep the artifacts?

circle.yml
```diff
@@ -3,22 +3,15 @@ machine:
...
+    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx3400m -XX:+HeapDumpOnOutOfMemoryError"'
 
 dependencies:
   override:
     - ./gradlew resolveConfigurations
 
 test:
   override:
-    - ./gradlew build testBouncyCastle testSun
+    - ./gradlew --daemon --parallel build testBouncyCastle testSun
```
markelliot, line 14: Shouldn't we daemon everywhere if we daemon at all?

circle.yml
uschi2000, line 13: As it stands, the artifact list was incomplete. Tests results and failures are collected automatically as far as I understand. Did you have another use for these?

circle.yml
uschi2000, line 14: Yep, can do.

PR #174, opened 2016-07-21
circle.yml: always use Gradle -- parallel; limit Gradle test memory

uschi2000:


circle.yml
```diff
@@ -3,18 +3,18 @@ machine:
     version: oraclejdk7
   environment:
     TERM: dumb
-    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx3400m -XX:+HeapDumpOnOutOfMemoryError"'
+    GRADLE_OPTS: '-Dorg.gradle.jvmargs="-Xmx3g -XX:+HeapDumpOnOutOfMemoryError"'
```
markelliot, line 6: think we might want to turn this down to 1g

build.gradle
```diff
@@ -51,4 +51,9 @@ subprojects {
     if (System.env.CIRCLE_TEST_REPORTS) {
         test.reports.junitXml.destination = new File(System.env.CIRCLE_TEST_REPORTS, it.getName())
     }
+
+    test {
+        minHeapSize = "128m"
```
markelliot: would set these to be the same, the realloc can cost extra

circle.yml
uschi2000, line 6: gradle doesn't even start up with 2g (at least on my machine) :(

markelliot, line 6: bummer, fair enough

circle.yml
```diff
@@ -3,18 +3,18 @@ machine:
...
 
 dependencies:
   override:
     - ./gradlew resolveConfigurations
 
 test:
   override:
-    - ./gradlew --daemon --parallel build testBouncyCastle testSun
+    - ./gradlew build testBouncyCastle testSun
```
markelliot, line 14: does `testBouncyCastle` still fork a docker container?

uschi2000, line 14: I don't think so.

PR #176, opened 2016-07-21
Rfink/nebula

uschi2000:


markelliot:
Looks fine -- what's the advantage of the lock file over just using the recommender?

uschi2000:
Lock files let's you observe dependency changes over time since it's
checked in. E.g., when you accidentally bump jackson/feign/okhttp etc.

markelliot:
Cool.

PR #183, opened 2016-07-25
Copy Dropwizard Duration class

uschi2000:


http-clients-api/src/main/java/com/palantir/remoting/clients/ClientConfig.java
```diff
@@ -87,13 +86,13 @@ public static ClientConfig fromServiceConfig(ServiceConfiguration serviceConfig)
 
         // timeouts
         if (serviceConfig.connectTimeout().isPresent()) {
-            clientConfig.connectTimeout(serviceConfig.connectTimeout().get().toMilliseconds(), TimeUnit.MILLISECONDS);
+            clientConfig.connectTimeout(serviceConfig.connectTimeout().get());
```
markelliot: so with this change we can probably just make `clientConfig#connectTimeout(Optional<Duration>)` as an additional method to simplify all this code

service-config/src/main/java/com/palantir/config/service/Duration.java
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
markelliot, line 2: I think you need the Apache license header here as well.

http-clients-api/src/main/java/com/palantir/remoting/clients/ClientConfig.java
uschi2000: you have to write this code somewhere, right? not sure what the better place is?

markelliot: Given that we're making a builder I think the builder should own this code?

uschi2000: sure

service-config/src/main/java/com/palantir/config/service/Duration.java
uschi2000, line 2: done

http-clients-api/src/main/java/com/palantir/remoting/clients/ClientConfig.java
uschi2000: done

uschi2000: Turned into @Immutable now.

http-clients-api/src/main/java/com/palantir/remoting/clients/ClientConfig.java
```diff
@@ -18,129 +18,81 @@
...
+    public Integer maxNumRetries() {
+        return MAX_NUM_RETRIES;
     }
 
-    public static ClientConfig empty() {
-        return new ClientConfig();
+    @Value.Check
+    public void check() {
+        Preconditions.checkState(sslSocketFactory().isPresent() == trustManager().isPresent(),
+                "Must set either both sslSocketFactory and TrustManager, or neither");
```
markelliot: should either read: "Must set either both SslSocketFactory and TrustManager, or neither" or "Must set either both sslSocketFactory and trustManager, or neither"

uschi2000: done

markelliot:
👍

PR #186, opened 2016-07-27
Harmonize package names

uschi2000:


http-servers/src/main/java/com/palantir/remoting/server/IllegalArgumentExceptionMapper.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.palantir.remoting.http.server;
+package com.palantir.remoting.server;
```
markelliot: we should probably pluralize to 'servers'

PR #195, opened 2016-07-31
remove jackson 2.4 compat

robert3005:
It was here only for spark compatibility. Spark 2.0 is on jackson 2.6.5 so we don't need special handling.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/195)


uschi2000:
I don't think we can upgrade this before a bunch of services relying on Spark have removed Spark 1.x support.

robert3005:
I will make sure services get migrated to 2.0 ASAP (probably this week). Can we merge this after you cut 1.0?

uschi2000:
I think we need to keep 2.4 Support in here for a little while so that
teams can take updates without the spark blocker.

uschi2000:
any update here, Rob?

robert3005:
We shouldn't need 2.4 support anymore. I believe nothing internally uses it anymore and otherwise don't see good reason to support it.

schlosna:
Ping on this as it would be nice to remove Jackson 2.4 support, though I guess we have to wait until 2.0?

uschi2000:
Yep. Is this hurting anyone as it is?

schlosna:
I don't know of it hurting anyone, but there are compatibility issues with Jackson Afterburner on 2.4 so we don't enable it, see https://github.com/palantir/http-remoting/pull/284/files#diff-01d6c75da22a1a001eb7a8e0e3198989R36

markelliot:
we solved this by providing the missing 2.4 feature we needed as s shim

PR #198, opened 2016-08-06
Add SslSocketFactories#createTrustContext

uschi2000:


uschi2000:
fyi @ellisjoe

http-clients-api/src/main/java/com/palantir/remoting1/clients/ClientConfig.java
```diff
@@ -65,20 +60,12 @@ public Integer maxNumRetries() {
...
-
     public static ClientConfig fromServiceConfig(ServiceConfiguration serviceConfig) {
         ClientConfig.Builder clientConfig = builder();
 
         // ssl
         if (serviceConfig.security().isPresent()) {
-            clientConfig.sslSocketFactory(SslSocketFactories.createSslSocketFactory(serviceConfig.security().get()));
-            clientConfig.trustManager(
-                    (X509TrustManager) SslSocketFactories.createTrustManagers(serviceConfig.security().get())[0]);
+            clientConfig.trustContext(SslSocketFactories.createTrustContext(serviceConfig.security().get()));
```
ellisjoe, line 68: why doesn't client config take an optional here? otherwise:
`serviceConfig.security().transform(clientConfig::trustContext)`

ellisjoe, line 68: oh this is java 7 isn't it...ok so that won't work but still wondering why it doesn't take an optional

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/TrustContext.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.palantir.remoting1.config.ssl;
+
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.X509TrustManager;
+import org.immutables.value.Value;
+
+/** A wrapper for {@link javax.net.ssl.SSLSocketFactory} and {@link javax.net.ssl.X509TrustManager}. */
+@Value.Immutable
+public abstract class TrustContext {
```
ellisjoe, line 25: Don't need the serialization annotations?

ellisjoe:
+1

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/TrustContext.java
uschi2000, line 25: Nope

http-clients-api/src/main/java/com/palantir/remoting1/clients/ClientConfig.java
uschi2000, line 68: don't understand; who should take an optional?

ellisjoe, line 68: as a convenience. it's a thing that's not needed clients will likely always have an optional TrustContext depending on config.

uschi2000, line 68: so you want SslSocketFactories.createTrustContext to take and return an Optional? hmm, not sure.

uschi2000, line 68: you'd really want `::transform` ...

ellisjoe, line 68: I think you're comment on this atlas pr is similar to what I'm talking about now: https://github.com/palantir/atlasdb/pull/650/files/a387f0ad029979fbfa818d9a323cb3eb479aec5b..9921ad880eec217bd90267e11cf35706e1165fd9#diff-eac7e82cc97989b4f7c7451d5b12087dR96

PR #202, opened 2016-08-16
Ensure slf4j message diagnostic context cleanup

schlosna:


schlosna:
@uschi2000 minor fix in prep for more tracing

http-servers/src/main/java/com/palantir/remoting1/servers/TraceIdLoggingFilter.java
```diff
@@ -39,16 +40,32 @@ public void init(FilterConfig filterConfig) throws ServletException {}
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
             throws IOException, ServletException {
+        try (Closeable ignored = populateTraceContext(request)) {
```
uschi2000: but it's not ignored, right?

uschi2000: (in the sense that the try-with-resource construct will clean up the MDC)

uschi2000: maybe add a comment here explaining how this construction works.

schlosna: Went with more explicit try/finally. 

Was previously `ignored` as Idea doesn't warn on unused var in try-with-resources (though it is implicitly used & closed).

PR #205, opened 2016-08-16
Update readme.md: add api contract section, update all usage information for 1.0.0

uschi2000:


readme.md
```diff
@@ -1,54 +1,116 @@
...
+## dropwizard-servers
+Provides Dropwizard/Jersey exception mappers for translating common JAX-RS exceptions to appropriate HTTP error codes. A
+Dropwizard server is configured for http-remoting as follows:
+
+```java
+public class MyServer extends Application<Configuration> {
+    @Override
+    public final void run(Configuration config, final Environment env) throws Exception {
+        env.jersey().register(new MyResource());
+        DropwizardServers.configure(env, config, MyServerc.class.getSimpleName(), false /** do not ship stacktraces */);
```
markelliot: should we change the boolean here to an enum so that comments aren't required?

```
 enum StackTraces {
     DO_NOT_SEND_OVER_WIRE,
     SEND_OVER_WIRE
 }
```

readme.md
```diff
@@ -92,36 +155,59 @@ The `pkcs1-reader-sun` does not include any extra dependencies, but assumes the
...
+
+http-remoting makes the following opinionated customizations to the standard Dropwizard/Feign/Retrofit behavior.
+
+#### Object serialization/deserialization
+
+All parameters and return values of `application/json` endpoints are serialized/deserialized to/from JSON using a
+Jackson `ObjectMapper` with `GuavaModule` and `Jdk7Module`. Servers must not expose parameters or return values that
+cannot be handled by this object mapper.
+
+#### Error propagation
```
markelliot, line 176: I think we should also define the `SerializableError` format as a specification here so that non-Java servers and non-Java clients know how to implement this.

readme.md
```diff
@@ -92,36 +155,59 @@ The `pkcs1-reader-sun` does not include any extra dependencies, but assumes the
...
 
+# API Contract
+
+http-remoting makes the following opinionated customizations to the standard Dropwizard/Feign/Retrofit behavior.
+
+#### Object serialization/deserialization
+
+All parameters and return values of `application/json` endpoints are serialized/deserialized to/from JSON using a
+Jackson `ObjectMapper` with `GuavaModule` and `Jdk7Module`. Servers must not expose parameters or return values that
+cannot be handled by this object mapper.
```
markelliot, line 174: Want to flag that this seems amorphous, but have no suggestions for how to fix.

readme.md
```diff
@@ -92,36 +155,59 @@ The `pkcs1-reader-sun` does not include any extra dependencies, but assumes the
...
+The `DropwizardServers#configure` routine installs exception mappers for `IllegalArgumentException`,
+`NoContentException`, `RuntimeException` and `WebApplicationException`. The exception mapper sets the response media
+type to `application/json` and returns as response body a JSON representation of a `SerializableError` capturing the
+message, exception name, and optionally stacktrace of the exception.
+
+Both JaxRsClient and Retrofit2Client intercept non-successful HTTP responses and throw a `RemoteException` wrapping the
+deserialized server-side `SerializableError`. The error name of the `RemoteException` is defined by the service API and
+clients should switch&dispatch based on the error name.
+
+#### Optional return values
```
markelliot, line 199: I think we should also document the wire format behavior of the Optional handler.

readme.md
```diff
@@ -92,36 +155,59 @@ The `pkcs1-reader-sun` does not include any extra dependencies, but assumes the
...
+
+Clients and servers propagate call trace ids across JVM boundaries according to the
+[Zipkin](https://github.com/openzipkin/zipkin) specification. Tracing is currently implemented with the
+[Brave](https://github.com/openzipkin/brave) library. In particular, clients insert `X-B3-TraceId: <Trace ID>` HTTP
+headers into all requests which get propagated by Jetty servers into subsequent client invocations.
+
+#### Endpoints returning plain strings
 
-## API Contracts
+Endpoints returning plan strings should produce media type `text/plain`. Return type `Optional<String>` is only
```
markelliot: plan -> plain

readme.md
uschi2000: done: https://github.com/palantir/http-remoting/pull/207

readme.md
uschi2000, line 176: done

readme.md
uschi2000, line 199: it's down there (although the sentence wasn't complete ...): 

```
 202 - a non-empty return value, then the HTTP response has error code 200 and the body carries the deserialized `T` object
 203   directly, rather than a deserialized `Optional<T>` object.
```

readme.md
uschi2000: done

uschi2000:
rebasing

readme.md
markelliot, line 199: That's not the behavior when the Optional is within a more complex object, IIRC.

uschi2000, line 199: it is. cf https://github.com/palantir/http-remoting/blob/develop/jaxrs-clients/src/main/java/feign/OptionalAwareDecoder.java#L45

verified by printing the BODY of an `Optional<Map<String, String>>` endpoint which is: `BODY: [{"something":"something"}]` in https://github.com/palantir/http-remoting/blob/develop/jaxrs-clients/src/test/java/com/palantir/remoting1/jaxrs/feignimpl/OptionalAwareDecoderTest.java#L56

markelliot, line 199: I'm talking about behavior of an object like

class Foo { Optional<String> blah }

uschi2000, line 199: Oh, I see. Will double check.

uschi2000, line 199: done here: https://github.com/palantir/http-remoting/pull/208

PR #208, opened 2016-08-17
Clarify serialization of Optional/Nullable objects in readme

uschi2000:


readme.md
```diff
@@ -196,6 +196,31 @@ serialized Java `StackTraceElement` objects indicating the server-side stack tra
...
+ComplexType value = new ComplexType(
+        Optional.of(
+                new ComplexType(
+                        Optional.<ComplexType>absent(),
+                        Optional.<String>absent(),
+        Optional.of("baz"));
+```
+The JSON-serialized representation of this object is:
+```json
+{"nested":{"nested":null,"string":null},"string":"baz"}
```
markelliot, line 221: Would it also be valid to omit the null properties entirely?

uschi2000, line 221: I don't know.

PR #210, opened 2016-08-17
Move SlfLoggingSpanCollector to com.github.kristofa.brave.ext package

uschi2000:
SlfLoggingSpanCollector has a dependency on brave and needs to be relocated together with Brave in order to prevent class conflicts when both jaxrs-clients and dropwizard-server packages are on the classpath.


schlosna:
Its probably worth pushing these upstream to Brave. I created PR https://github.com/openzipkin/brave/pull/226 for both synchronous and async slf4j logger implementations.

uschi2000:
ah, nice

PR #211, opened 2016-08-18
RemoteException#toString print remote stacktrace if available

uschi2000:


uschi2000:
ping @markelliot @pnepywoda

PR #212, opened 2016-08-18
Added MultiServerRetryInterceptor to support retrofit2

mateuszlitwin:


palantirtech:
Thanks for your interest in [palantir/http-remoting](https://github.com/palantir/http-remoting), @divath! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mateuszlitwin:
@ellisjoe @uschi2000 Can you take a look?

markelliot:
I'd like to propose we take a different (and unifying) approach here: instead of writing an interceptor we can write a more generic dynamic proxy that hangs on to a collection of Retrofitted or Feigned interfaces and implement retry at a higher level -- this gives us a single retry implementation to maintain.

mateuszlitwin:
@markelliot I understand this argument and it is a nice idea, but I am not sure if it is really possible to implement.

There are benefits of my approach:
- it works if you want to use just bare OkHttpClient to make requests then this interceptor can be used
- it works for asynchronous/lazy return values e.g. when your Retrofitted interface returns `Call<T>` - simplest implementation of proxy will fail here
- it does not use reflections, which gives more readable stack traces

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,94 @@
...
+        this.randomize = randomize;
+    }
+
+    @Override
+    public final Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        Exception lastException = null;
+
+        List<String> failoverUris = uris;
+        if (randomize) {
```
markelliot: this randomization should happen in the constructor

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
markelliot, line 2: this needs to have the correct open source header

retrofit2-clients/src/test/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptorTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
markelliot, line 2: this needs to have the correct open source header

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+package com.palantir.remoting1.retrofit2;
```
mateuszlitwin, line 17: Question: Is it the right package/subproject? This class and other interceptors from this package do not depend on retrofit2 and can be used with just OkHttpClient

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin, line 2: done

retrofit2-clients/src/test/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptorTest.java
mateuszlitwin, line 2: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class MultiServerRetryInterceptor implements Interceptor {
+    private static final Logger logger = LoggerFactory.getLogger(MultiServerRetryInterceptor.class);
+
+    private final List<String> uris;
+
+    private MultiServerRetryInterceptor(List<String> uris) {
+        this.uris = uris;
```
uschi2000: immutable copy

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+        Exception lastException = null;
+
+        for (String uri : uris) {
+            request = redirectRequest(request, uri);
+            try {
+                return chain.proceed(request);
+            } catch (SocketTimeoutException | UnknownHostException | HttpRetryException
+                    | MalformedURLException | SocketException | UnknownServiceException e) {
+                lastException = e;
+                logger.warn("Failed to send request to " + request.url(), e);
```
uschi2000, line 81: should note that you're retrying another host

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+            } catch (IOException e) {
+                throw e;
+            }
+        }
+        throw new IllegalStateException(
+                "Could not connect to any node. Please check that the URIs are valid and servers are up.",
```
uschi2000: consider logging the servers tried.

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+
+        for (String uri : uris) {
+            request = redirectRequest(request, uri);
+            try {
+                return chain.proceed(request);
+            } catch (SocketTimeoutException | UnknownHostException | HttpRetryException
+                    | MalformedURLException | SocketException | UnknownServiceException e) {
+                lastException = e;
+                logger.warn("Failed to send request to " + request.url(), e);
+            } catch (IOException e) {
```
uschi2000: is this block redundant?

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+        for (String uri : uris) {
+            if (currentUrl.startsWith(uri)) {
+                matchingUri = uri;
+                break;
+            }
+        }
+
+        if (matchingUri == null) {
+            throw new IllegalStateException(
+                    String.format("None of the URIs %s matched request URL %s", uris, currentUrl));
```
uschi2000: you're using URI elsewhere and URL here

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+                throw e;
+            }
+        }
+        throw new IllegalStateException(
+                "Could not connect to any node. Please check that the URIs are valid and servers are up.",
+                lastException);
+    }
+
+    private Request redirectRequest(Request request, String redirectToUri) {
+        final String currentUrl = request.url().toString();
```
uschi2000: no need for final, IMO

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+                break;
+            }
+        }
+
+        if (matchingUri == null) {
+            throw new IllegalStateException(
+                    String.format("None of the URIs %s matched request URL %s", uris, currentUrl));
+        }
+
+        final String newUrl = currentUrl.replaceFirst(matchingUri, redirectToUri);
```
uschi2000: ditto

uschi2000: this interprets matchingUri as a regular expression. while that may accidentally work, it seems a little strange.

uschi2000: just for my understanding: if `uris[0]` is responsive, then this replaces the uri with itself in every call, correct?

mateuszlitwin: I like to make constants `final`. I find it easier to distinguish them from variables like `uri` or `matchingUri`.

It always (for every request) tries sending request first to uris[0], then uris[1], ... We do not remember that uris[0] did not work for next request we make (which kind of sucks, maybe making it cyclic - starting from last working URI would be the best idea, but then this interceptor will not longer be stateless and threadsafe).

Currently we find which URI from `uris` is a prefix of the current request and then it should be safe to replace that prefix with other valid URI (as long as base URL used in retrofit was one of the `uris`).
I don't have a better idea for changing request path, since potentially different services might be using different ports, hosts, api versions and application context paths.

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -0,0 +1,115 @@
...
+                matchingUri = uri;
+                break;
+            }
+        }
+
+        if (matchingUri == null) {
+            throw new IllegalStateException(String.format("Unrecognized server URI in the request %s. "
+                            + "Supported servers are %s. Did you use different server URI for the initial request?",
+                    requestUri, uris));
+        }
```
ellisjoe, line 104: Seems like from here to the top of this method should only be done once?

mateuszlitwin: yep, I can use last uri without a search, but it does not really matter I think and I don't want to complicate this method

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
uschi2000: Yep, I think that approach is fine. `final` is not used in that way in this code-base, so I would prefer consistency here, please remove.

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
uschi2000: agree with @divath

uschi2000:
you you respond to the open comments, please, @divath ?

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptor.java
mateuszlitwin: done

mateuszlitwin:
I think I answered all questions

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -101,6 +98,7 @@ public Request authenticate(Route route, Response response) throws IOException {
             client.addInterceptor(new RetryInterceptor(config.maxNumRetries()));
         }
 
+        client.addInterceptor(MultiServerRetryInterceptor.create(uris));
```
uschi2000, line 101: not covered by a test

retrofit2-clients/src/test/java/com/palantir/remoting1/retrofit2/MultiServerRetryInterceptorTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+
+public final class MultiServerRetryInterceptorTest {
+    @Rule
+    public final MockWebServer unavailableServer = new MockWebServer();
+    @Rule
+    public final MockWebServer availableServer = new MockWebServer();
+
+    @Test
+    public void interceptsRequestAndRedirectsToAvailableServer() throws IOException, InterruptedException {
```
uschi2000: would like to see test for:
- no server available: throws expected exception
- first server available: second server is not hit

uschi2000:
cool, code looks good. three tests are missing, see above.

uschi2000:
thanks, Mateusz!

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
mateuszlitwin: check testRetrofit2ClientWithMultiServerRetryInterceptorRedirectToAvailableServer()

uschi2000:
👍

mateuszlitwin:
Glad I could help, we needed it anyway for other product

ash211:
Thanks for the contribution @divath !

PR #215, opened 2016-08-23
SerializeError#getStackTrace returns stacktrace as list of SerializableStackTraceElement

uschi2000:


error-handling/src/main/java/com/palantir/remoting1/errors/SerializableError.java
```diff
@@ -49,7 +54,7 @@
 
     /** An optional representation of the server-side stacktrace of this error. */
     @Nullable
-    public abstract List<StackTraceElement> getStackTrace();
+    public abstract List<String> getStackTrace();
```
markelliot: is this a wire-break?

uschi2000: yes.

markelliot: that's problematic, though?

maybe we should make `List<StackTraceElement>` totally optional and support also a `getCallStack` or something?

pnepywoda:
Prefer keeping things as-is (yes it's a bit java-centric, but so are most of our services so optimize on that) and adding `Optional<String> getFormattedError`

uschi2000:
Formatted as what?

pnepywoda:
Formatted as nothing. Whatever the thing making the exception wants.

markelliot:
I think it's probably useful to continue to provide a richer data structure that represents a callstack. Maybe we should turn it into a List<Map<String, Object>> or a List<Map<String, String>>? (I think either of these would also preserve back-compat, actually)

uschi2000:
Java's StacktraceElement has: methodName, fileName, lineNumber, className, nativeMethod. They seem to be pretty generic. I could provide a GenericStackTrace class in which all of these are Optional<>.

uschi2000:
plus an Optional `language` field to indicate whether this is java vs python vs foo.

markelliot:
Does Python have additional stuff that would be useful?

pnepywoda:
What advantage do we get from forcing things to be structured? The
front-ends that are displaying these are NOT going to parse the errors and
just want to display exactly what a user would be seeing from the
compiler/interpreter anyway. Just make it a string!

markelliot:
Same question for Go.

nmiyake:
I believe the things included for the Java exceptions cover everything for go as well (can't think of any extra information that the stack traces/exceptions have there)

uschi2000:
@pnepywoda , because the renderer may have different opinions on how to render it. showing it in html requires `</br>`, showing it in markdown requires indentation, showing it on linux needs `\n`, on windows `\r\n`, etc. it's just not the place for formatting, IMO.

pnepywoda:
Ah right, I see, well...maybe List<String> instead of
List<StackTraceElement> is sufficient...

uschi2000:
`List<String>` would be a backcompat (wire) break, unfortunately.

uschi2000:
How about this? Same wire format as before (see test).

uschi2000:
@pnepywoda

uschi2000:
@justinuang

error-handling/src/main/java/com/palantir/remoting1/errors/SerializableStackTraceElement.java
```diff
@@ -0,0 +1,47 @@
...
+import org.immutables.value.Value;
+
+@JsonDeserialize(as = ImmutableSerializableStackTraceElement.class)
+@JsonSerialize(as = ImmutableSerializableStackTraceElement.class)
+@Value.Immutable
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
+public abstract class SerializableStackTraceElement {
+
+    @Nullable
+    public abstract String getClassName();
```
markelliot: Maybe call these Optional instead of Nullable?

markelliot:
Happy with this besides nit, thanks.

error-handling/src/main/java/com/palantir/remoting1/errors/SerializableStackTraceElement.java
uschi2000: the objectmapper doesn't currently use Guava. it could, I guess

uschi2000: changing to optional

markelliot: ugh, fine either way I guess. Will be nice once we can just assume Java 8.

pnepywoda:
I defer to @justinuang since this change most affects him, but lgtm

markelliot:
👍

PR #216, opened 2016-08-25
Allow charset for text/plain Content-Type

splittingfield:
For many HTTP responses from various places, the Content-Type includes the charset, i.e.
`Content-Type: text/plain; charset=utf-8`. According to the
Content-Type spec, this is perfectly acceptable. The TextDelegateDecoder
was checking for exact string equality, which resulted in attempting
to use the Json delegate Decoder, as opposed to the String decoder in
this situation.

This commit:

1: Added a failing test to demonstrate the issue,
2: Fixes the failing test, but changing `.equals()` to `.startsWith()` (after subsequent conversation)

### DIFF
--- a/jaxrs-clients/src/main/java/feign/TextDelegateDecoder.java
+++ b/jaxrs-clients/src/main/java/feign/TextDelegateDecoder.java
@@ -48,7 +48,7 @@ public final class TextDelegateDecoder implements Decoder {
             contentTypes = ImmutableSet.of();
         }
         // In the case of multiple content types, or an unknown content type, we'll use the delegate instead.
-        if (contentTypes.size() == 1 && Iterables.getOnlyElement(contentTypes, "").equals(MediaType.TEXT_PLAIN)) {
+        if (contentTypes.size() == 1 && Iterables.getOnlyElement(contentTypes, "").contains(MediaType.TEXT_PLAIN)) {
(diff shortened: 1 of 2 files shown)


jaxrs-clients/src/main/java/feign/TextDelegateDecoder.java
```diff
@@ -48,7 +48,7 @@ public Object decode(Response response, Type type) throws IOException, FeignExce
             contentTypes = ImmutableSet.of();
         }
         // In the case of multiple content types, or an unknown content type, we'll use the delegate instead.
-        if (contentTypes.size() == 1 && Iterables.getOnlyElement(contentTypes, "").equals(MediaType.TEXT_PLAIN)) {
+        if (contentTypes.size() == 1 && Iterables.getOnlyElement(contentTypes, "").contains(MediaType.TEXT_PLAIN)) {
```
markelliot: should this be `startsWith`?

splittingfield: Let me go re-read the spec.  I originally had that, but then changed to `contains`.

splittingfield: Yes, updating PR now.

uschi2000:
👍

PR #227, opened 2016-09-08
SlfLoggingSpanCollector logs at TRACE level instead of INFO, fixes #225

### LINKED ISSUE #225: trace logging should use TRACE log level

uschi2000:


uschi2000:
fyi @schlosna

schlosna:
Rob, I'm OK with this once builds pass. Out of curiosity, what's the main motivation behind this? Is it just so that this isn't logged by default with dropwizard default INFO level?

uschi2000:
exactly.

PR #229, opened 2016-09-12
Upgrade to Brave 3.11.0

schlosna:
Bump Brave version to 3.11.0 which supports inheritable Brave tracing.

Note that we still need to make the Brave tracer accessible for application local tracing use.

See https://github.com/openzipkin/brave/issues/206 for one option of global `BraveRegistry` though we could expose through Jersey DI.

### DIFF
--- a/dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardTracingFilters.java
+++ b/dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardTracingFilters.java
@@ -16,6 +16,7 @@
 
 package com.palantir.remoting1.servers;
 
+import com.github.kristofa.brave.AnnotationSubmitter;
 import com.github.kristofa.brave.Sampler;
 import com.github.kristofa.brave.ServerRequestInterceptor;
 import com.github.kristofa.brave.ServerResponseInterceptor;
@@ -39,6 +40,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.EnumSet;
 import java.util.Random;
+import java.util.concurrent.TimeUnit;
 import javax.servlet.DispatcherType;
 
 /** Static utilities for registering Brave/Zipkin filters with Dropwizard applications. */
@@ -74,6 +76,12 @@ final class DropwizardTracingFilters {
                 .randomGenerator(new Random())
                 .state(new ThreadLocalServerClientAndLocalSpanState(ip, port, name))
                 .spanCollector(new SlfLoggingSpanCollector("tracing.server." + name))
+                .clock(new AnnotationSubmitter.Clock() {
+                    @Override
+                    public long currentTimeMicroseconds() {
+                        return TimeUnit.MICROSECONDS.convert(System.nanoTime(), TimeUnit.NANOSECONDS);
+                    }
+                })
                 .build();
     }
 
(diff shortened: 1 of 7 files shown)


schlosna:
@uschi2000 for SA

dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardTracingFilters.java
```diff
@@ -54,25 +54,26 @@ private DropwizardTracingFilters() {}
      * <p>
      * TODO(rfink) Is there a more stable way to retrieve IP/Port information?
      */
-    static void registerTracers(Environment environment, Configuration config, String tracerName) {
-        ServerTracer serverTracer = getServerTracer(extractIp(config), extractPort(config), tracerName);
+    static Brave registerTracers(Environment environment, Configuration config, String tracerName) {
```
markelliot: why do you need to return `Brave`?

dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardServers.java
```diff
@@ -48,7 +49,8 @@ public static void configure(
             Configuration config,
             String tracerName,
             Stacktraces stacktracePropagation) {
-        DropwizardTracingFilters.registerTracers(environment, config, tracerName);
+        Brave brave = DropwizardTracingFilters.registerTracers(environment, config, tracerName);
+        environment.jersey().register(brave);
```
uschi2000: wow, what does this mean?

uschi2000: that's seems very very magical

dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardTracingFilters.java
schlosna: We need a way to expose the brave instance for use by local tracing

markelliot: Do you have to plumb `Brave` through all the methods you want to trace locally? That seems to defeat the purpose of this setup. I'd be expecting a global static accessor.

schlosna: I think we can add a global static accessor method here to `DropwizardServers`, thoughts?

markelliot: Eh, this kind of stuff makes me wonder if Brave is actually the right library for us.

I guess we can add our own Tracing object with static methods to add and complete traces that delegate to a Brave instance that DropwizardServers sets up. This gets us pretty close to just doing the impl ourselves, though.

schlosna: I think we can add a simple Tracer interface with implementation backed by Brave for now. I should have Tritium ready shortly so that we can inject the tracer in there and wrap services as needed.

schlosna:
@uschi2000 @markelliot I reverted the parts exposing the Brave tracer for now just to get the Brave 3.11 upgrade in place first, and I'll create another PR once this merges with something along the lines of https://github.com/schlosna/http-remoting/compare/davids/brave-3.11...schlosna:davids/brave-tracer

uschi2000:
cheers

PR #235, opened 2016-09-21
Initialize Brave Zipkin tracing for Dropwizard servers and JAX RS clients

schlosna:
Introduces a Tracer wrapping Brave to produce Zipkin traces and spans to the "tracing" slf4j logger.

Typically a Dropwizard server should call `com.palantir.remoting1.servers.DropwizardServers#configure` as part of startup initialization, which will take care of registering the Brave singleton tracer in `com.palantir.remoting1.servers.Tracers#activeTracer` as a Jersey server interceptor.

JAX RS clients created via `com.palantir.remoting1.jaxrs.JaxRsClient#builder(com.palantir.remoting1.clients.ClientConfig)` will inherit the active tracer by default, or can provide their own via `ClientConfig`.

Similarly the Tracer allows for initiating local traces with its API.

Addresses #115


schlosna:
@uschi2000 @markelliot @clockfort for SA

schlosna:
We still need to fix the span collection plumbing, sampling configuration, and a few more things, but I'm considering those for a followup PR for now.

uschi2000:
before I start reading the code changes: what's the high-level goal of this change?

schlosna:
Rob, the high level goal is to provide a single unified Brave instance to Dropwizard based applications so that they can do local tracing, and inherit the proper incoming server trace & span as top level local spans' parents.

uschi2000:
Got it, thanks

schlosna:
This will require some tweaking as http-remoting is shading some Brave components which causes the following issue when using this:

```
Exception in thread "main" java.lang.NoSuchMethodError: com.palantir.remoting1.servers.BraveTracer.defaultClock()Lcom/github/kristofa/brave/AnnotationSubmitter$Clock;
  at com.palantir.remoting1.servers.DropwizardTracingFilters.createBraveTracer(DropwizardTracingFilters.java:99)
  at com.palantir.remoting1.servers.DropwizardTracingFilters.getOrCreateBraveTracer(DropwizardTracingFilters.java:82)
  at com.palantir.remoting1.servers.DropwizardTracingFilters.registerTracers(DropwizardTracingFilters.java:63)
  at com.palantir.remoting1.servers.DropwizardServers.configure(DropwizardServers.java:50)
```

Might also be worth separating the tracing initialization from the other aspects in `DropwizardServers` as the tracing needs to be setup first before anything creates JAX RS clients or nothing will be traced.

dropwizard-servers/src/main/java/com/palantir/remoting1/servers/DropwizardTracingFilters.java
```diff
@@ -57,32 +60,45 @@ private DropwizardTracingFilters() {}
      * TODO(rfink) Is there a more stable way to retrieve IP/Port information?
      */
     static void registerTracers(Environment environment, Configuration config, String tracerName) {
-        ServerTracer serverTracer = getServerTracer(extractIp(config), extractPort(config), tracerName);
+        final BraveTracer tracer = getOrCreateBraveTracer(config, tracerName);
```
uschi2000, line 63: final isn't normally used in this codebase

ext/brave-extensions/src/main/java/com/palantir/remoting1/servers/BraveContext.java
```diff
@@ -0,0 +1,51 @@
...
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.SpanId;
+import com.google.common.base.Preconditions;
+
+public final class BraveContext implements Tracer.TraceContext {
+
+    private final Brave brave;
+    private final SpanId spanId;
+
+    @SuppressWarnings("WeakerAccess") // public API
```
uschi2000, line 28: this is to suppress the intelliJ hint?

ext/brave-extensions/src/test/java/com/palantir/remoting1/servers/BraveTracerTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        verify(mockRunnable, times(1)).run();
+        verify(mockSpanCollector, times(1)).collect(any(Span.class));
+        verifyNoMoreInteractions(mockSpanCollector, mockRunnable);
+    }
+
+    @Test
+    public void traceCallable() throws Exception {
+        @SuppressWarnings("unchecked") // yay generic type erasure
+        Callable<String> mockCallable = mock(Callable.class);
```
uschi2000, line 93: can get around this with annotation-based mocking:

```
@Mock
private Foo<Bar> foo;
```

http-clients-api/src/main/java/com/palantir/remoting1/servers/Tracer.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package com.palantir.remoting1.servers;
+
+public interface Tracer {
+
+    /**
+     * Begins a trace context for the specified component and operation.
+     *
+     * @param component component name
```
uschi2000, line 24: remove

http-clients-api/src/main/java/com/palantir/remoting1/servers/Tracer.java
```diff
@@ -0,0 +1,41 @@
...
+     * @param operation operation name
+     *
+     * @return new trace context
```
uschi2000, line 27: remove

http-clients-api/src/main/java/com/palantir/remoting1/servers/Tracers.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.common.base.Preconditions;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicReference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utility methods for tracing components and operations.
+ */
+@SuppressWarnings("WeakerAccess")
```
uschi2000, line 28: not a fan of this

uschi2000:
Hmm, I find this all very hard to understand :( There's a strange mixture between static state (e.g., `Tracers.getActiveTracer()`) and injected state (e.g., `ClientConfig#setTracer`). David, can we brainstorm a little in person to find a simpler solution?

(Or maybe I'm just not getting the point...)

http-clients-api/src/main/java/com/palantir/remoting1/servers/Tracer.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    TraceContext begin(String component, String operation);
+
+    Class<? extends TraceContext> getContextClass();
```
uschi2000, line 31: why do we need this?

uschi2000 (commented):
Still struggling with the semi-coupled nature of our Tracers wrapper and the Brave implementation. Is there a particular reason we need to wrap Brave?

ext/brave-extensions/src/main/java/com/palantir/remoting1/servers/BraveContext.java
schlosna: Yes, this suppresses an IntelliJ warning since this is unused in this project

schlosna:
I'd be fine with exposing Brave directly to both servers (and possibly clients) for use as needed for local tracing -- my main goal here is to get the inheritable state shared across server, client, and local tracers, as well as have a consistent mechanism for configuring the Brave instance to sample & collect spans, so that we can have the configuration injected via service discovery.

The main concern for exposing Brave here is that we're tying ourselves to a concrete API. I don't have strong objections to that as I think we can handle breaks and incompatibility as needed for our own http-remoting consumers.

Another option might be to use the `opentracing-api` from 
https://github.com/opentracing/opentracing-java and the `brave-opentracing` bridge from https://github.com/openzipkin/brave-opentracing . That route might lose some of the access to Brave niceties, but I haven't looked into that route quite yet.

uschi2000:
The current PR feels like a limbo with some parts wrapped in our own API
and some parts using Brave directly.

uschi2000:
(Sorry, I have to run now.)

markelliot:
I think the challenge with this PR is that it's not a full-measure for wrapping Brave. My basic stance (and I don't feel strongly about which of these to choose) is that we need to either:
1. Just use Brave as-is and live with its semantics
2. Use our own tracing shim (which is less code than this PR and will have precisely the semantics we want) + write an adapter from that shim to integrate with Zipkin

If I had to choose, I'd prefer 2.

schlosna:
@markelliot I don't really want to wrap Brave, I'd rather inject it into all of the places that we want to do tracing, but a lot of the http-remoting APIs currently prevent that and make this more difficult than it needs to be -- e.g. we need to push the trace config into the JaxRs ClientConfig so that tracing there knows what to trace and where to push it. This is made more difficult by the shading done by http-remoting and dependency limitations, which is how I ended up with this PR for the time being.

markelliot:
Not sure I grok what you're saying – we own all of the JaxRs code, so we could in theory make this simpler?

schlosna:
I have a cleanup that I think is cleaner -- it will add brace-core and brace-okhttp dependency for http-clients-api.
I will push later tonight or tomorrow once I increase test coverage.

uschi2000:
Cheers David

schlosna:
See also #115

PR #236, opened 2016-09-22
Remove classifier from published jar

ellisjoe:
without this the jar gets published to `jaxrs-clients-all-1.0.0-rc4-all.jar`. the extra `-all` directly before `.jar` is incorrect


ellisjoe:
@uschi2000 tested this locally and we're now pulling in the jaxrs clients jar correctly. without this change we would get the pom file correctly but be missing the jar. @nmiyake for SA

uschi2000 (approved):

PR #238, opened 2016-09-24
Retrofit2ClientBuilder can handle URLs without trailing slash

uschi2000:


uschi2000:
fyi @divath

uschi2000:
The retry interceptor has a problem with endpoints with absolute paths, see test at https://github.com/palantir/http-remoting/pull/238/files#diff-10f52a06b08a46126219d21c3d52de75R68

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -50,15 +52,27 @@ public Retrofit2ClientBuilder(ClientConfig config) {
...
         Retrofit retrofit = new Retrofit.Builder()
                 .client(client)
-                .baseUrl(uris.get(0))
+                .baseUrl(sanitizedUris.get(0))
                 .addConverterFactory(JacksonConverterFactory.create())
                 .build();
         return retrofit.create(serviceClass);
     }
 
+    private List<String> sanitizeUris(List<String> uris) {
```
mateuszlitwin: missing static?

retrofit2-clients/src/test/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilderTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertThat(server.takeRequest().getPath(), is(String.format(expectedQueryPath, "")));
+
+        server.enqueue(new MockResponse().setBody("\"server\""));
+        assertThat(service.getRelative().execute().body(), is("server"));
+        assertThat(server.takeRequest().getPath(), is(String.format(expectedQueryPath, "relative")));
+    }
+
+    // See comment regarding Retrofit2 URL handling on TestService#getAbsolute()
+    @Test
+    public void testAbsoluteRetrofitEndpoints_failRetryInterceptor() throws Exception {
```
mateuszlitwin, line 60: `throws Exception` is unnecessary I think or at least can use smaller scope, same in other places

mateuszlitwin: yes, good idea to check both relative and absolute paths

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -50,15 +52,27 @@ public Retrofit2ClientBuilder(ClientConfig config) {
...
+        return Lists.transform(uris, new Function<String, String>() {
+            @Override
+            public String apply(String input) {
+                return input.substring(input.length() - 1).equals("/")
```
mateuszlitwin: nit: charAt is probably better choice, but does not really matter

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -50,15 +52,27 @@ public Retrofit2ClientBuilder(ClientConfig config) {
     @Override
     public <T> T build(Class<T> serviceClass, String userAgent, List<String> uris) {
```
mateuszlitwin, line 53: This will change default behavior that you would expect if a service was created with just Retrofit, which kind of sucks. **On the other hand Retrofit would not accept URL and just  throw some exception, right?** In such case it is probably not a big deal to accept more URLs.

mateuszlitwin (commented):
LGTM added few nit picks and one question

mateuszlitwin:
Absolute paths are hard to solve in interceptor level and in general can mess up a lot of things.

If you have 3 services:

http://aaa/serviceName
http://bbb/otherServiceName
http://ccc/serviceName

Then URLs for endpoint with relative path should be

http://aaa/serviceName/relative
http://bbb/otherServiceName/relative
http://ccc/serviceName/relative

And for endpoint with absolute path should be

http://aaa/absolute
http://bbb/absolute
http://ccc/absolute

I think we can't with 100% confidence check if intercepted request was created from absolute or relative path. Which means we don't know whether to just replace host or both host and base path (base paths are not necessarily the same for different hosts).
Even if we check if intercepted URL is using base path we don't know whether base path was part of the absolute path (if it was it should not be replaced with base path of other node).

---

Side note: I think also nested services might be broken e.g.

http://aaa/api/           (e.g. used for http://aaa/api/v0/endpoint)
http://aaa/api/v1/      

---

Unfortunately I don't know if there are better solutions. Reflection based retryier is probably hard to implement and might not work as expected in custom configurations (e.g. interface with endpoints returning some custom types like Observable).

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
uschi2000: Retrofit2 throws if you don't give it /-terminated base URLs. So there's no ambiguity in behaviour.

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
uschi2000: fixed

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
uschi2000: done

retrofit2-clients/src/test/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilderTest.java
uschi2000: re exceptions: i'm happy for all tests to throw exceptions and prefer to avoid syntax overhead.

uschi2000:
OK, the absolutely endpoint path thing is unresolved. ideas, anyone?

retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -50,15 +52,27 @@ public Retrofit2ClientBuilder(ClientConfig config) {
...
         Retrofit retrofit = new Retrofit.Builder()
                 .client(client)
-                .baseUrl(uris.get(0))
+                .baseUrl(sanitizedUris.get(0))
                 .addConverterFactory(JacksonConverterFactory.create())
                 .build();
         return retrofit.create(serviceClass);
     }
 
+    private static List<String> sanitizeUris(List<String> uris) {
```
cihati: I'd rather call this addTrailingSlash, as it doesn't do any other sanitization. Also, you wouldn't need a comment above to clarify as `// adds trailing slash`

uschi2000: done

uschi2000:
anything left here?

mateuszlitwin (approved):

mateuszlitwin:
Fine with me; I am wondering why would someone use absolute path?

uschi2000:
I don't know why. But people might :)

uschi2000:
ping @markelliot . thoughts?

markelliot (approved):

uschi2000:
rebasing

PR #242, opened 2016-09-30
JaxRsClient does not support retries

uschi2000:


jaxrs-clients/src/main/java/com/palantir/remoting1/jaxrs/FeignJaxRsClientBuilder.java
```diff
@@ -80,11 +81,12 @@
     FeignJaxRsClientBuilder(ClientConfig config, BackoffStrategy backoffStrategy) {
         this.backoffStrategy = backoffStrategy;
         this.config = config;
+        Preconditions.checkArgument(config.maxNumRetries() == 0,
```
nmiyake, line 79: If we're making this change, is there any reason to even have `backoffStrategy` as an ivar? Seems like we can just define it as a static final constant with the value `NeverRetryingBackoffStrategy.INSTANCE`.

uschi2000:
ping @nmiyake good to merge?

nmiyake (approved):

nmiyake:
Sorry for the delay -- these changes look good to me. Thanks!

PR #244, opened 2016-10-04
Brave tracer v2

schlosna:


http-clients-api/src/main/java/com/palantir/remoting1/clients/ClientBuilder.java
```diff
@@ -16,12 +16,15 @@
...
 package com.palantir.remoting1.clients;
 
+import com.github.kristofa.brave.Brave;
 import java.util.Arrays;
 import java.util.List;
 
 /** Abstraction for creating HTTP-invoking dynamic proxies around service interfaces. */
 public abstract class ClientBuilder {
 
+    public abstract ClientBuilder withTracer(Brave brave);
```
markelliot, line 26: I think we want this to be automatic, not to require plumbing, which speaks to my comment a bit on why I think Brave may not be the right framework for us.

schlosna:
See also #115

markelliot:
going to close this for now, think @uschi2000's branch has us covered

PR #245, opened 2016-10-06
Rfink/revert publish non shaded jaxrs clients

uschi2000:


uschi2000:
Non-shaded version is no longer used by @nmiyake . @j-baker had a good point about the increased danger of diamond dependencies if publish both the shaded and the non-shaded variant. fyi @ellisjoe

markelliot (approved):

PR #246, opened 2016-10-06
Do not include Guava in jaxrs-clients shadow jar

uschi2000:


markelliot:
what's impetus for having to shade Guava?

uschi2000:
before this PR: guava is not shaded, but included in the jar
after this PR: guava is not shaded, and _not_ included in the jar; instead, there's a Maven dependency on guava.

markelliot (approved):

PR #248, opened 2016-10-10
pass-through SerializableError for RemoteExceptions

pnepywoda:


pnepywoda:
@uschi2000 this one actually also passes through the original exception

uschi2000 (approved):

PR #249, opened 2016-10-11
Add ability to create a TrustContext from PEM/DER certs in memory

jared2501:


ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
```diff
@@ -80,12 +83,43 @@ private KeyStores() {
...
+    static KeyStore createTrustStoreFromCertificates(Map<String, byte[]> certificatesByAlias) {
         try {
             KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
             keyStore.load(null, null);
 
-            for (File currFile : getFilesForPath(path)) {
-                keyStore.setCertificateEntry(currFile.getName(), readX509Certificate(currFile.toPath()));
+            for (Map.Entry<String, byte[]> entry : certificatesByAlias.entrySet()) {
+                keyStore.setCertificateEntry(
+                        entry.getKey(), readX509Certificate(new ByteArrayInputStream(entry.getValue())));
```
markelliot: you don't close the BAIS

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
```diff
@@ -80,12 +83,43 @@ private KeyStores() {
...
+    /**
+     * Returns a {@link KeyStore} created by loading the given certificates.
+     *
+     * @param certificatesByAlias
+     *        a map of X.509 certificate in PEM or DER format by the alias to load the certificate as.
+     */
+    static KeyStore createTrustStoreFromCertificates(Map<String, byte[]> certificatesByAlias) {
         try {
             KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
             keyStore.load(null, null);
```
markelliot: these two lines should mirror the function above, which catch the exception near here and provide more granular per-certificate exception handling below

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
```diff
@@ -80,12 +83,43 @@ private KeyStores() {
...
+        return keyStore;
+    }
+
+    /**
+     * Returns a {@link KeyStore} created by loading the given certificates.
+     *
+     * @param certificatesByAlias
+     *        a map of X.509 certificate in PEM or DER format by the alias to load the certificate as.
+     */
+    static KeyStore createTrustStoreFromCertificates(Map<String, byte[]> certificatesByAlias) {
```
markelliot: I think `byte[]` is probably a weird way to carry the certificate bytes, maybe we should make an alias object so it's obvious what it is?

markelliot (changes requested):
not a huge fan of the code duplication, but probably needs a slightly bigger refactor to eliminate it, and probably isn't worth the refactor. (ie could have a `CertificateCreationStrategy<T>` and then create fixed instances of it to pass to the other methods, meh)

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: Sweet - how does `PemX509Certificate` sound as alias?

markelliot:
Works for me.

uschi2000:
wow, this github diff is quite terrible :(

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/SslSocketFactories.java
```diff
@@ -46,6 +49,11 @@ public static SSLSocketFactory createSslSocketFactory(SslConfiguration config) {
         return sslContext.getSocketFactory();
     }
 
+    public static SSLSocketFactory createSslSocketFactory(Map<String, byte[]> trustCertificatesByAlias) {
```
uschi2000: needs docs

uschi2000: you'll have to explain somewhere what exactly you mean by "trustCertificatesByAlias"

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: technically, the close method on the BAIS does nothing:

```
public void close() throws IOException {
    }
```

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: done

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: done

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: done

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: done

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/SslSocketFactories.java
jared2501: done

jared2501:
addressed comments

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
```diff
@@ -80,18 +82,57 @@ private KeyStores() {
...
      */
     static KeyStore createTrustStoreFromCertificates(Path path) {
+        KeyStore keyStore;
         try {
-            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
+            keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
             keyStore.load(null, null);
+        } catch (IOException | GeneralSecurityException e) {
+            throw Throwables.propagate(e);
+        }
```
markelliot: pull out this block of code into `private static KeyStore createKeyStore()`

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/SslSocketFactories.java
```diff
@@ -74,6 +88,24 @@ public static SSLContext createSslContext(SslConfiguration config) {
...
+    public static SSLContext createSslContext(Map<String, PemX509Certificate> trustCertificatesByAlias) {
+        TrustManager[] trustManagers = createTrustManagers(trustCertificatesByAlias);
+
+        try {
+            SSLContext sslContext = SSLContext.getInstance("TLS");
+            sslContext.init(new KeyManager[]{}, trustManagers, null);
+            return sslContext;
+        } catch (GeneralSecurityException e) {
+            throw Throwables.propagate(e);
+        }
```
markelliot, line 102: guessing this block of code also repeated: `private static SSLContext createSslContextFromTrustManagers(TrustManager[] trustManagers)`?

markelliot (changes requested):

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/KeyStores.java
jared2501: done

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/SslSocketFactories.java
jared2501: done

jared2501 (commented):
done review

markelliot (commented):
lgtm, Rob?

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/PemX509Certificate.java
```diff
@@ -0,0 +1,36 @@
...
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
+@JsonSerialize(as = ImmutablePemX509Certificate.class)
+@JsonDeserialize(as = ImmutablePemX509Certificate.class)
+public abstract class PemX509Certificate {
+
+    /** A X509.1 certificate in PEM format encoded in UTF-8. */
+    public abstract String pemCertificate();
```
uschi2000, line 34: what's the deal here with those PEM headers, something like `---PUBLIC KEY---`?

jared2501: they can be anything - `PUBLIC KEY`, `CERTIFICATE`, etc.. I don't think it actually matters. Can write test if you wish?

uschi2000:
just add a comment, maybe?

uschi2000:
or a test too if it's simple enough.

ssl-config/src/main/java/com/palantir/remoting1/config/ssl/PemX509Certificate.java
nmiyake: Haven't verified behavior myself, but sounds like content after `BEGIN` and `END` may matter -- see https://tools.ietf.org/html/rfc7468#page-11 and http://stackoverflow.com/a/5356351. May be worth verifying/investigating.

jared2501: done - added this comment:

```
    /**
     * A X509.1 certificate or certificate chain in PEM format encoded in UTF-8.
     * <p>
     * The certificates must be delimited by the the begin and end {@code CERTIFICATE} markers.
     */

```

uschi2000 (approved):

uschi2000:
@markelliot I think you'll have to "approve" explicitly

markelliot (approved):

PR #251, opened 2016-10-12
Separate Dropwizard and Jersey functionality, add jersey-servers project

uschi2000:


markelliot:
`ExceptionMappers` is package private, but we need it to be public. `jersey-servers` as a project should also have a different package space than `dropwizard-servers`

uschi2000:
Done

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/ExceptionMappers.java
```diff
@@ -14,20 +14,20 @@
...
     /** Java7-compatible version of Java8 Consumer. */
-    interface Consumer<T> {
+    public interface Consumer<T> {
         void accept(T object);
     }
 
     private ExceptionMappers() {}
 
-    static void visitExceptionMappers(boolean includeStackTrace,
+    public static void visitExceptionMappers(boolean includeStackTrace,
```
markelliot: one thing we might want to think about is if this thing should just return an array of `ExceptionMappers` rather than do this visiting thing.

markelliot: don't need to change that in this PR

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/NoContentExceptionMapper.java
```diff
@@ -14,15 +14,15 @@
...
+package com.palantir.remoting1.servers.jersey;
 
 import javax.ws.rs.core.NoContentException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 
 @Provider
-final class NoContentExceptionMapper implements ExceptionMapper<NoContentException> {
+public final class NoContentExceptionMapper implements ExceptionMapper<NoContentException> {
```
markelliot: why does this, among the others, need to be public?

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/ExceptionMappers.java
markelliot: worth moving the Stacktrace enum thing down to this level, though?

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/NoContentExceptionMapper.java
uschi2000: it's used in the dropwizards tests. replaced by DropwizardServers.configure, revert public change.

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/ExceptionMappers.java
uschi2000: done

markelliot (approved):

PR #253, opened 2016-10-12
Remove Brave tracing

uschi2000:


uschi2000:
fyi @markelliot : first step is to remove current brave tracing. will work against feature branch rfink/revamp-tracing and eventually merge that into develop.

SerialVelocity:
@uschi2000 What was the reasoning behind removing brave tracing and replacing it with a custom version?

markelliot:
Brave didn't have support for variable shape traces, where within a server, local traces could be used to more granularly instrument code.

Plenty of discussion here:
https://github.com/palantir/http-remoting/pull/115
https://github.com/palantir/http-remoting/pull/235

and the collected rework here: https://github.com/palantir/http-remoting/pull/272

Worth noting that the original referenced issue in Brave is still ongoing.

uschi2000:
Got nothing to add.

SerialVelocity:
Cool. it seems like they're aiming to have it implemented in Brave 4 (from a comment 17 hours ago). Will you consider switching back to that when that happens? Also, are there any tools that can parse our current tracing into an easy to digest format?

markelliot:
What would be the advantage of changing at this point? The library still has an impedance mismatch with how we want to wire things, and it doesn't seem worth the effort to fix something that's working just fine for us.

Current tracing lib emits Zipkin format events to the log...so anything that can eat Zipkin events can parse what's produced.

On Mar 28, 2017, at 9:53 PM, Ben Grabham <notifications@github.com<mailto:notifications@github.com>> wrote:


Cool. it seems like they're aiming to have it implemented in Brave 4 (from a comment 17 hours ago). Will you consider switching back to that when that happens? Also, are there any tools that can parse our current tracing into an easy to digest format?

SerialVelocity:
Ah, ok. Didn't realise we did the conversion into a Zipkin compatible format.

PR #255, opened 2016-10-12
Add tracers-okhttp2 and tracers-okhttp3 project; instrument jaxrs and retrofit2 clients

uschi2000:


jaxrs-clients/src/main/java/com/palantir/remoting1/jaxrs/FeignJaxRsClientBuilder.java
```diff
@@ -126,7 +124,7 @@ private Decoder createDecoder(ObjectMapper objectMapper) {
                 new InputStreamDelegateDecoder(new TextDelegateDecoder(new JacksonDecoder(objectMapper))));
     }
 
-    private feign.Client createOkHttpClient(String userAgent) {
```
markelliot, line 129: where do we set this now?

PR #258, opened 2016-10-13
Add tracing aware executor utils

markelliot:


markelliot:
@uschi2000

markelliot:
some atlasdb prior art from its original code base:
https://github.com/palantir/atlasdb/blob/develop/commons-executors/src/main/java/com/palantir/common/concurrent/ExecutorInheritableThreadLocal.java
https://github.com/palantir/atlasdb/blob/develop/commons-executors/src/main/java/com/palantir/common/concurrent/PTExecutors.java
(these would be the more generic approach)

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareRunnable.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting1.tracing;
+
+import java.util.Deque;
+
+public final class TracingAwareRunnable implements Runnable {
```
uschi2000, line 21: can this be package-private?

tracing/src/main/java/com/palantir/remoting1/tracing/Traces.java
```diff
@@ -47,6 +49,21 @@
...
+     */
+    static Deque<TraceState> getCopyOfState() {
+        return new ArrayDeque<>(STATE.get());
+    }
+
+    /**
+     * Package-local state forcing mechanism for thread inheritance.
+     */
+    static void forceState(Deque<TraceState> stateToForce) {
+        STATE.get().remove();
```
uschi2000, line 63: this just removes one TraceState, right? don't you need to remove all of them?

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareCallable.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting1.tracing;
+
+import java.util.Deque;
+import java.util.concurrent.Callable;
+
+public final class TracingAwareCallable<V> implements Callable<V> {
```
uschi2000, line 22: can be package-private?

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareCallable.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final Callable<V> delegate;
+    private final Deque<TraceState> state;
+
+    public TracingAwareCallable(Callable<V> delegate) {
+        this.delegate = delegate;
+        this.state = Traces.getCopyOfState();
```
uschi2000, line 29: I think it'll be more clear if you inject the `state`

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareExecutorService.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public final Future<?> submit(Runnable task) {
+        return delegate.submit(new TracingAwareRunnable(task));
+    }
+
+    @Override
+    public final <T> List<Future<T>> invokeAll(Collection<? extends Callable<T>> tasks) throws InterruptedException {
+        return delegate.invokeAll(Collections2.transform(tasks, new TracedCallableConverter<T>()));
```
uschi2000, line 80: use static instance of converter

uschi2000: oh, actually that may not work that nicely due to generics?

uschi2000: this code path is going to be quite hot, so maybe it's worthing avoiding the object creation

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareCallable.java
uschi2000: alternatively, leave a class-level comment: 
```A {@link Callable} that uses the thread-local tracing state at the time of it's construction during its {@link #call execution}. The {@link TracingAwareCallable} constructor is called by a tracing-aware executor service on the same thread on which a user creates a Callable; the {@link call} function is executed on an arbitrary (likely different) thread with different tracing state. In order to execute the task with the original (and intuitively expected) tracing state, we remember the original state and set it for the duration of the {@link #call execution}.

```
```

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareRunnable.java
uschi2000: add a `See {@link TracingAwareCallable}` comment

tracing/src/main/java/com/palantir/remoting1/tracing/Traces.java
uschi2000: we should some tests for this ...

uschi2000 (changes requested):

markelliot:
do you want to take this over?

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareCallable.java
markelliot: Probably want to provide a way to wrap a `Callable<V>` generally, fine with that being a utility class that hides this implementation. Maybe the utility class should nest these implementations as private static classes?

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareCallable.java
markelliot: utility class mentioned in previous comment would make it more apparent how to do the wrapping and more obvious that the state is being injected. The main thing is that a caller that needs to wrap their `Callable<V>` submission should be able to do so without having to know about how to grab present state.

markelliot: Speaking of which, I suspect we probably want to do something in the ExecutorService implementations that checks if the callable being passed is already an implementation of this, so we can avoid the duplicate work.

tracing/src/main/java/com/palantir/remoting1/tracing/TracingAwareExecutorService.java
markelliot: unfortunately the `<T>` makes this very hard. `invokeAll` likely isn't a hot code path, though, the `Callable<T>`s are likely the bulk of the work.

markelliot:
obe

PR #261, opened 2016-10-13
Add TraceEnrichingFilterTest; DropwizardServer#configure applies TraceEnrichingFilter

uschi2000:


uschi2000:
copy from original PR

dropwizard-servers/src/main/java/com/palantir/remoting1/servers/dropwizard/DropwizardServers.java
```diff
@@ -44,5 +41,6 @@ public void accept(ExceptionMapper<? extends Throwable> mapper) {
                     }
                 });
         environment.jersey().register(new OptionalAsNoContentMessageBodyWriter());
+        environment.jersey().register(new TraceEnrichingFilter());
```
markelliot, line 44: we should convert `ExceptionMappers` to `JerseyComponents` and probably change `visitExceptionMappers` to return a `Collection<?>` of those components to register. In most cases, we might want to just return the class instance.

markelliot: also this whole class can go away now that there's nothing Dropwizard specific left?

PR #266, opened 2016-10-17
Add wrappers for tracing-aware ExecutorServices/Callables/Runnables

uschi2000:


tracing/src/main/java/com/palantir/remoting1/tracing/Traces.java
```diff
@@ -16,13 +16,103 @@
 
 package com.palantir.remoting1.tracing;
 
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
 
+/** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Traces {
```
markelliot: should we call this Tracers?

markelliot (approved):

tracing/src/main/java/com/palantir/remoting1/tracing/Traces.java
uschi2000: sure

uschi2000:
renamed class and rebased

ashrayjain:
@uschi2000 why was this copied from Guava? Do we still need it?
Edit: nvm, just saw its package private in Guava.

PR #268, opened 2016-10-17
Observers are identified by name in Tracer#subscribe

uschi2000:


uschi2000:
Thoughts on this, @markelliot ? Goal is to make sure different observers don't trample each other.

uschi2000:
Or maybe exactly is desired? I don't know.

markelliot:
how does this make it more likely observers won't trample each other?

uschi2000:
my explanation was the opposite of what I had in mind: this makes it so that trampling must be explicit. if your library registers an "SLF4J" observer and mine does too, then this will throw. i'm a little worried that we'll end with an unbounded number of similar observers if we dedupe them by object identity.

markelliot:
Ok, agree this does that, though the namespacing is still unbounded, and it might be confusing to have your subscriber trampled, so we should probably add some logging if you trample but don't have object equivalence?

We can also say there's exactly one observer, and allow for a multi-observer observer that aggregates.

Practically speaking, I expect there to be ~2 observers max, ever.

uschi2000:
the multi thing in itself doesn't quite help if someone else can come along
and replace it with something else.
i be down for (1) named observers, (2) "warn" log entry when you overwrite
an existing one (by name). deal?

markelliot:
Yep, that's fine.

markelliot:
Probably also want to warn any time size of the map grows to > 5, too.

markelliot (approved):

PR #270, opened 2016-10-18
Fix/serializable error handler null handling

### LINKED ISSUE #260: RetrofitSerializableErrorErrorHandler assumes request body is not null.
https://github.com/palantir/http-remoting/blob/develop/retrofit-clients/src/main/java/com/palantir/remoting1/retrofit/errors/RetrofitSerializableErrorErrorHandler.java

Assumes that the request body is not null. This can lead to an NPE rather than properly returning that the body is null.

jamesneale:
Fixes #260.


uschi2000 (approved):

uschi2000:
Thanks!

PR #274, opened 2016-10-18
jaxrs shadow does not include :tracing project

uschi2000:


jaxrs-clients/build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     shadow project(':error-handling')
     shadow project(':service-config')
     shadow project(':ssl-config')
+    shadow project(':tracing-okhttp3')
```
markelliot: are we able to programmatically tie this to the shadowJar dependencies configuration?

markelliot (approved):

jaxrs-clients/build.gradle
uschi2000: i don't know a way

PR #275, opened 2016-10-18
Add TraceIdLoggingFilter

uschi2000:


uschi2000:
@schlosna : thoughts on closing the MDC put? (you had made the corresponding change earlier, if I remember right)

uschi2000:
As far as I know, this is the last thing that needs to go in before we can make another RC.

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/HttpRemotingJerseyFeature.java
```diff
@@ -51,6 +51,9 @@ public boolean configure(FeatureContext context) {
 
         // Tracing
         context.register(new TraceEnrichingFilter());
+//        environment.servlets()
+//                -                .addFilter(TraceIdLoggingFilter.class.getSimpleName(), TraceIdLoggingFilter.INSTANCE)
+//                -                .addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
```
markelliot: just delete?

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -53,6 +59,10 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                 Tracer.startSpan(operation, spanId); // caller's span is this span's parent.
             }
         }
+
+        // Give SLF4J appenders access to the trace id
+        // TODO(rfink) We should use putCloseable; when and how can we remove it though? There is no filter chain.
+        MDC.put(MDC_KEY, Tracer.getTraceId());
```
markelliot, line 65: in addition to this we should add to the request attributes, e.g.:

```
// convention is package naming
requestContext.setProperty("com.palantir.remoting1.traceId", Tracer.getTraceId());
```

markelliot: (this allows picking things up in the Access logger, assuming it's not asynchronous or logback fixes the bug I submitted a patch for)

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/HttpRemotingJerseyFeature.java
uschi2000: done

jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/TraceEnrichingFilter.java
uschi2000: assuming it's not _synchronous_?

uschi2000: done

markelliot: actually meant synchronous, but whether they're sync or async shouldn't actually matter (modulo fact a bug exists)

markelliot (approved):

PR #322, opened 2017-01-20
Remove support for Jackson 2.4

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/322)


ash211:
Spark is on Jackson 2.6.5 as of Spark 2.0.0 -- https://github.com/apache/spark/blob/v2.0.0/pom.xml#L165

So as long as this changes the http-remoting dependency to be onto a version of Jackson that's compatible with code compiled against Jackson 2.6.5 we should be good here.

PR #331, opened 2017-01-25
Add OptionalObjectToStringConverterFactory to support Optional parameters in retrofit2 clients, fixes #326

### LINKED ISSUE #326: Add Java8-Optional functionality to retrofit2-clients

uschi2000:
…ters in retrofit2 clients, fixes 326

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/331)

### DIFF
--- a/retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/OptionalObjectToStringConverterFactory.java
+++ b/retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/OptionalObjectToStringConverterFactory.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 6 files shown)


retrofit2-clients/src/main/java/com/palantir/remoting1/retrofit2/OptionalObjectToStringConverterFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+        return null;
+    }
+
+    enum Java8OptionalStringConverter implements Converter<java.util.Optional<?>, String> {
+        INSTANCE;
+
+        @Override
+        public String convert(java.util.Optional<?> value) throws IOException {
+            return value.map(Object::toString).orElse("");
```
markelliot, line 64: don't we want to nest the mapping here? ie this should be non-empty optionals get converted using the converter of appropriate type for the inner type?

markelliot: (say by using the JacksonConverter?)

uschi2000: Arguably, all query or path params should be mostly stringish thing anyway, no?

uschi2000:
I'd also prefer not diverging from plain retrofit2 behavior too much, if
possible. out-of-the-box behavior is to serialize everything as #toString,
even if you apply the JacksonConverterFactory.

markelliot:
What about return types?

markelliot:
Suppose I have some API:

interface Blargh {
    Optional<AliasType> getThing();
}

For:

class AliasType {
    private final int val;

    @JsonCreator
    public AliasType(int val) {
        this.val = val;
    }

    @JsonValue
    public int get() {
        return val;
    }
}

Will this work?

uschi2000:
I think so, let me try.

uschi2000:
yep, see test.

PR #340, opened 2017-01-27
Use ObjectMappers factory instead of custom mappers

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/340)


markelliot:
👍

PR #345, opened 2017-02-04
Binary message handling

ellisjoe:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/345)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java
@@ -56,7 +56,6 @@ public final class BinaryMessageProvider extends AbstractMessageReaderWriterProv
             MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
             throws IOException, WebApplicationException {
         MAPPER.writeValue(entityStream, bytes);
-        //entityStream.write(Base64.encode(bytes));
     }
 
     @Override
@@ -69,8 +68,6 @@ public final class BinaryMessageProvider extends AbstractMessageReaderWriterProv
             MultivaluedMap<String, String> httpHeaders, InputStream entityStream)
             throws IOException, WebApplicationException {
         return MAPPER.readValue(entityStream, type);
-//        byte[] encodedBytes = ByteStreams.toByteArray(entityStream);
-//        return Base64.decode(encodedBytes);
     }
 
 }
--- a/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting1.servers.jersey;
+
+import com.google.common.io.ByteStreams;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.Provider;
+import org.glassfish.jersey.internal.util.Base64;
+import org.glassfish.jersey.message.internal.AbstractMessageReaderWriterProvider;
+
+/**
+ * Reads and writes {@code byte[]}s as base64 encoded strings.
+ */
+@Provider
+public final class BinaryMessageProvider extends AbstractMessageReaderWriterProvider<byte[]> {
+
+    // Jersey ignores this
+    @Override
+    public long getSize(byte[] bytes, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return 0;
+    }
+
+    @Override
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return byte[].class.isAssignableFrom(type);
+    }
+
+    @Override
+    public void writeTo(byte[] bytes, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType,
+            MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
+            throws IOException, WebApplicationException {
+        entityStream.write(Base64.encode(bytes));
+    }
+
+    @Override
+    public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return byte[].class.isAssignableFrom(type);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java
```diff
@@ -0,0 +1,69 @@
...
+
+    // Jersey ignores this
+    @Override
+    public long getSize(byte[] bytes, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return 0;
+    }
+
+    @Override
+    public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
+        return byte[].class.isAssignableFrom(type);
```
markelliot, line 51: maybe just:

    type instanceof byte[]

markelliot: oh nvm

jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java
```diff
@@ -0,0 +1,84 @@
...
+        @Path("binary-return")
+        public byte[] showWithQueryParam(String string) {
+            return string.getBytes(StandardCharsets.UTF_8);
+        }
+
+        @POST
+        @Path("binary-consume")
+        public String byteArrayToString(byte[] bytes) {
+            return new String(bytes, StandardCharsets.UTF_8);
+        }
```
markelliot, line 92: also test HeaderParam and QueryParam byte[] use? (sadly I think these may show up in the real world)

markelliot: (also, hint: you likely need ParamConverters too)

uschi2000:
Do we want binary data in header/query/path params?

markelliot:
unfortunately I think this isn't a question so much of want vs. actually have.

uschi2000:
But we can control what interfaces are allowed, at least in conjure.

uschi2000:
Review status: 0 of 3 files reviewed at latest revision, 9 unresolved discussions.

---

*[jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java, line 29 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5q1Xzwgcw8bwEUAfX:-Kc5q1Y-K1xIq8rP4WpL:b-5nzp2g) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java#L29)):*
> ```Java
> import javax.ws.rs.core.MultivaluedMap;
> import javax.ws.rs.ext.Provider;
> import org.glassfish.jersey.internal.util.Base64;
> ```

would prefer java.util.Base64, I think

---

*[jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java, line 33 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5p2_cGHbuT15-8B7N:-Kc5p2_cGHbuT15-8B7O:bi14756) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java#L33)):*
> ```Java
> 
> /**
>  * Reads and writes {@code byte[]}s as base64 encoded strings.
> ```

nit: base64-encoded

---

*[jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java, line 41 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5q4UZNbSueAdCMoYE:-Kc5q4UZNbSueAdCMoYF:b-9bw09y) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java#L41)):*
> ```Java
>     @Override
>     public long getSize(byte[] bytes, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
>         return 0;
> ```

jersey's ByteArrayProvider returns bytes.length, let's do the same

---

*[jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java, line 46 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5qDLsIx6SKQLfLKSm:-Kc5qDLsIx6SKQLfLKSn:b-4rnaj4) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/main/java/com/palantir/remoting1/servers/jersey/BinaryMessageProvider.java#L46)):*
> ```Java
>     @Override
>     public boolean isWriteable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
>         return byte[].class.isAssignableFrom(type);
> ```

jersey's implementation does types == byte[].class. which one is preferred?

---

*[jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java, line 31 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5qdH_kyoHShKas6q0:-Kc5qdH_kyoHShKas6q1:bc07tp0) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java#L31)):*
> ```Java
> import javax.ws.rs.core.Application;
> import javax.ws.rs.core.MediaType;
> import org.glassfish.jersey.internal.util.Base64;
> ```

java.util.Base64

---

*[jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java, line 38 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5qh21sdD071yR2RT8:-Kc5qh21sdD071yR2RT9:b9wxmk0) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java#L38)):*
> ```Java
> public final class BinaryMessageProviderTest extends JerseyTest {
> 
>     private static final String STRING = "bytes";
> ```

let's add a second test case with `byte[] BINARY_BYTES = randomBytes(100)`

---

*[jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java, line 63 at r1](https://reviewable.io:443/reviews/palantir/http-remoting/345#-Kc5r8QeNlAZ7UDGVoak:-Kc5r8QeNlAZ7UDGVoal:b-oeh1fh) ([raw file](https://github.com/palantir/http-remoting/blob/b943fb9a93b76a5564cb1d978a5dd2c327167058/jersey-servers/src/test/java/com/palantir/remoting1/servers/jersey/BinaryMessageProviderTest.java#L63)):*
> ```Java
>         String actual = target("/binary-consume")
>                 .request()
>                 .buildPost(Entity.json(ENCODED_BYTES))
> ```

would prefer to send the base64 encoded String here, not the byte[] for the base64-encoded String. less magic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/http-remoting/345)*

uschi2000:
ah, let's add an e2e test too with a JaxRsClient

ellisjoe:
re supporting binary headers and query params: do we actually even want to support `byte[]` as the body? doesn't seem particularly useful and we probably should just support streaming

uschi2000:
Isn't that what this PR is about?

ellisjoe:
yeah. was talking to mark and maybe this isn't useful?

uschi2000:
I'm fine either way. So you want to ditch your conjure PR too then?

ellisjoe:
no, I think we should just allow binary (byte[]) as a field in data objects. not as a raw input/output type

uschi2000:
works for me.

Issue #347, opened 2017-02-09
SerializableStackTraceElement is hard to read in the logs


a-k-g:
When a RemoteException is logged, it prints out the `SerializableError remoteException` with a simple `toString()`. The `SerializableError` has a custom `toString()` which nicely prints out each individual `SerializableStackTraceElement` in the remote stacktrace. However, the `SerializableStackTraceElement` doesn't have a `toString()` and uses the default one generated by Immutables. This results in a hard to read log which looks something like -

```
Remote stacktrace:
	at SerializableStackTraceElement{className=com.palantir.remoting.http.errors.SerializableErrorToExceptionConverter, methodName=getException, fileName=SerializableErrorToExceptionConverter.java, lineNumber=82}
	at SerializableStackTraceElement{className=com.palantir.remoting.http.errors.FeignSerializableErrorErrorDecoder, methodName=decode, fileName=FeignSerializableErrorErrorDecoder.java, lineNumber=43}
	at SerializableStackTraceElement{className=feign.SynchronousMethodHandler, methodName=executeAndDecode, fileName=SynchronousMethodHandler.java, lineNumber=126}
	at SerializableStackTraceElement{className=feign.SynchronousMethodHandler, methodName=invoke, fileName=SynchronousMethodHandler.java, lineNumber=74}
	at SerializableStackTraceElement{className=feign.ReflectiveFeign$FeignInvocationHandler, methodName=invoke, fileName=ReflectiveFeign.java, lineNumber=103}
	at SerializableStackTraceElement{className=com.sun.proxy.$Proxy55, methodName=openFile, lineNumber=-1}
	at SerializableStackTraceElement{className=sun.reflect.NativeMethodAccessorImpl, methodName=invoke, fileName=NativeMethodAccessorImpl.java, lineNumber=62}
	at SerializableStackTraceElement{className=sun.reflect.DelegatingMethodAccessorImpl, methodName=invoke, fileName=DelegatingMethodAccessorImpl.java, lineNumber=43}
	at SerializableStackTraceElement{className=java.lang.reflect.Method, methodName=invoke, fileName=Method.java, lineNumber=498}
	at SerializableStackTraceElement{className=com.palantir.tritium.proxy.InvocationEventProxy, methodName=execute, fileName=InvocationEventProxy.java, lineNumber=111}
[...]

PR #349, opened 2017-02-09
Delegate SerializableStackTraceElement.toString() to Java's own

a-k-g:
#347

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/349)

### DIFF
--- a/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
+++ b/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
@@ -92,10 +92,10 @@ public abstract class SerializableError implements Serializable {
                         @Override
                         public SerializableStackTraceElement apply(@Nullable StackTraceElement input) {
                             return SerializableStackTraceElement.builder()
-                                    .className(Optional.ofNullable(input.getClassName()))
-                                    .methodName(Optional.ofNullable(input.getMethodName()))
+                                    .className(input.getClassName())
+                                    .methodName(input.getMethodName())
                                     .fileName(Optional.ofNullable(input.getFileName()))
-                                    .lineNumber(Optional.ofNullable(input.getLineNumber()))
+                                    .lineNumber(input.getLineNumber())
                                     .build();
                         }
                     }));
--- a/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableStackTraceElement.java
+++ b/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableStackTraceElement.java
@@ -30,13 +30,23 @@ import org.immutables.value.Value;
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class SerializableStackTraceElement implements Serializable {
 
-    public abstract Optional<String> getClassName();
+    public abstract String getClassName();
 
-    public abstract Optional<String> getMethodName();
+    public abstract String getMethodName();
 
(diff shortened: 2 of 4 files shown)


error-handling/src/main/java/com/palantir/remoting2/errors/SerializableStackTraceElement.java
```diff
@@ -30,13 +30,23 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class SerializableStackTraceElement implements Serializable {
 
-    public abstract Optional<String> getClassName();
+    public abstract String getClassName();
```
a-k-g: Not sure if we really needed these `Optionals` as `StackTraceElement` has only the following constructor-
```
    public StackTraceElement(String declaringClass, String methodName,
                             String fileName, int lineNumber) {
        this.declaringClass = Objects.requireNonNull(declaringClass, "Declaring class is null");
        this.methodName     = Objects.requireNonNull(methodName, "Method name is null");
        this.fileName       = fileName;
        this.lineNumber     = lineNumber;
    }
```
And given that we're 2.0-ing it seems like a good opportunity to make this breaking change.

error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
```diff
@@ -92,10 +92,10 @@ public static SerializableError of(
                         @Override
                         public SerializableStackTraceElement apply(@Nullable StackTraceElement input) {
                             return SerializableStackTraceElement.builder()
-                                    .className(Optional.ofNullable(input.getClassName()))
-                                    .methodName(Optional.ofNullable(input.getMethodName()))
+                                    .className(input.getClassName())
```
uschi2000, line 95: cheers, this was silly

error-handling/src/main/java/com/palantir/remoting2/errors/SerializableStackTraceElement.java
uschi2000: this isn't necessarily coming from java, so we need to keep the optionality in the wire format

error-handling/src/main/java/com/palantir/remoting2/errors/SerializableStackTraceElement.java
```diff
@@ -30,13 +30,23 @@
...
+    public abstract String getMethodName();
 
     public abstract Optional<String> getFileName();
 
-    public abstract Optional<Integer> getLineNumber();
+    public abstract int getLineNumber();
+
+    @Override
+    public final String toString() {
+        StackTraceElement element = new StackTraceElement(getClassName(),
```
uschi2000: cannot guarantee that these are non-null, so cannot simply construct this. IMO, it's reasonable to do:
```
StackTraceElement element = new StackTraceElement(
  getClassName().orElse("UnknownClass"),
  getMethodName().orElse("unknownMethod"),
  getFileName.orElse(null),
  getLineNumer.orElse(0));
```

uschi2000: please add a test verifying that this works when class and method are invalid java identifiers, i.e., 'Bogus Class Name', and 'bogus method name'.

uschi2000 (changes requested):
few changes, see above

uschi2000 (approved):
cheers!

PR #374, opened 2017-03-16
Drop Guava back to 18.0

markelliot:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/374)


markelliot:
@uschi2000 I think we probably want to take this change before we cut 2.1.0 for real. Thoughts?

uschi2000:
Yep

nathklei:
@markelliot Was upgrading to DW 1.1.0 (came out yesterday) which bumps to 21.0, and noticed this.  What was the rationale for the downgrade?

PR #375, opened 2017-03-17
Fix shading in jaxrs-client

uschi2000:
Don't merge yet; need to check that this works in our projects.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/375)

### DIFF
--- a/jaxrs-clients/build.gradle
+++ b/jaxrs-clients/build.gradle
@@ -29,18 +29,23 @@ dependencies {
     testCompile "org.hamcrest:hamcrest-all"
 
     // Keep in sync with the dependencies excluded from shadowJar.
+    shadow project(":extras:jackson-support")
+    shadow project(":extras:refresh-utils")
+    shadow project(":http-clients-api")
+    shadow project(":http2-agent")
+    shadow project(":tracing")
+    shadow project(":tracing-okhttp3")
     shadow project(':error-handling')
     shadow project(':service-config')
     shadow project(':ssl-config')
-    shadow project(":tracing")
-    shadow project(":tracing-okhttp3")
-    shadow "javax.ws.rs:javax.ws.rs-api"
     shadow "com.fasterxml.jackson.core:jackson-databind"
     shadow "com.fasterxml.jackson.datatype:jackson-datatype-guava"
     shadow "com.fasterxml.jackson.datatype:jackson-datatype-jdk8"
     shadow "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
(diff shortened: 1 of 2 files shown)


uschi2000:
FYI @diogoholanda

uschi2000:
Ready for review, @markelliot

jaxrs-clients/build.gradle
```diff
@@ -50,8 +55,15 @@ shadowJar {
     relocate('zipkin', 'httpremoting.shaded.zipkin')
     relocate('org.joda.time', 'httpremoting.shaded.org.joda.time')
     relocate('okio', 'httpremoting.shaded.okio')
+    relocate('com.ea.agentloader', 'httpremoting.shaded.com.ea.agentloader')
```
markelliot: Surprised we're relocating this.

jaxrs-clients/build.gradle
```diff
@@ -50,8 +55,15 @@ shadowJar {
...
+    relocate('org.mortbay.jetty.alpn.agent', 'httpremoting.shaded.org.mortbay.jetty.alpn.agent')
```
markelliot: And also this.

uschi2000:
Other option is to add an external dependency on it. Fine with me either
way, but the current model (include in shadowjar with relocation) is broken.

markelliot:
i think external dependency on that would be fine

uschi2000:
Done. Turns out this actually had a bug (added dependency on http2-agent project and contained the relocated classes)

jaxrs-clients/build.gradle
```diff
@@ -50,8 +55,13 @@ shadowJar {
     relocate('zipkin', 'httpremoting.shaded.zipkin')
     relocate('org.joda.time', 'httpremoting.shaded.org.joda.time')
     relocate('okio', 'httpremoting.shaded.okio')
+    relocate('org.jvnet.animal_sniffer', 'httpremoting.shaded.org.jvnet.animal_sniffer')
```
markelliot, line 58: what pulls this in?

markelliot: think this is probably related to the agentloader of the alpn thing, in whcih case also want to exclude it

uschi2000:
Feign if I remember right. Good old animal sniffer.

markelliot:
That's weird, because it's new here?

uschi2000:
Double-checked, it comes from Feign 8.17.

markelliot:
Cool, then +1

PR #376, opened 2017-03-17
Add support for Java 8 primitive optionals

markelliot:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/376)

### DIFF
--- a/jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/feignimpl/Java8EmptyOptionalDoubleExpander.java
+++ b/jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/feignimpl/Java8EmptyOptionalDoubleExpander.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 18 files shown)


uschi2000 (approved):
Wow, that's tedious. Tried very hard to find a typo, but didn't manage to spot one.

uschi2000:
It's a lot of code to avoid boxing/unboxing. Can't judge if it's worth it :/

markelliot:
Yep, it's mostly boilerplate. We could probably use Jackson to coerce everything that's a PathParam or HeaderParam that's not already a string, but since that wasn't how we were already doing this I did the tedious repetition thing.

I suspect better to support these things than not, but likely worth a discussion about code and test quality of the whole project.

Also, latest Dropwizard already has all the Jersey stuff by default, we may want to split that stuff out so that Dropwizard 1+ users do not double install.

For good measure, I did also check to see if we could depend on dropwizard-jersey, but it unfortunately has a long dependency tree, including Jackson, which is a persistent issue for us when trying to run servers that talk to Spark.

On Mar 17, 2017, at 4:34 AM, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:


It's a lot of code to avoid boxing/unboxing. Can't judge if it's worth it :/

uschi2000:
Half of this will go away when we remove jaxrs-clients, and yeah, definitely agree the code&test quality needs some love here. Anyway, +1 for this PR.

PR #378, opened 2017-03-23
compare fqdn case insensitively and paths case sensitively

### LINKED ISSUE #377: Does http-remoting automatically lower case uris?
I'm generating a Retrofit2Client with a URI whose FQDN has upper case letters. When I actually make a request with that client, I hit this error: https://github.com/palantir/http-remoting/blob/ae3be5349b1634183798ff52cc5df40925af7840/retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java#L101

"Unrecognized server URI in the request `lower case fqdn`. Supported servers are `original upper case fqdn`". 

Should we be doing case insensitive comparisons instead?

cc @uschi2000

rahij:
Fixes #377 

cc @ash211 @uschi2000

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/378)

### DIFF
--- a/retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
+++ b/retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
@@ -86,12 +86,12 @@ public final class MultiServerRetryInterceptor implements Interceptor {
     }
 
     private Request redirectRequest(Request request, String redirectToUri) {
-        String requestUri = request.url().toString();
+        HttpUrl requestUrl = request.url();
 
         String matchingUri = null;
         // Find which server from `uris` is used in the current request, then ...
         for (String uri : uris) {
-            if (requestUri.startsWith(uri)) {
+            if (doServersMatch(requestUrl, HttpUrl.parse(uri))) {
                 matchingUri = uri;
                 break;
             }
@@ -100,16 +100,26 @@ public final class MultiServerRetryInterceptor implements Interceptor {
         if (matchingUri == null) {
             throw new IllegalStateException(String.format("Unrecognized server URI in the request %s. "
                             + "Supported servers are %s. Did you use different server URI for the initial request?",
-                    requestUri, uris));
+                    requestUrl, uris));
         }
 
         // ... replace it with the URI of the server to redirect to.
-        String newRequestUrl = requestUri.replaceFirst(matchingUri, redirectToUri);
+        String newRequestUrl = requestUrl.toString().replaceFirst(matchingUri, redirectToUri);
         return request.newBuilder()
                 .url(HttpUrl.parse(newRequestUrl))
                 // Request.this.tag field by default points to request itself if it was not set in RequestBuilder.
                 // We don't want to reference old request in new one - that is why we need to reset tag to null.
                 .tag(request.tag().equals(request) ? null : request.tag())
                 .build();
+    }
+
+    private boolean doServersMatch(HttpUrl serverUri, HttpUrl server) {
+        String serverUriFqdn = String.format("%s://%s:%s", serverUri.scheme(), serverUri.host(), serverUri.port());
+        String serverUriPath = serverUri.encodedPath();
+
+        String serverFqdn = String.format("%s://%s:%s", server.scheme(), server.host(), server.port());
+        String serverPath = server.encodedPath();
+
+        return serverUriFqdn.equalsIgnoreCase(serverFqdn) && serverUriPath.startsWith(serverPath);
     }
 }
--- a/retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
+++ b/retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.stream.Collectors;
 import okhttp3.HttpUrl;
 import okhttp3.Interceptor;
 import okhttp3.Request;
@@ -39,7 +40,7 @@ import org.slf4j.LoggerFactory;
 public final class MultiServerRetryInterceptor implements Interceptor {
     private static final Logger logger = LoggerFactory.getLogger(MultiServerRetryInterceptor.class);
 
-    private final List<String> uris;
+    private final List<HttpUrl> urls;
 
     private static final Comparator<HttpUrl> FQDN_COMPARATOR =
             Comparator.comparing(HttpUrl::scheme)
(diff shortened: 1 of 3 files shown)


retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -86,12 +86,12 @@ public Response intercept(Chain chain) throws IOException {
...
 
     private Request redirectRequest(Request request, String redirectToUri) {
-        String requestUri = request.url().toString();
+        HttpUrl requestUrl = request.url();
 
         String matchingUri = null;
         // Find which server from `uris` is used in the current request, then ...
         for (String uri : uris) {
-            if (requestUri.startsWith(uri)) {
+            if (doServersMatch(requestUrl, HttpUrl.parse(uri))) {
```
uschi2000: want to keep a List<HttpUrl> instead of strings?

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -86,12 +86,12 @@ public Response intercept(Chain chain) throws IOException {
     }
 
     private Request redirectRequest(Request request, String redirectToUri) {
-        String requestUri = request.url().toString();
+        HttpUrl requestUrl = request.url();
 
         String matchingUri = null;
         // Find which server from `uris` is used in the current request, then ...
         for (String uri : uris) {
```
uschi2000: pick one or the other: uri/url

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -100,16 +100,26 @@ private Request redirectRequest(Request request, String redirectToUri) {
...
+        String newRequestUrl = requestUrl.toString().replaceFirst(matchingUri, redirectToUri);
         return request.newBuilder()
                 .url(HttpUrl.parse(newRequestUrl))
                 // Request.this.tag field by default points to request itself if it was not set in RequestBuilder.
                 // We don't want to reference old request in new one - that is why we need to reset tag to null.
                 .tag(request.tag().equals(request) ? null : request.tag())
                 .build();
     }
+
+    private boolean doServersMatch(HttpUrl serverUri, HttpUrl server) {
```
uschi2000: - docs, please
- parameters should have symmetric names; for instance `left`, `right`
// edit: scrap that, it's not symmetric. so how about: fullUrl vs prefixUrl or something like that.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -100,16 +100,26 @@ private Request redirectRequest(Request request, String redirectToUri) {
...
+        String serverUriFqdn = String.format("%s://%s:%s", serverUri.scheme(), serverUri.host(), serverUri.port());
```
uschi2000: as above... `leftFqdn` and `rightFqdn`

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
```diff
@@ -100,16 +100,26 @@ private Request redirectRequest(Request request, String redirectToUri) {
...
+        String serverUriPath = serverUri.encodedPath();
+
+        String serverFqdn = String.format("%s://%s:%s", server.scheme(), server.host(), server.port());
```
uschi2000: actually, let's do something like this:
```
// does HttpUrl::host always return a lowercase host?
private static final Comparator<HttpUrl> FQDN_COMPARATOR = Comparator.comparing(HttpUri::scheme).thenComparing(HttpUri::host).thenComparing(HttpUri::port);

boolean startsWith(HttpUrl fullUrl, HttpUrl prefixUrl) {
  return
    FQDN_COMPARATOR.compare(fullUrl, prefixUrl) == 0
    && fullUrl.encodedPath().startsWith(prefixUrl.encodedPath());
}
```

retrofit2-clients/src/test/java/com/palantir/remoting2/retrofit2/TestService.java
```diff
@@ -51,4 +51,7 @@
     // Note: Retrofit2 will discard any baseUrl path when the endpoint path starts with "/"
     @GET("/absolute")
     Call<String> getAbsolute();
+
+    @GET("/Api")
+    Call<String> getAbsoluteApiTitleCase();
```
uschi2000, line 56: can just use one of the existing endpoints for the new tests

rahij: I need an absolute url with uppercase letters in it in order to test this correctly :/

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
rahij: We seem to use uri everywhere including when calling the constructors and the Request object has `url` anyway. I could pass a String to doServersMatch for the prefixUrl, but not sure if that's any cleaner

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
rahij: Done. Yes, it does return a lower case host.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
uschi2000: ah, swallowed some characters, sorry; wondering if the class should keep a `List<HttpUrl>` instead of a `List<String>`?

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
uschi2000: ok

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/MultiServerRetryInterceptor.java
rahij: Yeah good idea. Done

uschi2000 (approved):

PR #381, opened 2017-03-23
Introduce fromClassStrict and remove canDetectUserAgent

leeavital:
Addressing offline feedback from @markelliot

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/381)

### DIFF
--- a/http-clients-api/src/main/java/com/palantir/remoting2/clients/UserAgents.java
+++ b/http-clients-api/src/main/java/com/palantir/remoting2/clients/UserAgents.java
@@ -55,7 +55,7 @@ public final class UserAgents {
 
         Preconditions.checkArgument(userAgent.isPresent(), "Implementation-Title missing from the manifest of %s",
                 clazz);
-        Preconditions.checkArgument(userAgent.isPresent(), "Implementation-Version missing from the manifest of %s",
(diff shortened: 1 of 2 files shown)


markelliot (approved):

pnepywoda:
@leeavital @markelliot 
This is a pretty annoying change - it means you can't startup whatever process uses `fromClassStrict` if it's not running from an already-built-jar....meaning you can't run it in your IDE :(

markelliot:
Yep, this was the behavior Lee was looking for. Can certainly roll it back, or we can make it throw conditional on some debug system property.

pnepywoda:
Lee, what's the rationale for disallowing running in an IDE?

leeavital:
I had made this change https://github.com/palantir/sls-packaging/pull/207 to allow for running from gradle. That's how I usually run things, so I overlooked running from the IDE.

The idea was to get people to use the jar properties to get nice user agents, not to break IDE workflows.

We could do what @markelliot  suggesting add a `com.palantir.remoting.debug` property to suppress this, or we could just turn these exceptions into logs. 

I'm partial to logging because using a debug property would mean users need to mess with their launch configs and/or gradle configs

uschi2000:
I like the logging idea.

leeavital:
@pnepywoda looks like a workaround was added in #390

PR #388, opened 2017-04-03
Don't install Http2Agent by default

justinuang:
- By the time the ClientBuilder is referenced, it is often too late for
  the agent to be able to replace the core SSL classes.
- If it is too late, a cryptic java.lang.NoSuchFieldError: EXT_ALPN is
  thrown
- The framework using remoting should be responsible for installing the
  Http2Agent (for example witchcraft or spark-module-lib), as to ensure
  that it is done early enough.

Discussed offline with @markelliot

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/388)


jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/FeignJaxRsClientBuilder.java
```diff
@@ -61,10 +60,6 @@
 
 public final class FeignJaxRsClientBuilder extends ClientBuilder {
 
-    static {
-        Http2Agent.install();
```
uschi2000, line 65: can we remove the project dependency?

justinuang:
@uschi2000 can we get this merged and tagged, it's a blocker for the platform 1.2.0 for monocat

uschi2000:
Closing in favour of https://github.com/palantir/http-remoting/pull/393

PR #390, opened 2017-04-04
more fromClass fun

markelliot:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/390)


markelliot:
@uschi2000

uschi2000 (approved):

uschi2000 (approved):

ghost:
Hmm, if we deprecate the strict version and use the non-strict version this means that if your service doesn't contain the version in the manifest then the only way to catch this is by integration testing...

ghost:
This is especially easy to mess up considering that the version is injected by a copy/pasted bit of gradle code we move around

ghost:
I guess I can achieve my desired behaviour by doing:

```
UserAgents.fromClass(FooServer.class, null, null)
```

PR #396, opened 2017-04-12
[blocked on upstream] when converting optional-valued query parameters, convert to null if not present

### LINKED ISSUE #384: optional path parameter on retrofit
For a retrofit client with an optional path parameter e.g. `@Query("foo") Optional<String> foo`, passing in an `Optional.empty()` will result in the query parameter still being included in the request, just with a no value... e.g. `GET my/path?foo=`

then on the server side, this results in an `Optional.of("")` rather than the expected `Optional.empty()`. Not sure if there's something we can do here...

bavardage:
fixes #384

this is blocked on retrofit upstream https://github.com/square/retrofit/pull/2243

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/396)


uschi2000:
Do you still need this, @bavardage ?

uschi2000:
I think the upstream thing has merged in May.

bavardage:
yeah we should do this, it's really horrible to have different semantics between feign and retrofit

uschi2000:
closing in favour of https://github.com/palantir/http-remoting/pull/501

PR #401, opened 2017-04-13
cbor support

bavardage:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/401)

### DIFF
--- a/jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
+++ b/jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ */
+
+package feign;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.net.HttpHeaders;
+import com.palantir.remoting2.jaxrs.feignimpl.HeaderAccessUtils;
+import feign.codec.DecodeException;
+import feign.codec.Decoder;
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.Type;
+import java.util.Collection;
+
+
+public final class CborDelegateDecoder implements Decoder {
+
+    private final ObjectMapper cborObjectMapper;
+    private final Decoder delegate;
+
+    public CborDelegateDecoder(ObjectMapper cborObjectMapper, Decoder delegate) {
+        this.cborObjectMapper = cborObjectMapper;
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Object decode(Response response, Type type) throws IOException, DecodeException, FeignException {
+        Collection<String> contentTypes =
+                HeaderAccessUtils.caseInsensitiveGet(response.headers(), HttpHeaders.CONTENT_TYPE);
+        if (contentTypes == null) {
+            contentTypes = ImmutableSet.of();
+        }
+
+        if (contentTypes.size() == 1
+                && Iterables.getOnlyElement(contentTypes, "").startsWith(CborDelegateEncoder.MIME_TYPE)) {
+            InputStream inputStream = response.body().asInputStream();
+            if (!inputStream.markSupported()) {
+                inputStream = new BufferedInputStream(inputStream, 1);
+            }
+
+            // Read the first byte to see if we have any data
+            inputStream.mark(1);
+            if (inputStream.read() == -1) {
+                return null; // Eagerly returning null avoids "No content to map due to end-of-input"
+            }
+
+            inputStream.reset();
+            return cborObjectMapper.readValue(inputStream, cborObjectMapper.constructType(type));
+
+        } else {
+            return delegate.decode(response, type);
+        }
+    }
+
+}
--- a/jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
+++ b/jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
@@ -11,9 +11,8 @@ import com.google.common.net.HttpHeaders;
 import com.palantir.remoting2.jaxrs.feignimpl.HeaderAccessUtils;
 import feign.codec.DecodeException;
 import feign.codec.Decoder;
-import java.io.BufferedInputStream;
 import java.io.IOException;
-import java.io.InputStream;
+import java.io.PushbackInputStream;
 import java.lang.reflect.Type;
 import java.util.Collection;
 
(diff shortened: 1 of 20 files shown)


bavardage:
not sure how contentious this is...

bavardage:
@dtobin

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/CborConverterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Converter.Factory;
+import retrofit2.Retrofit;
+import retrofit2.http.Headers;
+
+public final class CborConverterFactory extends Converter.Factory {
+
+    private static final MediaType MEDIA_TYPE = MediaType.parse("application/cbor");
```
markelliot: CBOR_MIME_TYPE

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -61,7 +61,9 @@ public Retrofit2ClientBuilder(ClientConfig config) {
         Retrofit retrofit = new Retrofit.Builder()
                 .client(client)
                 .baseUrl(sanitizedUris.get(0))
-                .addConverterFactory(JacksonConverterFactory.create(OBJECT_MAPPER))
+                .addConverterFactory(new CborConverterFactory(
+                        JacksonConverterFactory.create(OBJECT_MAPPER),
```
markelliot, line 66: inline OBJECT_MAPPER or pull out the one below

bavardage:
image

is the http2 agent stuff known to be flakey? (I didn't touch this stuff afaik)

bavardage:
retriggering... have verified the tests pass locally

uschi2000:
haven't seen that error before

bavardage:
retriggering made it go away ¯\\\_(ツ)_/¯

jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
```diff
@@ -0,0 +1,60 @@
...
+                HeaderAccessUtils.caseInsensitiveGet(response.headers(), HttpHeaders.CONTENT_TYPE);
+        if (contentTypes == null) {
+            contentTypes = ImmutableSet.of();
+        }
+
+        if (contentTypes.size() == 1
+                && Iterables.getOnlyElement(contentTypes, "").startsWith(CborDelegateEncoder.MIME_TYPE)) {
+            InputStream inputStream = response.body().asInputStream();
+            if (!inputStream.markSupported()) {
+                inputStream = new BufferedInputStream(inputStream, 1);
```
uschi2000: buffer size 1, why?

bavardage: I copied the approach used by `feign.jackson.JacksonDecoder` - it does the buffering so that it can mark/try to read a byte and avoid the `"No content to map due to end-of-input"`

uschi2000: hmm. i don't know why it should buffer a byte at a time; that seems to have quite a bit of overhead, no?

uschi2000 (commented):
question about the 1-byte buffer, otherwise I'm fine with this.

jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
bavardage: yep, agreed it's sorta fishy

http://stackoverflow.com/questions/1524299/how-can-i-check-if-an-inputstream-is-empty-without-reading-from-it ... the pushback input stream :P?

bavardage: that said, all of our existing json serde goes through the same codepath...

uschi2000: yeah, but that code path might not necessarily great ;)

uschi2000: @adriancole Can you explain why it's cool to make a 1-byte-buffered stream in Feign's JacksonDecoder?

https://github.com/OpenFeign/feign/blob/master/jackson/src/main/java/feign/jackson/JacksonDecoder.java#L56

codefromthecrypt:
Heh probably git blame will say, but I believe some http clients return an
empty stream and maybe jackson wasnt having it?

On 14 Apr 2017 08:01, "Robert Fink" <notifications@github.com> wrote:

*@uschi2000* commented on this pull request.
------------------------------

In jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
<https://github.com/palantir/http-remoting/pull/401#discussion_r111510224>:

> +        this.delegate = delegate;
+    }
+
+    @Override
+    public Object decode(Response response, Type type) throws
IOException, DecodeException, FeignException {
+        Collection<String> contentTypes =
+
HeaderAccessUtils.caseInsensitiveGet(response.headers(),
HttpHeaders.CONTENT_TYPE);
+        if (contentTypes == null) {
+            contentTypes = ImmutableSet.of();
+        }
+
+        if (contentTypes.size() == 1
+                && Iterables.getOnlyElement(contentTypes,
"").startsWith(CborDelegateEncoder.MIME_TYPE)) {
+            InputStream inputStream = response.body().asInputStream();
+            if (!inputStream.markSupported()) {
+                inputStream = new BufferedInputStream(inputStream, 1);

@adriancole <https://github.com/adriancole> Can you explain why it's cool
to make a 1-byte-buffered stream in Feign's JacksonDecoder?

https://github.com/OpenFeign/feign/blob/master/jackson/src/
main/java/feign/jackson/JacksonDecoder.java#L56

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/palantir/http-remoting/pull/401#discussion_r111510224>,
or mute the thread

bavardage:
@uschi2000 @markelliot ping

also, am happy to build our own jackson converter thing which doesn't shove everything through the 1-byte buffered reader as FLUP (though actually, could well be that okhttp gives us markable streams in which case we never hit that code path)

jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
```diff
@@ -0,0 +1,60 @@
...
+import com.palantir.remoting2.jaxrs.feignimpl.HeaderAccessUtils;
+import feign.codec.DecodeException;
+import feign.codec.Decoder;
+import java.io.IOException;
+import java.io.PushbackInputStream;
+import java.lang.reflect.Type;
+import java.util.Collection;
+
+
+public final class CborDelegateDecoder implements Decoder {
```
uschi2000: this is a bit silly since we know the content-type at compile-time. ideally, we'll eventually conjure-generate code that has the json/cbor switching built-in.

can you leave a comment to this effect?

uschi2000 (approved):

uschi2000:
needs rebase and a few docs on the key classes, see above. otherwise LGTM.

jaxrs-clients/src/main/java/feign/CborDelegateDecoder.java
bavardage: in the case of the decoder actually, we don't know at compile time... 

it's valid for a single endpoint on the server to return responses with different content types - your only way of telling is to look at the `Content-Type` header in the response

bavardage: but yes to the code-gen approach over the mess of decoders/delegation

bavardage: will leave a comment on the encoder though (there we *do* know at compile time) + the sillyness in the retrofit factory

bavardage: meh, ok, so I guess there are times when we could indeed determine this entirely at compile time - if the `Accepts` header we send only contains cbor...

bavardage:
ping?

uschi2000:
merge conflict?

bavardage:
sorry, it doesn't show up in the view I have

image

I will rebase

PR #408, opened 2017-04-19
Return all causes and suppressed exceptions

### LINKED ISSUE #407: Exception causes and suppressed exceptions aren't returned in SerializableError
Exception causes and suppressed exceptions aren't being returned. This makes it very difficult to use the returned stack traces because you end up with something like:
```
UncheckedIOException: An error occurred deleting a folder
   <stack trace here>
```
instead of:
```
UncheckedIOException: An error occurred deleting a folder
   <stack trace here>
Caused by: IOException: Folder was not empty
   <stack trace here>
```

SerialVelocity:
Fixes #407

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/408)

### DIFF
--- a/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
+++ b/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
@@ -61,9 +61,11 @@ public abstract class SerializableError implements Serializable {
     @Nullable
     public abstract List<SerializableStackTraceElement> getStackTrace();
 
+    /** Refer to {@link Throwable#getCause()}. */
     @Nullable
     public abstract SerializableError getCause();
 
+    /** Refer to {@link Throwable#getSuppressed()}. */
     public abstract List<SerializableError> getSuppressed();
 
     /** Constructs a new error whose error name is the fully-qualified name of the given class. */
--- a/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
+++ b/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
@@ -26,6 +26,7 @@ import com.google.common.collect.Lists;
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
@@ -101,7 +102,7 @@ public abstract class SerializableError implements Serializable {
                 .collect(Collectors.toList());
 
         return ImmutableSerializableError.builder()
-                .message(throwable.getMessage())
+                .message(Objects.toString(throwable.getMessage()))
                 .errorName(throwable.getClass().getName())
                 .stackTrace(stackTrace)
                 .suppressed(suppressed)
--- a/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
+++ b/error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
@@ -24,8 +24,10 @@ import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.immutables.value.Value;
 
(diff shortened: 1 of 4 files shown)


error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
```diff
@@ -58,6 +61,11 @@
     @Nullable
     public abstract List<SerializableStackTraceElement> getStackTrace();
 
+    @Nullable
```
uschi2000: needs docs (just refer to `Throwable#getCause`)

error-handling/src/main/java/com/palantir/remoting2/errors/SerializableError.java
```diff
@@ -58,6 +61,11 @@
...
+    public abstract SerializableError getCause();
+
+    public abstract List<SerializableError> getSuppressed();
```
uschi2000: as above

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
```diff
@@ -73,6 +74,37 @@ public void testSpecificException() {
     }
 
     @Test
+    public void testExceptionWithoutMessage() {
+        Exception exception = encodeAndDecode(new NullPointerException());
+        assertThat(exception, is(instanceOf(RemoteException.class)));
+        assertThat(exception.getMessage(), is("null"));
```
uschi2000, line 80: was that the behavior before your change, too? maybe this is a bug and it should actually be an empty string.

uschi2000:
implementation looks fine to me. just to confirm, we're only adding the cause&suppressed exceptions when the propagate flag is `true`, right?

uschi2000:
@markelliot , concerns?

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
SerialVelocity: This was the behaviour that was in `JsonExceptionMapper` before. I had to move that logic in here.

SerialVelocity:
@uschi2000 Yes, we only add the cause and suppressed exceptions when propagate is `true`. We call a different static constructor when it is `false`. (See `JsonExceptionMapper.java`)

markelliot:
I'm not sure I understand why we need to transmit suppressed exceptions ever. 

Also think we should set expectations that for 3.0 we'll no longer propagate more than the exception name/type.

SerialVelocity:
We would use this for communication from multiple daemons to a server so we want the exceptions to be sent back to the server for easy consumption.

markelliot:
Ben, why do you need suppressed exceptions? These normally wouldn't be emitted to logs.

Also, if you want to return exceptions in the future, I'd recommend you have an `Either` style object as your return type and make the return of an exception explicit to communicate the problem.

SerialVelocity:
I added suppressed exceptions for two reasons:
 - To make it a complete representation of Throwable instead of just a subset
 - So that when we do parallel operations, multiple exceptions can be returned properly

I find having `Either` style objects throughout your codebase makes the code less readable and more complex so I try to avoid them at all costs

markelliot:
I think if you need the error state back (not sure why, we'd recommend you use the log aggregator), then you need to explicitly return it.

Re: suppressed exceptions: generally the suppressed bits are with high confidence unnecessary info, so don't think we should ever return them. I'm unfamiliar with why parallel operations would generate suppressed exceptions.

pkoenig10:
I think including the cause with the exception information is something that is valuable, mainly to get the complete stack trace of an exception.  I agree that we probably don't want to include suppressed exception (my understanding is that they are only generated in try-with-resources blocks anyways). Thoughts on changing this PR just to include the cause exception?

uschi2000:
We're moving away from propagating stacktraces. They will be logged locally and made available through log aggregation infrastructure.
Cf. https://github.com/palantir/http-remoting-api/blob/develop/errors/src/main/java/com/palantir/remoting/api/errors/SerializableError.java

uschi2000:
Further, exception names (including nested/cause exceptions) are not going to be the recommended way to indicate failure.
Cf https://github.com/palantir/http-remoting-api/blob/develop/errors/src/main/java/com/palantir/remoting/api/errors/ErrorType.java

PR #410, opened 2017-04-20
add rate limiter

ellisjoe:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/410)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting2.servers.jersey;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.RateLimiter;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.function.Function;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Response;
+
+/**
+ * A {@link ContainerRequestFilter} that limits the number of requests per second based on a feature {@code T} of the
+ * {@link ContainerRequestContext}. A rate of no more than {@code maxRequestsPerSecond} will be allowed for each unique
+ * feature.
+ */
+public final class RequestRateLimiter<T> implements ContainerRequestFilter {
+
+    private static final int DEFAULT_MAX_FEATURES = 100_000;
+    private final Function<ContainerRequestContext, Optional<T>> requestFeatureFunc;
+    private final LoadingCache<T, RateLimiter> limiters;
+
+    public static <T> RequestRateLimiter<T> create(long maxRequestsPerSecond,
+            Function<ContainerRequestContext, Optional<T>> requestFeatureFunc) {
+        return new RequestRateLimiter<>(maxRequestsPerSecond, DEFAULT_MAX_FEATURES, requestFeatureFunc);
+    }
+
+    @VisibleForTesting
+    RequestRateLimiter(long maxRequestsPerSecond, long maxUniqueFeatures,
+            Function<ContainerRequestContext, Optional<T>> requestFeatureFunc) {
+        this.requestFeatureFunc = requestFeatureFunc;
+        this.limiters = CacheBuilder.newBuilder()
+                .maximumSize(maxUniqueFeatures)
+                .build(new CacheLoader<T, RateLimiter>() {
+                    @Override
+                    public RateLimiter load(T key) throws Exception {
+                        return RateLimiter.create(maxRequestsPerSecond);
+                    }
+                });
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        requestFeatureFunc.apply(requestContext)
+                .ifPresent(feature -> {
+                    if (!limiters.getUnchecked(feature).tryAcquire()) {
+                        requestContext.abortWith(Response.status(429).build());
+                    }
+                });
+    }
+
+}
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
@@ -32,10 +32,14 @@ import javax.ws.rs.core.Response;
  * A {@link ContainerRequestFilter} that limits the number of requests per second based on a feature {@code T} of the
  * {@link ContainerRequestContext}. A rate of no more than {@code maxRequestsPerSecond} will be allowed for each unique
  * feature.
+ * <p>
+ * The features of type {@code T} will be used a cache keys and therefore must be immutable and implement hashCode and
+ * equals properly.
  */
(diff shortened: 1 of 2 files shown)


jersey-servers/src/test/java/com/palantir/remoting2/servers/jersey/RequestRateLimiterTest.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    public void testLimiterRemembersAndAllowsBursts() throws InterruptedException {
+        execute(NO_DELAY, request1);
+
+        Thread.sleep(1_000);
+
+        List<ContainerRequestContext> requests = Collections.nCopies(10, request1);
+        execute(NO_DELAY, requests);
```
uschi2000, line 92: how is this burst different from the the first test case? don't get it.

jersey-servers/src/test/java/com/palantir/remoting2/servers/jersey/RequestRateLimiterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        execute(NO_DELAY, request1, request2);
+
+        verify(request1, never()).abortWith(any());
+        verify(request2, never()).abortWith(any());
+    }
+
+    @Test
+    public void testUnderLimitAllowed() {
+        List<ContainerRequestContext> requests = Collections.nCopies(10, request1);
+        execute(ALLOWED_DELAY, requests);
```
uschi2000, line 80: these would still be OK with NO_DELAY because they are a burst, no?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                });
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        requestFeatureFunc.apply(requestContext)
+                .ifPresent(feature -> {
+                    if (!limiters.getUnchecked(feature).tryAcquire()) {
+                        requestContext.abortWith(Response.status(429).build());
```
uschi2000, line 72: do we need to set a Retry-After header to control when to retry? (i don't know the answer)

jersey-servers/src/test/java/com/palantir/remoting2/servers/jersey/RequestRateLimiterTest.java
ellisjoe: think of the limiter as adding tokens to a bucket every 1 / REQUESTS_PER_SECOND, so after the first request I need to wait 1 / PPS until making the next request. The test below that you commented on verifies that if I make a request and then wait 1 full second that I can then make up to the maximum number of requests because I've accrued some number of tokens.

jersey-servers/src/test/java/com/palantir/remoting2/servers/jersey/RequestRateLimiterTest.java
ellisjoe: After making some requests and then waiting 1 second, I've earned some amount of bandwidth that I can use in the future.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
ellisjoe: not required, but we can include it: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

uschi2000:
Oh, so you don't start with any money in your bank, I.e. you cannot start
with a burst and then recover. Got it.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.common.util.concurrent.RateLimiter;
+import java.io.IOException;
+import java.util.Optional;
+import java.util.function.Function;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.core.Response;
+
+/**
+ * A {@link ContainerRequestFilter} that limits the number of requests per second based on a feature {@code T} of the
```
schlosna, line 33: Should probably note that the Ts will be used as cache key and thus must be immutable and implement hashCode/equals properly and will be retained for life of limiter

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
```diff
@@ -0,0 +1,71 @@
...
+        return new RequestRateLimiter<>(maxRequestsPerSecond, DEFAULT_MAX_FEATURES, requestFeatureFunc);
+    }
+
+    @VisibleForTesting
+    RequestRateLimiter(long maxRequestsPerSecond, long maxUniqueFeatures,
+            Function<ContainerRequestContext, Optional<T>> requestFeatureFunc) {
+        this.requestFeatureFunc = requestFeatureFunc;
+        this.limiters = CacheBuilder.newBuilder()
+                .maximumSize(maxUniqueFeatures)
+                .build(new CacheLoader<T, RateLimiter>() {
```
schlosna: Not sure exactly how you're planning on using, but do we want to recordStats() on the cache and expose them via metrics?

ellisjoe: initial use case is to rate limit based on `AuthHeader`. since this is a public repo, not sure what the best way to expose metrics is

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
ellisjoe: done

jersey-servers/src/test/java/com/palantir/remoting2/servers/jersey/RequestRateLimiterTest.java
uschi2000: Oh, so you don't start with any money in your bank, I.e. you cannot start
with a burst and then recover. Got it.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
uschi2000: I guess I don't really understand what a client would do upon seeing 429. Ideas?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
schlosna: Not ideal as it would expose Guava In this API, but providing getter for `Cache#stats()` would let us expose some, though most important is probably just cache size which we could expose directly, as we don't really care about hit rates.

That brings up one edge case where you have more than size unique tokens in short period and  cache evicts, you may allow more requests through than desired. Probably

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
ellisjoe: the simplest options are to wait for a period of time and retry or try another server. there's a bunch of other things we can explore later too

uschi2000 (approved):

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
ellisjoe: correct. that's what the `testNoLimitingWhenTooManyUniqueFeatures` verifies. I'd like to pass on dealing with metrics for the moment. don't feel great about exposing guava but can revisit.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RequestRateLimiter.java
```diff
@@ -0,0 +1,75 @@
...
+            Function<ContainerRequestContext, Optional<T>> requestFeatureFunc) {
+        return new RequestRateLimiter<>(maxRequestsPerSecond, DEFAULT_MAX_FEATURES, requestFeatureFunc);
+    }
+
+    @VisibleForTesting
+    RequestRateLimiter(long maxRequestsPerSecond, long maxUniqueFeatures,
+            Function<ContainerRequestContext, Optional<T>> requestFeatureFunc) {
+        this.requestFeatureFunc = requestFeatureFunc;
+        this.limiters = CacheBuilder.newBuilder()
+                .maximumSize(maxUniqueFeatures)
```
markelliot, line 57: also need an expiry?

ellisjoe: done

PR #416, opened 2017-05-09
Palantir Discoverable

gm2211:
The only sane way of doing polymorphic deserialization in Jackson and not using reflection is using @Discoverable and @AutoService. However @Discoverable is currently a Dropwizard-specific interface.

This PR introduces 2 benefits:

1. A Palantir Discoverable interface, so that we don't need to introduce unnecessary dependencies on Dropwizard just to get polymorphic deserialization

2. A more thorough and extensible subtype resolver: currently Dropwizard's implementation of Jackson's subtype resolver only looks for direct subclasses of @Discoverable within the `META-INF/services` folder. The proposed implementation here, instead, supports arbitrary hierarchy depth: this still relies entirely on information found in the `META-INF/services` folder and does not use any reflection. 

So both of the following would work:

```
@AutoService(Discoverable.class)
interface A extends Discoverable { }

@AutoService(A.class)
interface B extends A { }
```

and

```
@AutoService(Discoverable.class)
interface A extends Discoverable { }

@AutoService(Discoverable.class)
interface B extends A { }
```

In the first example, the contents of `META-INF/services` would look like:

`META-INF/services/com.palantir.remoting2.ext.jackson.discoverable.Discoverable` => `com.palantir.test.A`
`META-INF/services/com.palantir.test.A` => `com.palantir.test.B`

while in the second example you'd have:

`META-INF/services/com.palantir.remoting2.ext.jackson.discoverable.Discoverable` => 
`com.palantir.test.A`
`com.palantir.test.B`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/416)

### DIFF
--- a/error-handling/versions.lock
+++ b/error-handling/versions.lock
@@ -73,7 +73,10 @@
             "locked": "2.0.1"
         },
         "org.slf4j:slf4j-api": {
-            "locked": "1.7.12"
+            "locked": "1.7.12",
+            "transitive": [
+                "com.palantir.remoting2:jackson-support"
+            ]
         }
     },
     "runtime": {
@@ -150,7 +153,10 @@
             "locked": "2.0.1"
         },
         "org.slf4j:slf4j-api": {
-            "locked": "1.7.12"
+            "locked": "1.7.12",
+            "transitive": [
+                "com.palantir.remoting2:jackson-support"
+            ]
         }
     }
 }
--- a/extras/jackson-support/build.gradle
+++ b/extras/jackson-support/build.gradle
@@ -11,4 +11,5 @@ dependencies {
 
     testCompile "junit:junit"
     testCompile "org.assertj:assertj-core"
+    testCompile "org.mockito:mockito-core"
 }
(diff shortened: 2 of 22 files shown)


uschi2000:
I am not convinced we really want this. While I appreciate that there are circumstances where polymorphic appears useful, I believe it's a can of worms, in particular when it straddles language boundaries.

Gulio, what's the concrete use-case you have in mind?

gm2211:
When objects that have a hierarchy are persisted in an external store (e.g. atalasdb) which is accessed via network, you need an object mapper that can deal with polymorphic types. This happens already today, it just uses `@JsonSubTypes`, which I think is much less clean than `@Discoverable` + `@Autoservice`. 

This PR does not introduce support for polymorphic serialization, nor it encourages it - it just allows for it to be done cleaner, were devs to stick to it, for whatever reasons.

As a side note, while I can see how sometimes union types can be preferred to inheritance when dealing with cross-language interactions (although they can lead to a fair share of code duplication in Java), there are cases in which either for legacy reasons or dev preferences, inheritance will be used. In those cases I still believe it's better to encourage not having implementations listed on the interface

uschi2000:
this has been stale for a while, closing for now.

uschi2000:
let us know if you feel this is still relevant.

PR #417, opened 2017-05-09
customizable service exceptions

nziebart:
Taking a stab at implementing https://github.com/palantir/http-remoting/issues/243

Goals:
- Allow servers to customize the error returned to clients
- Enable support for safely logging exception messages (currently they are just redacted)

Concerns:
- I'm not sure if `SafeParam` / `UnsafeParam` belong here. An alternative would be to subclass `ServiceException` in the logging library, and enforce the use of params there.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/417)


jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,60 @@
...
+import com.palantir.remoting2.ext.jackson.ObjectMappers;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class ServiceExceptionMapper implements ExceptionMapper<AbstractServiceException> {
+
+    private static final Logger log = LoggerFactory.getLogger(JsonExceptionMapper.class);
```
schlosna: should this be logging to `ServiceExceptionMapper.class`?

error-handling/src/main/java/com/palantir/remoting2/errors/AbstractServiceException.java
```diff
@@ -0,0 +1,49 @@
...
+public abstract class AbstractServiceException extends RuntimeException {
+
+    private final String errorId = UUID.randomUUID().toString();
+
+    protected AbstractServiceException(String message, @Nullable Throwable cause) {
+        super(message, cause);
+    }
+
+    /** Logs this exception. This method is called by the corresponding {@link ExceptionMapper}. */
+    public abstract void logTo(Logger log);
```
markelliot: Per the discussion on the thread, I don't think you should be taking a logger. Instead, let's just make the constructor of this:

    AbstractServiceException(String message, Param... params)

nziebart: done

error-handling/src/main/java/com/palantir/remoting2/errors/Param.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.remoting2.errors;
+
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** A wrapper around an argument used to build a formatted message. */
+public abstract class Param<T> {
```
schlosna: Does this need to implement Serializable as we're embedding these in exceptions which inherently are serializable? One example might be an Either returned via Spark.

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -0,0 +1,89 @@
...
+        return status;
+    }
+
+    /** The format of the exception message. */
+    public final String getMessageFormat() {
+        return messageFormat;
+    }
+
+    /** The parameters for the exception message. */
+    public final Param<?>[] getMessageParams() {
```
schlosna: Would prefer returning an unmodifiable collection of Param and not to expose raw array, otherwise likely need to clone here.

nziebart: done

error-handling/src/main/java/com/palantir/remoting2/errors/Param.java
nziebart: yeah, think this makes sense. added the annotation

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -0,0 +1,91 @@
...
+    /** A unique identifier for this error. */
+    public final String getErrorId() {
+        return errorId;
+    }
+
+    /**
+     * The error that should be returned to the remote client. Subclasses may override this method to return custom
+     * errors.
+     */
+    @SuppressWarnings("checkstyle:designforextension")
```
markelliot, line 71: don't think you should need to suppress this

error-handling/src/main/java/com/palantir/remoting2/errors/Param.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package com.palantir.remoting2.errors;
+
+import java.io.Serializable;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** A wrapper around an argument used to build a formatted message. */
+public abstract class Param<T> implements Serializable {
```
markelliot: I think we're going to want to pull out Param, SafeParam and UnsafeParam to a separate repo

nziebart: Yep, this feels like the right move. Happy to set this up.

I think as part of this we should consider renaming to `SafeArg` / `UnsafeArg` (and I'd propose we call the repo logging-args).

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -74,18 +81,34 @@ public final int getStatus() {
...
+    protected static final String formatMessage(String message, Arg<?>... args) {
+        if (args.length == 0) {
+            return message;
+        }
+
+        String[] stringifiedArgs = Arrays.stream(args)
+                .map(arg -> String.format("%s=%s", arg.getName(), toJson(arg.getValue())))
+                .toArray(String[]::new);
+
+        return String.format("%s {%s}", message, String.join(", ", stringifiedArgs));
```
schlosna: I'd highly encourage writing some JMH benchmarks for this method and making sure that it is as efficient as possible because as written this will allocate a ton of garbage with streams, joiner, and `String.format`, at a time when the service is throwing exceptions, so I'm concerned about cascading failures.

schlosna (changes requested):

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+        try {
+            SerializableError error = exception.getError();
+            builder.type(MediaType.APPLICATION_JSON);
+            String json = MAPPER.writeValueAsString(error);
+            builder.entity(json);
+        } catch (RuntimeException | JsonProcessingException e) {
+            log.warn("Unable to translate exception to json for request {}", exception.getErrorId(), e);
+
+            builder = Response.status(status);
```
uschi2000: flow is a little weird here. maybe build a separate Response in try and catch

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -0,0 +1,114 @@
...
+    public final String getErrorId() {
+        return errorId;
+    }
+
+    /**
+     * The error that should be returned to the remote client. Subclasses may override this method to return custom
+     * errors.
+     */
+    @SuppressWarnings("checkstyle:designforextension")
+    public final SerializableError getError() {
```
uschi2000: would prefer to make the conversion from exception to SerializableError symmetric in ServiceExceptionMapper and JsonExceptionMapper

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -0,0 +1,114 @@
...
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.remoting2.ext.jackson.ObjectMappers;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import javax.ws.rs.core.Response;
+
+public class ServiceException extends RuntimeException implements SafeLoggable {
```
uschi2000, line 30: docs

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+final class ServiceExceptionMapper implements ExceptionMapper<ServiceException> {
+
+    private static final Logger log = LoggerFactory.getLogger(ServiceExceptionMapper.class);
+
+    private static final ObjectMapper MAPPER = ObjectMappers.newClientObjectMapper()
+            .enable(SerializationFeature.INDENT_OUTPUT);
+
+    @Override
+    public Response toResponse(ServiceException exception) {
+        log.info("Error handling request {}", exception.getErrorId(), exception);
```
uschi2000: JsonExceptionMapper logs as `.error`, why is this different?

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
```diff
@@ -0,0 +1,114 @@
...
+        this(DEFAULT_STATUS, cause, messageFormat,  messageParams);
+    }
+
+    public ServiceException(int status, String messageFormat, Arg<?>... messageParams) {
+        this(status, null, messageFormat,  messageParams);
+    }
+
+    public ServiceException(int status, @Nullable Throwable cause, String message,
+            Arg<?>... args) {
+        super(formatMessage(message, args), cause);
```
uschi2000, line 55: should test that `.toString()` does the right thing

error-handling/src/test/java/com/palantir/remoting2/errors/ServiceExceptionTests.java
```diff
@@ -0,0 +1,126 @@
...
+
+        ServiceException ex = new ServiceException(message, args);
+
+        assertThat(ex.getMessage()).isEqualTo(expectedMessage);
+    }
+
+    @Test
+    public void testExceptionMessageWithNoArgs() {
+        String message = "a message";
+
```
uschi2000: nit: remove `\n`s here and above

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+
+final class ServiceExceptionMapper implements ExceptionMapper<ServiceException> {
+
+    private static final Logger log = LoggerFactory.getLogger(ServiceExceptionMapper.class);
+
+    private static final ObjectMapper MAPPER = ObjectMappers.newClientObjectMapper()
+            .enable(SerializationFeature.INDENT_OUTPUT);
+
+    @Override
+    public Response toResponse(ServiceException exception) {
```
uschi2000, line 39: I don't understand how this interacts with the safe-logging infrastructure.

nziebart: ServiceException implements `SafeLoggable`, which is recognized by sls-logging. So, we just log the exception as normal, and the logger will correctly log the message and arguments within the stacktrace (as opposed to current behavior, where it redacts the entire exception message)

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
nziebart: the goal here is to allow subclasses to override this method, as a way to customize the message returned to the caller (https://github.com/palantir/http-remoting/issues/243)

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
nziebart: we have tests that `Exception#getMessage` does the right thing (i.e., that `formatMessage` is correct) - is that what you mean here?

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
nziebart: need to suppress to allow this method to be non-final

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
nziebart: my initial thought was that we didn't care so much about perf when throwing, but yeah i guess in scenarios where we are maxing out resources, more allocations aren't good.

i changed this to use a StringBuilder

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
markelliot: why bother with doing a formatMessage at all?

error-handling/src/main/java/com/palantir/remoting2/errors/ServiceException.java
markelliot: I guess it's not obvious to me why we need to enable returning custom `SerializableError` messages

markelliot: Perhaps a better pattern would be that this is the only implementation of this object and that the constructor take a `SerializableError` as an alternative to use; then various other impls can be factories for this one class instead.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,61 @@
...
+        log.error("Error handling request {}", exception.getErrorId(), exception);
```
diogoholanda, line 40: Should all ServiceExceptions be logged at error? Options discussed were to either specify the log level in the ServiceException itself or log it in info.

uschi2000:
Think this should replace/consolidate RemoteException in 3.0.0. Let's not merge this into 2.x.

ash211:
@nziebart is this PR still active?  I'm trying to debug a service solely through a UI on the safe view of its logs, and not being able to see exception messages from stacktraces is the biggest gap.

uschi2000:
I'll merge a variation of this into the remoting 3 branch.

PR #418, opened 2017-05-09
add enableScala option

hturki:
hey @uschi2000 this is what I was talking about re: the scala support a week or two back

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/418)


markelliot:
low confidence we want to support this; do you really *have* to use Scala objects in your API?

hturki:
per discussion will move to support to a separate jaxrs-scala-clients project

jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+import feign.okhttp.OkHttpClient;
+import feign.slf4j.Slf4jLogger;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import okhttp3.ConnectionPool;
+import okhttp3.ConnectionSpec;
+import okhttp3.Credentials;
+import okhttp3.TlsVersion;
+
+public abstract class AbstractFeignJaxRsClientBuilder extends ClientBuilder {
```
uschi2000: would prefer to not make this public

uschi2000: (I don't want people to make their own subclasses outside this project.)

jaxrs-scala-clients/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+dependencies {
+    compile project(":jaxrs-clients")
+    compile "com.fasterxml.jackson.module:jackson-module-scala_2.11"
+
+    // Keep in sync with the dependencies excluded from shadowJar.
+    shadow project(":jaxrs-clients")
+    shadow "com.fasterxml.jackson.module:jackson-module-scala_2.11"
+}
+
+shadowJar {
```
uschi2000: :( this is pretty unmaintainable already, but duplicating this logic into two places makes it worse. can we at least factor out?

uschi2000 (commented):
I'm not convinced this is the right decomp. Maybe it's easier to add a flag to JaxRsClientBuilder? let's discuss in person, ja?

hturki:
@uschi2000 I actually initially wrote this via the flag approach :) but @markelliot raised some concerns around making the default client fatter / adding a dependency on the scala library for example. but yeah let's just catch up next week in a higher bandwidth medium

jaxrs-scala-clients/build.gradle
hturki: yeah, open to ideas on the cleanest way of going about this

jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private final ClientConfig config;
+
+    AbstractFeignJaxRsClientBuilder(ClientConfig config) {
+        this.config = config;
+        Preconditions.checkArgument(config.maxNumRetries() == 0,
+                "Connection-level retries are not supported by %s", JaxRsClient.class.getSimpleName());
+    }
+
+    protected abstract ObjectMapper getObjectMapper();
```
uschi2000, line 75: this unfortunately makes it very very easy for anyone outside this library to fuck with the object mapper. there's probably no way around this :( maybe service loading for adding the scala support at runtime? at the minimum, please add a class-level comment saying, that nobody outside this library must implement this class. thoughts, @markelliot ?

uschi2000:
shading logics looks OK.

jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/AbstractFeignJaxRsClientBuilder.java
uschi2000: ping @markelliot

uschi2000:
Closing in favour of https://github.com/palantir/http-remoting/pull/439

PR #423, opened 2017-05-11
add RandomSampler

gracew:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/423)

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/RandomSampler.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/RandomSampler.java
@@ -18,17 +18,21 @@ package com.palantir.remoting2.tracing;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
-import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
 
+/**
+ * A sampler that returns true at a rate approximately equivalent to the one specified. The specified rate should be
+ * between 0 and 1 (inclusive).
+ */
 public final class RandomSampler implements TraceSampler {
 
+    private static final ThreadLocalRandom random = ThreadLocalRandom.current();
+
     private final float rate;
-    private final Random random;
 
     public RandomSampler(float rate) {
-        checkArgument(rate > 0 && rate < 1, "rate should be between 0 and 1: was %s", rate);
+        checkArgument(rate >= 0 && rate <= 1, "rate should be between 0 and 1: was %s", rate);
         this.rate = rate;
-        this.random = new Random();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


tracing/src/main/java/com/palantir/remoting2/tracing/RandomSampler.java
```diff
@@ -0,0 +1,39 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.Random;
+
+public final class RandomSampler implements TraceSampler {
+
+    private final float rate;
+    private final Random random;
+
+    public RandomSampler(float rate) {
```
markelliot: should add docs

tracing/src/main/java/com/palantir/remoting2/tracing/RandomSampler.java
```diff
@@ -0,0 +1,39 @@
...
+        checkArgument(rate > 0 && rate < 1, "rate should be between 0 and 1: was %s", rate);
+        this.rate = rate;
+        this.random = new Random();
```
markelliot: should probably use a ThreadLocalRandom

markelliot: ...and just make this a static

uschi2000 (approved):

tracing/src/main/java/com/palantir/remoting2/tracing/RandomSampler.java
```diff
@@ -0,0 +1,43 @@
...
+    private final float rate;
+
+    public RandomSampler(float rate) {
+        checkArgument(rate >= 0 && rate <= 1, "rate should be between 0 and 1: was %s", rate);
+        this.rate = rate;
+    }
+
+    @Override
+    public boolean sample() {
+        return random.nextFloat() < rate;
```
schlosna, line 40: I don't think we want this in a static field, as it effectively serializes all threads through a single random that was set when this class is loaded, see https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html

Instead we need to call current each time to get the appropriate thread local random for the current thread: 

`return ThreadLocalRandom.current().nextFloat() < rate;`

gracew: https://github.com/palantir/http-remoting/pull/426

schlosna: Thanks @gracew

PR #429, opened 2017-05-16
support adding metadata to spans at completion

avisbal:
@uschi2000 / @schlosna - here is one approach of how we can add userId metadata to zipkin trace logs so we can filter on them in the Zipkin UI.

#### Proposed changes
 - `Span` now has a `Map<String,String>` field that is used to store contextual information on spans via zipkin `BinaryAnnotation`s
 - `TraceEnrichingFilter` stores userId metadata on the spans it creates

#### Followup changes
 - Add configuration to enable/disable trace logging on a per user basis

This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/429)


tracing/src/main/java/com/palantir/remoting2/tracing/AsyncSlf4jSpanObserver.java
```diff
@@ -95,6 +98,13 @@ static ZipkinCompatSpan fromSpan(Span span, ZipkinCompatEndpoint endpoint) {
             return annotations;
         }
 
+        private static Iterable<? extends ZipkinCompatBinaryAnnotation> spanMetadataToZipkinBinaryAnnotations(
+                Span span, ZipkinCompatEndpoint endpoint) {
+            return span.getMetadata().entrySet().stream()
+                    .map(entry -> ZipkinCompatBinaryAnnotation.of(entry.getKey(), entry.getValue(), endpoint))
+                    .collect(Collectors.toList());
```
schlosna: I'm a little worried about streams on this code path as it allocates a fair bit for what will typically be a single element.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -91,4 +98,24 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
     }
+
+    private static void addUserIdMetadataToSpan(ContainerRequestContext requestContext) {
+        Optional<String> userId = getUserIdFromAuthHeader(requestContext);
+        if (userId.isPresent()) {
+            Tracer.addSpanMetadata(ImmutableMap.of(USER_ID_METADATA_KEY, userId.get()));
+        }
+    }
+
+    private static Optional<String> getUserIdFromAuthHeader(ContainerRequestContext requestContext) {
```
markelliot: we're already doing this processing to populate the MDC and the request context, probably ought to revisit how the filters end up populating these various threadlocals. The processing isn't free and these are per-request overheads.

avisbal: @markelliot - could you point me to where else we are doing this? I don't see auth-header related filters in http-remoting. Do you mean we already do this in the filters added from the multipass dropwizard bundle (e.g. `UserIdFilter`)?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -91,4 +98,24 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
+        String authHeaderString = requestContext.getHeaderString(HttpHeaders.AUTHORIZATION);
+        if (authHeaderString == null) {
+            return Optional.empty();
+        } else {
+            AuthHeader authHeader = AuthHeader.valueOf(authHeaderString);
+            BearerToken bearerToken = authHeader.getBearerToken();
+            UnverifiedJsonWebToken token = UnverifiedJsonWebToken.of(bearerToken);
```
schlosna: What happens if this method throws due to invalid bearer token? I assume we should `return Optional.empty()`. Can we add tests that cover this case?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
schlosna: @avisbal see:
* https://github.com/palantir/auth-tokens/blob/develop/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java#L58
* https://github.com/palantir/auth-tokens/blob/develop/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java#L43

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -17,24 +17,30 @@
...
 import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
 import org.slf4j.MDC;
 
 @Provider
 public final class TraceEnrichingFilter implements ContainerRequestFilter, ContainerResponseFilter {
     public static final TraceEnrichingFilter INSTANCE = new TraceEnrichingFilter();
+    public static final String USER_ID_METADATA_KEY = "userId";
```
schlosna: Should probably indicate this is unverified user ID.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -91,4 +98,24 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
             return Optional.of(header.equals("1"));
         }
     }
+
+    private static void addUserIdMetadataToSpan(ContainerRequestContext requestContext) {
+        Optional<String> userId = getUserIdFromAuthHeader(requestContext);
+        if (userId.isPresent()) {
+            Tracer.addSpanMetadata(ImmutableMap.of(USER_ID_METADATA_KEY, userId.get()));
```
schlosna: Do we need a way to mark metadata as safe to log?

tracing/src/main/java/com/palantir/remoting2/tracing/AsyncSlf4jSpanObserver.java
avisbal: changed to loop. I didn't instantiate List with fixed size though since it's unknown.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
avisbal: we no longer parse the token, so this is not applicable

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
avisbal: thanks @schlosna , I'm pulling the userId from the request context now

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
avisbal: no longer using this field and instead using key directly from `BearerTokenLoggingFilter`

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -91,4 +94,23 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
+        Optional<String> userId = tryGetUnverifiedUserId(requestContext);
+        if (userId.isPresent()) {
+            Tracer.addSpanMetadata(ImmutableMap.of(BearerTokenLoggingFilter.USER_ID_KEY, userId.get()));
+        }
+    }
+
+    private static Optional<String> tryGetUnverifiedUserId(ContainerRequestContext requestContext) {
+        String userIdKey = BearerTokenLoggingFilter.getRequestPropertyKey(BearerTokenLoggingFilter.USER_ID_KEY);
+        Object maybeUserId = requestContext.getProperty(userIdKey);
+        if (maybeUserId != null && maybeUserId instanceof String) {
```
schlosna: nit: this can be simplified to `if (maybeUserId instanceof String) {` as `null instanceof String` is always false

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -91,4 +94,23 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
+    private static void addUserIdMetadataToSpan(ContainerRequestContext requestContext) {
+        Optional<String> userId = tryGetUnverifiedUserId(requestContext);
+        if (userId.isPresent()) {
+            Tracer.addSpanMetadata(ImmutableMap.of(BearerTokenLoggingFilter.USER_ID_KEY, userId.get()));
+        }
+    }
+
+    private static Optional<String> tryGetUnverifiedUserId(ContainerRequestContext requestContext) {
+        String userIdKey = BearerTokenLoggingFilter.getRequestPropertyKey(BearerTokenLoggingFilter.USER_ID_KEY);
+        Object maybeUserId = requestContext.getProperty(userIdKey);
```
uschi2000: do we have any guarantee about the ordering of these two filters?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
avisbal: fixed

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
avisbal: not currently, @uschi2000, which is a problem. some approaches I can think of:

1. use `@Priority` annotations on the two filters. That feels ugly though since it creates an implicit relationship between these two different repos.
2. register `BearerTokenLoggingFilter` in `HttpRemotingJerseyFeature` in the order we care about. I would have to think about what happens when services are already independently registering `BearerTokenLoggingFilter`.
3. parse the userId ourselves from the authorization header

any thoughts?

uschi2000: I find the `@Priority` business pretty janky :(
3. is what you had initially, right?

gracew: BearerTokenLoggingFilter already has `@Priority(Priorities.AUTHORIZATION)`. this is lower (i.e. higher priority) than the default value of Priorities.USER, so I think we're ok here.

avisbal: @uschi2000 - yeah (3) was the first approach, ruled out for duplicate processing
@gracew - I met with Rob and Dave and it was decided that we didn't want to rely on the `Priority` for now. Instead, this PR now just adds supporting adding metadata on spans. I will make a contour specific filter that composes the `BearerTokenLoggingFilter` and another of lower priority that creates a local span with the userId in the request context.

avisbal:
@schlosna / @uschi2000 - this is ready for review

tracing/src/main/java/com/palantir/remoting2/tracing/OpenSpan.java
```diff
@@ -62,6 +63,12 @@
     public abstract SpanType type();
 
     /**
+     * Returns a map of custom key-value metadata with which spans will be annotated. For example, a "userId" key
+     * could be added to associate spans with the requesting user.
+     */
+    public abstract Map<String, String> getMetadata();
```
uschi2000: don't think we need annotations on open spans?

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -114,6 +114,17 @@ private static OpenSpan startSpanInternal(String operation, SpanType type) {
      * completed span.
      */
     public static Optional<Span> completeSpan() {
+        return completeSpanInternal(Collections.emptyMap());
+    }
+
+    /**
+     * Like {@link #completeSpan()}, but adds {@code metadata} to the current span being completed.
+     */
+    public static Optional<Span> completeSpanWithMetadata(Map<String, String> metadata) {
```
uschi2000: let's overload the method name: `completeSpan(...)`

uschi2000: thoughts on introducing a `SpanMetadata` class?

uschi2000 (changes requested):

tracing/src/main/java/com/palantir/remoting2/tracing/OpenSpan.java
avisbal: done, moved javadoc to `Span`

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
avisbal: done

uschi2000 (approved):

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
avisbal: i don't feel very strongly about it, do you see any specific advantages? `SpanMetadata` provides some additional readability but a `metadata` member on the `Span` class doesn't feel too hard to understand either. It could help future proof us wanting to handle other binary annotation Types but that may be premature.

PR #432, opened 2017-05-22
Shade okhttp and okio in jaxrs-clients

uschi2000:
Required by https://github.com/palantir/http-remoting/issues/431

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/432)

PR #433, opened 2017-05-23
Rfink/upgrade okhttp

uschi2000:
depends on #432

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/433)


uschi2000:
fyi @jamesthomp

versions.props
```diff
@@ -7,8 +7,9 @@ com.google.guava:guava = 18.0
 com.jcraft:jzlib = 1.1.3
 com.netflix.feign:feign-*= 8.17.0
 com.palantir.tokens:auth-tokens = 1.0.0
-com.squareup.okhttp3:* = 3.4.1
+com.squareup.okhttp3:* = 3.8.0
```
schlosna, line 10: I thought we had actually already been using OkHttp 3.6.0 & okio 1.11.0, so the previous 3.4.1 here surprises me, but I may be misremembering as I don't have my laptop right now.

PR #434, opened 2017-05-24
Fix Tracers.wrap to keep MDC up to date

cihati:
Previously, even though `Tracer.getTraceId()` would get propagated correctly across threads when one used `Tracers.wrap`, the generated logs would still not have a `traceId` set.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/434)

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
@@ -21,9 +21,13 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
+import org.slf4j.MDC;
 
 /** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Tracers {
+    /** The key under which trace ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
+    static final String MDC_KEY = "traceId";
+
     private Tracers() {}
 
     /** Returns a random ID suitable for span and trace IDs. */
@@ -96,11 +100,14 @@ public final class Tracers {
         @Override
         public V call() throws Exception {
             Trace originalTrace = Tracer.copyTrace();
+            String originalMdcTraceIdValue = MDC.get(MDC_KEY);
             Tracer.setTrace(trace);
+            MDC.put(MDC_KEY, trace.getTraceId());
             try {
                 return delegate.call();
             } finally {
                 Tracer.setTrace(originalTrace);
+                MDC.put(MDC_KEY, originalMdcTraceIdValue);
             }
         }
     }
@@ -121,11 +128,14 @@ public final class Tracers {
         @Override
         public void run() {
             Trace originalTrace = Tracer.copyTrace();
+            String originalMdcTraceIdValue = MDC.get(MDC_KEY);
             Tracer.setTrace(trace);
+            MDC.put(MDC_KEY, trace.getTraceId());
             try {
                 delegate.run();
             } finally {
                 Tracer.setTrace(originalTrace);
+                MDC.put(MDC_KEY, originalMdcTraceIdValue);
             }
         }
     }
--- a/tracing/src/test/java/com/palantir/remoting2/tracing/TracersTest.java
+++ b/tracing/src/test/java/com/palantir/remoting2/tracing/TracersTest.java
@@ -28,12 +28,14 @@ import java.util.concurrent.TimeUnit;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.MockitoAnnotations;
+import org.slf4j.MDC;
 
 public final class TracersTest {
 
     @Before
     public void before() {
         MockitoAnnotations.initMocks(this);
+        MDC.clear();
     }
 
     @Test
@@ -145,6 +147,7 @@ public final class TracersTest {
             public Void call() throws Exception {
                 assertThat(Tracer.getTraceId()).isEqualTo(expectedTraceId);
                 assertThat(getCurrentFullTrace()).isEqualTo(expectedTrace);
+                assertThat(MDC.get(Tracers.MDC_KEY)).isEqualTo(expectedTraceId);
                 return null;
             }
         };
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/http-remoting](https://github.com/palantir/http-remoting), @cjjojoba! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
```diff
@@ -21,9 +21,13 @@
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadLocalRandom;
+import org.slf4j.MDC;
 
 /** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Tracers {
+    /** The key under which trace ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
+    static final String MDC_KEY = "traceId";
```
cihati, line 29: I copied this from `Tracers.MDC_KEY` -- not sure where to put it to reduce duplication.

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
```diff
@@ -96,11 +100,14 @@ public static Runnable wrap(Runnable delegate) {
         @Override
         public V call() throws Exception {
             Trace originalTrace = Tracer.copyTrace();
+            String originalMdcTraceIdValue = MDC.get(MDC_KEY);
             Tracer.setTrace(trace);
+            MDC.put(MDC_KEY, trace.getTraceId());
```
markelliot, line 105: @uschi2000 we may want to think about a non-MDC mechanism for inclusion of traceId (and userId, sessionId, or others) in logs, thoughts on using our custom layout to do this instead of this approach?

diogoholanda: The solution in this PR seems to be simpler for anyone not using a custom layout, right? Right now you just need to refer to the expected MDC key. What would be the benefits of doing that change?

markelliot: Instead of having to figure out all the places to inject MDC context we could:
 1. use the custom layout
 2. provide a logback extension (there's a standard extension point) to do the extraction

For instance, this doesn't fix that UserId and SessionId will carry into the logging. The problem is broader than traceIds and fixing it piecemeal doesn't seem ideal.

diogoholanda: Any fix would have to be piecemeal right? In all those cases we are going to need:
- A thread local strucuture to hold the information we care about (traceId, userId...)
- A wrapper for ExecutorServices to propagate those thread local structures to their threads

We are just saying that this thread local structure, instead of being the MDC, will be a custom thing for every case. I can see the point of a strongly typed structure being better than the string-keyed MDC though. If you think the logback extension thing is smooth (it would need to depend on all the libraries that produce those thread locals), we could go for that. Otherwise, I don't think it makes the pain much smaller

uschi2000: I'd prefer a solution that limits the amount of stuff we have to propagate as static context. Mark, I don't fully grok your proposal, can you point me to the logback extension point you had in mind?

markelliot: https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/PatternLayout.java

markelliot: Or more specifically https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/pattern/MDCConverter.java

diogoholanda: @uschi2000 @markelliot this PR seems like a bug fix for the current state of affairs. Is there a reason to block it while we figure out if it is worth to change the approach?

uschi2000:
closing in favour of https://github.com/palantir/http-remoting/pull/440

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
gracew: I'm not sure how logback's Converter class could be used to provide a more structured object, as it only has access to the ILoggingEvent. So we could use it to provide more structured access to the MDC, but I'm not seeing how we can avoid the MDC altogether.

gracew: ah, we could rely on thread-local storage like in Tracer. We'd have to do the same for the user/session/token id info too.

gracew: hm, Converter is invoked on the async thread, so you don't have access to the thread-local storage anymore and you're pretty much limited to what's in the ILoggingEvent =/

markelliot:
You also have to modify the prepareForDeferredProcessing method.

On Jun 14, 2017, at 02:09, Grace Wang <notifications@github.com<mailto:notifications@github.com>> wrote:


@gracew commented on this pull request.

In tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_http-2Dremoting_pull_434-23discussion-5Fr121827051&d=DwMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=QA-GsBSRpF9SJ_5U_doz71fp9GvG98qdxjPtZ6P8I6Q&s=emtZI3qoToO0U5gupe_cT5KM5tYbexW2C2OSvDUeg3M&e=>:

>              Tracer.setTrace(trace);
+            MDC.put(MDC_KEY, trace.getTraceId());


hm, Converter is invoked on the async thread, so you don't have access to the thread-local storage anymore and you're pretty much limited to what's in the ILoggingEvent =/

PR #435, opened 2017-05-24
Allow direct proxy configuration

leeavital:
This allows users to bypass the proxy settings set at the JVM  with `-Dhttps.proxyHost`/`-Dhttps.proxyPort`. This can be useful if you are using a third-party library which needs to make proxied http calls, but the library does not allow you to set an external proxy.

Previously okhttp would default to using a direct connection if the proxy failed, but that changed in 3.5.0 (https://github.com/square/okhttp/blob/master/CHANGELOG.md)

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/435)

### DIFF
--- a/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
+++ b/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
@@ -35,6 +35,10 @@ import org.immutables.value.Value.Style;
 @JsonSerialize(as = ImmutableProxyConfiguration.class)
 @Style(visibility = Style.ImplementationVisibility.PACKAGE, builder = "new")
 public abstract class ProxyConfiguration {
+    enum Type {
+        direct, http;
+    }
+
     /**
      * The hostname and port of the HTTP/HTTPS Proxy. Recognized formats include those recognized by {@link
      * com.google.common.net.HostAndPort}, for instance {@code foo.com:80}, {@code 192.168.3.100:8080}, etc.
@@ -46,16 +50,23 @@ public abstract class ProxyConfiguration {
      */
     public abstract Optional<BasicCredentials> credentials();
 
-    public abstract Optional<Boolean> direct();
+    @Value.Default
+    @SuppressWarnings("checkstyle:designforextension")
+    public Type type() {
+        return Type.http;
+    }
 
     @Value.Check
     protected final void check() {
-        Preconditions.checkArgument(direct().isPresent() || hostAndPort().isPresent(), "proxy configuration must either be direct or configured with host-and-port");
-
         if (hostAndPort().isPresent()) {
+            Preconditions.checkArgument(type() == Type.http, "host-and-port only valid for http proxies");
             HostAndPort host = HostAndPort.fromString(hostAndPort().get());
             Preconditions.checkArgument(host.hasPort(),
                     "Given hostname does not contain a port number: " + host);
+        }
+
+        if (credentials().isPresent()) {
+            Preconditions.checkArgument(type() == Type.http, "credentials only valid for http proxies");
         }
     }
 
@@ -63,11 +74,15 @@ public abstract class ProxyConfiguration {
     @SuppressWarnings("checkstyle:designforextension")
     @JsonIgnore
     public Proxy toProxy() {
-        if (hostAndPort().isPresent()) {
-            HostAndPort hostAndPort = HostAndPort.fromString(hostAndPort().get());
-            return new Proxy(Proxy.Type.HTTP, new InetSocketAddress(hostAndPort.getHostText(), hostAndPort.getPort()));
-        } else {
-            return Proxy.NO_PROXY;
+        switch (type()) {
+            case http:
+                HostAndPort hostAndPort = HostAndPort.fromString(hostAndPort().get());
+                InetSocketAddress addr = new InetSocketAddress(hostAndPort.getHostText(), hostAndPort.getPort());
+                return new Proxy(Proxy.Type.HTTP, addr);
+            case direct:
+                return Proxy.NO_PROXY;
+            default:
+                throw new IllegalStateException("unrecognized proxy type; this is a library error");
         }
     }
 
--- a/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
+++ b/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
@@ -94,6 +94,10 @@ public abstract class ProxyConfiguration {
         return new ProxyConfiguration.Builder().hostAndPort(hostAndPort).credentials(credentials).build();
     }
 
+    public static ProxyConfiguration direct() {
+        return new ProxyConfiguration.Builder().type(Type.direct).build();
+    }
+
     // TODO(jnewman): #317 - remove kebab-case methods when Jackson 2.7 is picked up
     static final class Builder extends ImmutableProxyConfiguration.Builder {
 
--- a/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
+++ b/service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
@@ -74,6 +74,9 @@ public abstract class ProxyConfiguration {
      */
     public abstract Optional<BasicCredentials> credentials();
 
+    /**
+     * The type of Proxy.  Defaults to {@link Type#http}.
+     */
     @Value.Default
     @SuppressWarnings("checkstyle:designforextension")
     public Type type() {
@@ -84,6 +87,8 @@ public abstract class ProxyConfiguration {
     protected final void check() {
         switch (type()) {
             case http:
+                Preconditions.checkArgument(maybeHostAndPort().isPresent(), "host-and-port must be "
+                        + "configured for an http proxy");
                 HostAndPort host = HostAndPort.fromString(maybeHostAndPort().get());
                 Preconditions.checkArgument(host.hasPort(),
                         "Given hostname does not contain a port number: " + host);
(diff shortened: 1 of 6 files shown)


service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -35,30 +35,55 @@
...
+    enum Type {
+        direct, http;
+    }
 
     /**
      * The hostname and port of the HTTP/HTTPS Proxy. Recognized formats include those recognized by {@link
      * com.google.common.net.HostAndPort}, for instance {@code foo.com:80}, {@code 192.168.3.100:8080}, etc.
      */
-    public abstract String hostAndPort();
+    public abstract Optional<String> hostAndPort();
```
ash211: this is an API break that would affect consumers

ash211 (commented):
I think going direct instead of using a proxy is specified by passing an absent into the optional proxy configuration higher up in the API, on the ClientConfig.

Is leaving `ClientConfig#proxy()` empty not going direct?

leeavital:
@ash211  ommiting the proxy setting will use `ProxySelect.getDefault().select(uri)`, which will respect settings like `http.proxyHost`.  I'm forced to use those flags because I'm using a third-party library which needs to be proxied, but doesn't let me set the proxy at the library level

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
leeavital: true. I can add a (deprecated) method that returns a non-optional String and throws when the type is direct. That should not break any existing consumers.

markelliot:
If you’re using a third-party library, why are http-remoting settings involved?

leeavital:
The third party library is forcing me to configure proxy settings globally. I need to tell service created by http-remoting to explicitly ignore those global settings.

uschi2000:
I'd like to make this explicit by having users configure one of three
things: NO_PROXY, SYSTEM_PROXY, PROXY. Default is probably SYSTEM_PROXY. We
need to do this in a way that doesn't break existing config.

leeavital:
Ok, could I can change/add to the names of the Type enum so it reads:

```
enum Type {
    noProxy, systemProxy, proxy
}
```

and leave the rest of my PR the same? I don't think that breaks any existing config

leeavital:
Actually, I think my current approach does this exactly what you want. the default is the system proxy, you can configure directly use of a special proxy or a direct proxy

uschi2000:
but the config space is a little weird, for instance i could say proxy=direct and still specify a hostAndIPort.

leeavital:
There are preconditions that prevent you from doing that

uschi2000:
let me summarize what i understand the possible valid configurations are:
```
usesSystemProxy:
  uris: 
    - https://foo

usesExplicitProxy:
  uris: 
    - https://foo
  proxy:
    hostAndPort: myProxy:8080

# same as above
usesExplicitProxy2:
  uris: 
    - https://foo
  proxy:
    type: http
    hostAndPort: myProxy:8080

usesNoProxyEvenIfSystemProxyIsSet:
  uris: 
    - https://foo
  proxy:
    type: direct
```

and these configurations are invalid:
```
invalidDirect:
  uris: 
    - https://foo
  proxy:
    type: direct
    hostAndPort: myProxy:8080
```

there are still undefined cases, e.g., type=http, but no hostAndPort given

leeavital:
You're right, that's a mistake on my part. Is the ask that I fix that (it should trigger a failure) or that I change the configuration further?

leeavital:
@uschi2000 the case you mentioned is handled correctly now

@ash211 I had to do some interesting naming because of erasure issues, but I managed to get it so the existing source API is unbroken

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,105 @@
 @Style(visibility = Style.ImplementationVisibility.PACKAGE, builder = "new")
 public abstract class ProxyConfiguration {
 
+    enum Type {
+
+        /**
+         * Use a direct connection. This option will bypass any JVM-level configured proxy settings.
+         */
+        direct,
```
uschi2000: other enums in this project are UPPER_CASE, please do the same here.

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,105 @@
...
+         */
+        http;
+    }
+
     /**
      * The hostname and port of the HTTP/HTTPS Proxy. Recognized formats include those recognized by {@link
      * com.google.common.net.HostAndPort}, for instance {@code foo.com:80}, {@code 192.168.3.100:8080}, etc.
      */
-    public abstract String hostAndPort();
+    @JsonProperty("hostAndPort")
```
ash211, line 57: these are snake case `host-and-port` usually, but might also need to support `hostAndPort` for backcompat

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,105 @@
 @Style(visibility = Style.ImplementationVisibility.PACKAGE, builder = "new")
 public abstract class ProxyConfiguration {
```
uschi2000, line 37: please update the ServiceConfiguration#proxyConfiguration docs: if absent, uses system proxy configuration.

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,105 @@
...
+        Preconditions.checkState(maybeHostAndPort().isPresent(), "hostAndPort was not configured");
+        return maybeHostAndPort().get();
+    }
 
     /**
      * Credentials if the proxy needs authentication.
      */
     public abstract Optional<BasicCredentials> credentials();
 
+    @Value.Default
```
uschi2000: document default.

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,105 @@
...
+        return Type.http;
+    }
+
     @Value.Check
     protected final void check() {
-        HostAndPort host = HostAndPort.fromString(hostAndPort());
-        Preconditions.checkArgument(host.hasPort(), "Given hostname does not contain a port number: " + host);
+        switch (type()) {
+            case http:
+                HostAndPort host = HostAndPort.fromString(maybeHostAndPort().get());
```
uschi2000: doesn't this NPE when type=http and no host is specified?

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
leeavital: See the builder specified below. This supports deserializing in snake-case and serializing in camel

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
leeavital: These leak into the configuration, which usually doesn't contain upper case values.  Glad to change, but that's why I had made this particular enum lowercase

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
leeavital: Nice catch yes

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
leeavital: do you mean *de*serializing the camel-cased hostAndPort? I don't think that was ever supported because of the annotation `@JsonDeserialize(builder = ProxyConfiguration.Builder.class)`

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
uschi2000: SslConfiguration uses UPPER_CASE, i think. What are example of lower-case configs?

leeavital:
enum is now upper cased

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,110 @@
...
+            case DIRECT:
+                Preconditions.checkArgument(!maybeHostAndPort().isPresent() && !credentials().isPresent(),
+                        "Neither credential nor host-and-port may be configured for DIRECT proxies");
+                break;
+            default:
+                throw new IllegalStateException("Unrecognized case; this is a library bug");
+        }
+
+        if (credentials().isPresent()) {
+            Preconditions.checkArgument(type() == Type.HTTP, "credentials only valid for http proxies");
```
uschi2000, line 105: nit: "for HTTP proxies"

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -36,45 +36,110 @@
...
+    }
+
     @Value.Check
     protected final void check() {
-        HostAndPort host = HostAndPort.fromString(hostAndPort());
-        Preconditions.checkArgument(host.hasPort(), "Given hostname does not contain a port number: " + host);
+        switch (type()) {
+            case HTTP:
+                Preconditions.checkArgument(maybeHostAndPort().isPresent(), "host-and-port must be "
+                        + "configured for an http proxy");
```
uschi2000: HTTP

uschi2000 (approved):
one last nit, sorry. otherwise looks good to me.

uschi2000:
@ash211 anything left regarding the serialization format?

ash211:
@uschi2000 I'm good as long as it's not a backcompat break!  Doesn't look like it, but I'm not familiar with Jackson intricacies

uschi2000 (approved):

PR #436, opened 2017-05-24
Retrofit can return CompletableFutures

j-baker:
Remote calls are currently indistinguishable from local calls, which
makes them hard to use.

Additionally, Retrofit currently sucks to use; the Call api is hard
to use and throws IOExceptions.

This lets it return CompletableFutures, which is better than Feign
and

All of the AsyncCallTracker crap is just there to maintain backcompat;
the current method used to throw a RemoteException is wholly
incompatible with anything async.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/436)


j-baker:
@markelliot is this the right way to copy Apache 2 code from other places? I remember you knowing how to do this a while back.

Also, I'm sorry about the glaring hack, but I'm not sure there's any other way to do it :(

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/DefaultAsyncCallTracker.java
```diff
@@ -0,0 +1,38 @@
...
+public final class DefaultAsyncCallTracker implements AsyncCallTracker {
+    private final Set<UUID> asyncCalls = new HashSet<>();
+
+    @Override
+    public <T> void registerAsyncCall(Call<T> call) {
+        asyncCalls.add((UUID) call.request().tag());
+    }
+
+    @Override
+    public boolean isAsyncRequest(Request request) {
```
uschi2000: can we guarantee that this gets called exactly once per registerAsyncCall ?

retrofit2-clients/src/test/java/com/palantir/remoting2/retrofit2/Retrofit2ClientApiTest.java
```diff
@@ -96,6 +106,38 @@ public void testCborRequests() throws IOException, InterruptedException {
...
+        server.enqueue(new MockResponse()
+                .setResponseCode(500)
+                .setHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON)
+                .setBody(ObjectMappers.newClientObjectMapper().writeValueAsString(error)));
+
+        CompletableFuture<String> future = service.makeFutureRequest();
+
+        try {
+            future.join();
+            fail();
```
uschi2000, line 134: use `assertThatThrownBy` (rest of this class uses assertj, i think)

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/AsyncCallTracker.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting2.retrofit2;
+
+import okhttp3.Request;
+import retrofit2.Call;
+
+public interface AsyncCallTracker {
```
uschi2000: minimal docs, explain why we need this.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
```diff
@@ -0,0 +1,116 @@
...
+                }
+            };
+
+            call.enqueue(new Callback<R>() {
+                @Override public void onResponse(Call<R> call, Response<R> response) {
+                    if (response.isSuccessful()) {
+                        future.complete(response.body());
+                    } else {
+                        Collection<String> contentTypes = response.raw().headers("Content-Type");
+                        InputStream body = response.errorBody().byteStream();
```
uschi2000, line 94: can byteStream be null?

uschi2000 (changes requested):
this doesn't exactly make the retrofit library prettier :(
as an alternative, could we make a static utility that takes a `retrofit2.Call<T>` and turns it into a `CompletableFuture<T>`?

retrofit2-clients/src/test/java/com/palantir/remoting2/retrofit2/Retrofit2ClientApiTest.java
j-baker: you can't assert all that much of the cause, sadly

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
j-baker: Don't think so!

j-baker:
Static utility - no, sadly not. The meat of this is in the request adapter, which is essentially something that takes a call and registers a callback; you can't do better because that's the same interface that Call gives you - in fact, this literally wraps Call.

You could wrap it in something like

```
Call<T> call;
CompletableFuture<T> future = new CompletableFuture<>();
try {
    future.complete(call.execute());
} catch (Exception e) {
    future.completeExceptionally(e);
}
```

but this blocks which subverts most of the purpose of doing this.

retrofit2-clients/src/test/java/com/palantir/remoting2/retrofit2/Retrofit2ClientApiTest.java
j-baker: equals, instance, message. It's ugly, can't think of a better way than this.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/DefaultAsyncCallTracker.java
j-baker: Not without keeping some kind of information about the fact we executed it, which leads to issues.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
j-baker: (errorBody can be, but only if isSuccessful is true)

uschi2000:
for copying code, you can do what the WrappingExecutorService header does, plus add a link:
```
/*
 * Copied from Guava 18.0; original copyright notice below.
 * ... link
 *
 * ... original license ...
 */
```

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/DefaultAsyncCallTracker.java
uschi2000: ok, then i think we should keep these in a cache with timeout.

uschi2000:
i'm not super-sold on the idea of hacking the shit out of retrofit, but i appreciate the future/async is interesting to support. thoughts, @markelliot ?

j-baker:
also - I definitely accept that this doesn't make the retrofit code any better. However, I also think that this entirely transforms the Retrofit interfaces currently writeable;

it turns

```
Call<String> myCall = service.doSomething(args);
try {
    return myCall.execute();
} catch (IOException e) {
    throw new RuntimeException(e);
}
```

into

```
return service.doSomething(args).join();
```

in the most basic case, or in fancier cases, things like

```
return service1.doSomething(args).thenCombine(service2.doSomething(args2), (res1, res2) -> stuff(res1, res2));
```

j-baker:
for example of a real world usage, on the project I've been working on we frequently have to open many streams of data concurrently, and it's really common for the programmer to either just block on them all, or become aware it'd be silly and add an executor service to run their blocking calls (frequently static because no one likes injecting executors all over the place). It means that we have many executors, and frequently our Yourkits just show threads blocking on threads blocking on threads. It'd be really awesome if instead of blocking opening three streams, blocking getting the lengths of each of them, and then blocking reading a small amount of metadata from them, we could just do most of the lifting with the CompletionStage api asynchronously.

j-baker:
@uschi2000 I made it a fair bit less crazy with a touch of mutability. Now it should be basically palatable, no more 'hacking the shit out of retrofit'.

retrofit2-clients/src/main/java/com/palantir/remoting2/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -121,6 +123,13 @@ private OkHttpClient createOkHttpClient(String userAgent, List<String> uris) {
         // increase default connection pool from 5 @ 5 minutes to 100 @ 10 minutes
         client.connectionPool(new ConnectionPool(100, 10, TimeUnit.MINUTES));
 
+        Dispatcher dispatcher = new Dispatcher();
```
uschi2000, line 126: sorry, let's keep these in a separate PR.

uschi2000 (commented):
one last comment, otherwise i'm fine with this.

markelliot:
Before we merge this (and pick up support load for it), one question we should poke at is whether there's a better way to support James' remoting needs. In particular, would 'dialogue' would be a better fit here?

uschi2000:
It probably would be, but it's not ready.

markelliot:
Ok, it’s not obvious to me this is the right way to proceed.

Why can we not wrap synchronous calls with CompletableFutures external to the library, or instead provide the library function that appropriately turns a Call<T> into a CompletableFuture<T>?

uschi2000:
That was my suggestion above.

markelliot:
Hm, actually, why are you unable to use https://square.github.io/retrofit/2.x/retrofit/retrofit2/Call.html#enqueue-retrofit2.Callback- ?

j-baker:
The RemoteException thing breaks Callback.

markelliot:
I don't understand why, sorry.

j-baker:
I documented why here: https://github.com/palantir/http-remoting/pull/436/files#diff-e75d19c49d5d16f39054b6a769305f37R22 it's the same reason

uschi2000:
Can we re-architect the thing so it throws exceptions in synchronous mode, but plays nicely with the `enqueue()` async mode?

uschi2000:
alternatively, stop throwing exceptions if that's how retrofit is meant to be used.

j-baker:
So in terms of wrapping the synchronous calls external to the library - explicitly I'm trying to achieve more parallelism because there are plenty of places where we do 10 remote calls in a row before being able to consume any data - so if I were to do that I'd have to create my own thread pools and pass them in to each usage site. At the moment, the code I'm looking at has a bunch of static executor services, and I'd rather use the retrofit built in thread pool for this.

I'm hooking this in in the way Retrofit hopes people do this - see https://github.com/square/retrofit/blob/master/retrofit-adapters/java8/src/main/java/retrofit2/adapter/java8/Java8CallAdapterFactory.java - it's the idiomatic way to do it, with a pretty straightforward hoop to jump through http-remoting changes.

In terms of just using Calls and callbacks, the call interface is pretty hard to consume for the user (for example, a pain to mock, since you need to create a Response object which is final) and in a remoting world probably exposes too much to the user.

j-baker:
Stopping throwing exceptions - fine to do that, but this is rather little additional code to avoid the breaking change for consumers (and work required to communicate it) - 25 lines of code.

uschi2000:
Spoke offline; I'm fine taking this. The CompletableFuture style is part of the standard retrofit game ( https://github.com/square/retrofit/blob/master/retrofit-adapters/java8/src/main/java/retrofit2/adapter/java8/Java8CallAdapterFactory.java ) and I agree they are more usable than Calls.

uschi2000 (approved):

markelliot:
Ok

PR #458, opened 2017-06-12
Log client-error at INFO level

meditativeape:
This helps reduce noise in service logs. See https://github.com/palantir/http-remoting/issues/447

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/458)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
@@ -53,9 +53,13 @@ abstract class JsonExceptionMapper<T extends Exception> implements ExceptionMapp
     public final Response toResponse(T exception) {
         String exceptionMessage = Objects.toString(exception.getMessage());
         String errorId = UUID.randomUUID().toString();
-        log.error("Error handling request {}", SafeArg.of("errorId", errorId), exception);
-
         StatusType status = this.getStatus(exception);
+        if (status.getFamily().equals(Response.Status.Family.CLIENT_ERROR)) {
+            log.info("Error handling request {}", SafeArg.of("errorId", errorId), exception);
+        } else {
+            log.error("Error handling request {}", SafeArg.of("errorId", errorId), exception);
+        }
+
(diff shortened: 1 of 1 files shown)


jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
```diff
@@ -52,9 +52,13 @@
     public final Response toResponse(T exception) {
         String exceptionMessage = Objects.toString(exception.getMessage());
         String errorId = UUID.randomUUID().toString();
-        log.error("Error handling request {}", errorId, exception);
-
         StatusType status = this.getStatus(exception);
+        if (status.getFamily().equals(Response.Status.Family.CLIENT_ERROR)) {
+            log.info("Error handling request {}", errorId, exception);
+        } else {
+            log.error("Error handling request {}", errorId, exception);
```
schlosna: Is there a reason we don't include the actual status code here as a safe arg? I realize you could get it from the request log, but seems like this could be more useful context here, in addition to the error ID.

gracew: not sure if it is that useful -- the exception will provide a lot more info.

gracew (approved):
lgtm -- @meditativeape can you merge in develop and resolve conflicts now that I've merged https://github.com/palantir/http-remoting/pull/456?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
schlosna: except that the exception message itself is not trusted as safe, while an HTTP response code should be at a minimum

gracew: yeah, I think we probably need to make a more concerted effort internally on adoption of SafeLoggable.

meditativeape:
Rebased. Thanks @gracew !

ash211:
@gracew ready to merge?

PR #459, opened 2017-06-16
In clients, log 4xx errors at higher level than 5xx errors

ash211:
Since 4xx errors are client-side, the client should care more about the issue
than problems that are server-side (5xx).

This is the complement to https://github.com/palantir/http-remoting/pull/458 by @meditativeape which changes logging on the server to put 5xx responses at a higher level than 4xx responses.

I'm unsure if this is the correct exception mapper used in clients rather than in servers.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/459)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
@@ -16,8 +16,6 @@
 
 package com.palantir.remoting2.servers.jersey;
 
-import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
-
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.remoting2.errors.RemoteException;
@@ -44,7 +42,7 @@ final class RemoteExceptionMapper implements ExceptionMapper<RemoteException> {
         Status status = Status.fromStatusCode(exception.getStatus());
 
         // here in the client, log responses that indicate client error (4xx) at higher level than server error (5xx)
-        if (CLIENT_ERROR.equals(status.getFamily())) {
+        if (Status.Family.CLIENT_ERROR.equals(status.getFamily())) {
             log.error("Received response status code {} from server handling request. errorId: {}",
                     SafeArg.of("statusCode", status.getStatusCode()),
                     SafeArg.of("errorId", errorId),
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
@@ -48,7 +48,7 @@ final class RemoteExceptionMapper implements ExceptionMapper<RemoteException> {
                     SafeArg.of("errorId", errorId),
                     exception);
         } else {
-            log.warn("Received response status code {} from server handling request. errorId: {}",
+            log.info("Received response status code {} from server handling request. errorId: {}",
                     SafeArg.of("statusCode", status.getStatusCode()),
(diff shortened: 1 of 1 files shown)


jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -38,10 +40,23 @@
...
         Status status = Status.fromStatusCode(exception.getStatus());
+
+        // here in the client, log responses that indicate client error (4xx) at higher level than server error (5xx)
+        if (CLIENT_ERROR.equals(status.getFamily())) {
+            log.error("Received response status code {} from server handling request. errorId: {}",
+                    SafeArg.of("statusCode", status.getStatusCode()),
+                    SafeArg.of("errorId", errorId),
+                    exception);
+        } else {
+            log.warn("Received response status code {} from server handling request. errorId: {}",
```
uschi2000: do we have warn-level logging enabled by default?

ash211: Yes -- without configuration, generally logging libraries log at INFO level and higher (WARN, ERROR, FATAL), and our dropwizard-esque project has those defaults too.

ash211:
Is the correct exception mapper that's used in clients rather than in servers?

ash211:
@uschi2000 ready for review

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -38,10 +38,23 @@
     @Override
     public Response toResponse(RemoteException exception) {
         String errorId = UUID.randomUUID().toString();
-        log.error("Error handling request {}", SafeArg.of("errorId", errorId), exception);
 
-        SerializableError error = exception.getRemoteException();
         Status status = Status.fromStatusCode(exception.getStatus());
+
+        // here in the client, log responses that indicate client error (4xx) at higher level than server error (5xx)
```
uschi2000, line 44: this is in the server, not in the client

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -38,10 +38,23 @@
...
+        if (Status.Family.CLIENT_ERROR.equals(status.getFamily())) {
+            log.error("Received response status code {} from server handling request. errorId: {}",
+                    SafeArg.of("statusCode", status.getStatusCode()),
+                    SafeArg.of("errorId", errorId),
+                    exception);
+        } else {
+            log.info("Received response status code {} from server handling request. errorId: {}",
```
uschi2000, line 51: should this be INFO or WARN?

ash211: I set to INFO to match https://github.com/palantir/http-remoting/pull/458/files#diff-bca185972ae917cdb714c8c40eacb218R58 but @mahmoudm also suggested WARN

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
uschi2000: well, it's in a server that gets an exception from another server. in that sense it's both in a client and in a server :)

uschi2000 (approved):

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
ash211: Moving to WARN in #462

PR #462, opened 2017-06-20
Log 5xx responses as WARN in client

ash211:
In this call stack:

> client -> server A -> server B

where an exception is thrown in B and propagates back from server B:

> client <- server A <- server B

We log in these ways at each point in the return across server-spanned call stack.

In server B at outgoing time (JsonExceptionMapper):
- 4xx => INFO
- 5xx => ERROR

In server A at incoming time (SerializableErrorToExceptionConverter):
- 4xx => none
- 5xx => none

In server A at outgoing (forwarding) time (RemoteExceptionMapper):
- 4xx => WARN
- 5xx => WARN

In server A we log as WARN as the exceptions signal an error in server B which impacts the availability of A.

Followup to https://github.com/palantir/http-remoting/pull/459

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/462)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
@@ -48,7 +48,7 @@ final class RemoteExceptionMapper implements ExceptionMapper<RemoteException> {
                     SafeArg.of("errorId", errorId),
                     exception);
         } else {
-            log.info("Received response status code {} from server handling request. errorId: {}",
+            log.warn("Received response status code {} from server handling request. errorId: {}",
                     SafeArg.of("statusCode", status.getStatusCode()),
                     SafeArg.of("errorId", errorId),
                     exception);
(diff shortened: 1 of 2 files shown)


jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -30,29 +29,35 @@
...
-                    SafeArg.of("errorId", errorId),
-                    exception);
-        } else {
-            log.info("Received response status code {} from server handling request. errorId: {}",
-                    SafeArg.of("statusCode", status.getStatusCode()),
-                    SafeArg.of("errorId", errorId),
-                    exception);
-        }
+        // log at WARN instead of ERROR because although this indicates an issue in a remote server, it is not
+        log.warn("Forwarding response and status code {} from remote server back to caller",
```
uschi2000: can you leave a comment somewhere in here:
```
// TODO(rfink): Propagate errorId across service boundaries?
```

ash211: added.

it's propagated via the exception message, but not in a structured way right now

uschi2000 (approved):

PR #468, opened 2017-06-27
Add explicit JsonProperty annotation in ProxyConfiguration#type()

leeavital:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/468)


leeavital:
This allows ProxyConfiguration to be parsed when a KebabCaseStrategy is being used.

service-config/src/main/java/com/palantir/remoting2/config/service/ProxyConfiguration.java
```diff
@@ -79,6 +79,7 @@ public String hostAndPort() {
      */
     @Value.Default
     @SuppressWarnings("checkstyle:designforextension")
+    @JsonProperty("type")
```
uschi2000, line 82: weird. i don't quite understand why this is necessary :/

uschi2000 (approved):

PR #469, opened 2017-06-29
Replace errors/ package by http-remoting-api:errors, use RemoteException#errorCode+errorName

uschi2000:
Removes ability to configure stacktrace propagation.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/469)


jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/HttpRemotingJerseyFeature.java
```diff
@@ -22,30 +22,20 @@
...
-
-    private HttpRemotingJerseyFeature(StacktracePropagation stacktracePropagation) {
-        this.propagateStackTraces = stacktracePropagation == StacktracePropagation.PROPAGATE;
-    }
-
-    public static HttpRemotingJerseyFeature with(StacktracePropagation stacktracePropagation) {
-        return new HttpRemotingJerseyFeature(stacktracePropagation);
-    }
+public enum HttpRemotingJerseyFeature implements Feature {
+    INSTANCE;
```
uschi2000, line 26: note: no more stacktrace propagation

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
```diff
@@ -44,7 +44,7 @@
 
     private static final String message = "hello";
     private static final int STATUS_42 = 42;
-    private static final ObjectMapper OBJECT_MAPPER = ObjectMappers.newClientObjectMapper();
+    private static final ObjectMapper OBJECT_MAPPER = ObjectMappers.newServerObjectMapper();
```
gracew: would be more realistic to serialize using the server object mapper and deserialize using the client one

jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/feignimpl/FeignSerializableErrorErrorDecoder.java
```diff
@@ -45,7 +45,6 @@ public Exception decode(String methodKey, Response response) {
         } catch (IOException e) {
             return new RuntimeException("Cannot get input stream from response: " + e.getMessage(), e);
         }
-        return SerializableErrorToExceptionConverter.getException(contentTypes, response.status(), response.reason(),
-                body);
+        return SerializableErrorToExceptionConverter.getException(contentTypes, response.status(), body);
```
gracew, line 48: why stop passing through the reason? seems useful as a more human-readable version of the status code.

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
```diff
@@ -54,95 +54,79 @@ public void testWebApplicationExceptions() {
     }
 
     @Test
+    public void testServiceException() throws Exception {
+        ServiceException originalException =
+                new ServiceException(ErrorType.FAILED_PRECONDITION, SafeArg.of("key", "value"));
+
+        SerializableError error = SerializableError.forException(originalException);
+        String json = OBJECT_MAPPER.writeValueAsString(error);
+        RemoteException exception = (RemoteException) decode(MediaType.APPLICATION_JSON, json, 400);
```
gracew: could use `encodeAndDecode()` here

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
```diff
@@ -54,95 +54,79 @@ public void testWebApplicationExceptions() {
...
 
         RemoteException exception = (RemoteException) encodeAndDecode(originalException);
-        assertThat(exception.getCause(), is(nullValue()));
-        assertThat(exception.getStatus(), is(Status.UNAUTHORIZED.getStatusCode()));
-        assertThat(exception.getRemoteException().getErrorName(), is(NotAuthorizedException.class.getName()));
-        assertThat(exception.getMessage(), is(message));
+        assertThat(exception.getCause()).isNull();
+        assertThat(exception.getStatus()).isEqualTo(Status.UNAUTHORIZED.getStatusCode());
+        assertThat(exception.getError().errorCode()).isEqualTo(NotAuthorizedException.class.getName());
+        assertThat(exception.getError().errorName()).isEqualTo("hello");
```
gracew: think `isEqualTo(message)` is a bit clearer

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
```diff
@@ -69,15 +61,11 @@ public final Response toResponse(T exception) {
...
-                StackTraceElement[] stackTrace = exception.getStackTrace();
-                error = SerializableError.of(exceptionMessage, exception.getClass(),
-                        Arrays.asList(stackTrace));
-            } else {
-                error = SerializableError.of("Refer to the server logs with this errorId: "
-                        + errorId, exception.getClass());
-            }
+            SerializableError error = SerializableError.builder()
+                    .errorCode(exception.getClass().getName())
+                    .errorName("Refer to the server logs with this errorId: " + errorId)
```
gracew: this usage of the API is different than usage by ServiceException. I think for consistency we should set errorName to `exception.getClass().getName()`.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
```diff
@@ -86,7 +74,7 @@ public final Response toResponse(T exception) {
             // simply write out the exception message
             builder = Response.status(status);
             builder.type(MediaType.TEXT_PLAIN);
-            builder.entity(exceptionMessage);
+            builder.entity(Objects.toString(exception.getMessage()));
```
gracew: why call `Objects.toString`? `exception.getMessage()` is already a string

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Serializes {@link ServiceException}s into a JSON representation of a {@link SerializableError}s (see {@link
+ * SerializableError#forException}). Also see {@link JsonExceptionMapper} for more information on the interplay of
+ * Jersey and exception mappers.
+ */
+final class ServiceExceptionMapper implements ExceptionMapper<ServiceException> {
+
+    private static final Logger log = LoggerFactory.getLogger(ServiceExceptionMapper.class);
+
+    static final ObjectMapper MAPPER = ObjectMappers.newClientObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
```
gracew: server object mapper?

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,71 @@
...
+        try {
+            SerializableError error = SerializableError.forException(exception);
+            builder.type(MediaType.APPLICATION_JSON);
+            String json = MAPPER.writeValueAsString(error);
+            builder.entity(json);
+        } catch (RuntimeException | JsonProcessingException e) {
+            log.warn("Unable to translate exception to json for request {}",
+                    SafeArg.of("errorId", exception.getErrorId()), e);
+            // simply write out the exception message
+            builder = Response.status(httpStatus);
```
gracew: this is already done in line 55 above

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
gracew: I don't think we should be propagating the exception message here as a fallback (it might not be safe). instead, we should write out the string `"Refer to the server logs with this errorId: " + errorId`

pnepywoda:
ping @uschi2000 must have ServiceExceptions! :)

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
uschi2000: done

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
uschi2000: done

error-handling/src/test/java/com/palantir/remoting2/errors/SerializableErrorErrorDecoderTests.java
uschi2000: that one's for non-ServiceExceptions (have clarified in code)

jaxrs-clients/src/main/java/com/palantir/remoting2/jaxrs/feignimpl/FeignSerializableErrorErrorDecoder.java
uschi2000: the reason is null anyway in http/2 and can be statically derived from the status.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
uschi2000: This will make current UI experience even worse, I think. @pnepywoda

uschi2000: Chatted with Paul. If we remove the message here, then today's UI's will just show `java.util.IllegalArgumentException`

markelliot:
We should share some kind of error name information as well as the lookup code. The UIs need to map the error names either way.

On Jul 17, 2017, at 18:44, Robert Fink <notifications@github.com<mailto:notifications@github.com>> wrote:


@uschi2000 commented on this pull request.

In jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_http-2Dremoting_pull_469-23discussion-5Fr127866486&d=DwMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=ZCLNH19-lZHm6O9yPebg2dnvewPvommSOZ-4gvPDrU8&s=lRi5avMz0xali_zgzNKhkq3DjvzFEYf8hiJE-v-_iGc&e=>:

> @@ -69,15 +61,11 @@ public final Response toResponse(T exception) {

         ResponseBuilder builder = Response.status(status);
         try {
-            final SerializableError error;
-            if (includeStackTrace) {
-                StackTraceElement[] stackTrace = exception.getStackTrace();
-                error = SerializableError.of(exceptionMessage, exception.getClass(),
-                        Arrays.asList(stackTrace));
-            } else {
-                error = SerializableError.of("Refer to the server logs with this errorId: "
-                        + errorId, exception.getClass());
-            }
+            SerializableError error = SerializableError.builder()
+                    .errorCode(exception.getClass().getName())
+                    .errorName("Refer to the server logs with this errorId: " + errorId)


Chatted with Paul. If we remove the message here, then today's UI's will just show java.util.IllegalArgumentException

uschi2000:
Not clear to me how to do this in a backwards-compatible way. This code here is for products that throw IllegalArgumentException. Consumers of that API today:
- if they are a frontend, just display the message (=errorName)
- if they are a backend, either just propagate the exception forward, or switch on the class (=errorCode)

uschi2000:
so the error name you're asking for here is the fully-qualified class name for these legacy things.

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
uschi2000: done

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
uschi2000: done

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
uschi2000: done

uschi2000:
ready for you, @gracew

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+        ResponseBuilder builder = Response.status(httpStatus);
+        try {
+            SerializableError error = SerializableError.forException(exception);
+            builder.type(MediaType.APPLICATION_JSON);
+            String json = MAPPER.writeValueAsString(error);
+            builder.entity(json);
+        } catch (RuntimeException | JsonProcessingException e) {
+            log.warn("Unable to translate exception to json for request {}",
+                    SafeArg.of("errorId", exception.getErrorId()), e);
+            // simply write out the exception message
```
gracew: update comment

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/JsonExceptionMapper.java
gracew: ok, so this is just for backcompat for generic exceptions

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/ServiceExceptionMapper.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Serializes {@link ServiceException}s into a JSON representation of a {@link SerializableError}s (see {@link
+ * SerializableError#forException}). Also see {@link JsonExceptionMapper} for more information on the interplay of
+ * Jersey and exception mappers.
+ */
+final class ServiceExceptionMapper implements ExceptionMapper<ServiceException> {
+
+    private static final Logger log = LoggerFactory.getLogger(ServiceExceptionMapper.class);
+    private static final ObjectMapper MAPPER =
+            ObjectMappers.newServerObjectMapper().enable(SerializationFeature.INDENT_OUTPUT);
```
gracew, line 42: is the pretty printing intentional?

markelliot:
Not sure I actually understand the change that frontends will get.

gracew (approved):
lgtm

uschi2000:
@markelliot 
The wire-format is fully backwards-compatible, so this change will not *require* them to change. However, if a server throws exceptions in the proposed new mechanism:
```
throw new ServiceException(DATASET_NOT_FOUND, SafeArg.of("datasetRid", datasetRid));
```
, then they will receive something like
```
400 Bad Request
{
  errorCode: INVALID_ARGUMENT
  errorName: DatasetNotFound
  parameters:
    datasetRid: 123abc
}
```
and can dispatch accordingly.

Conversely, a remoting2 server may
```
throw new IllegalArgumentException("Dataset not found: " + datasetRid);
```

The Web client will see this as
```
400 Bad Request
{
  errorCode: java.lang.IllegalArgumentException
  className: java.lang.IllegalArgumentException
  errorName: Refer to the server logs with this errorId: 789xyz
  message: Refer to the server logs with this errorId: 789xyz
}
```

I don't think we can afford a harder break here, unfortunately.

markelliot:
Got it, don’t we still want to populate the message going forward with something similar to the historic message field? (also doesn’t that fix the break?)

uschi2000:
Yep, doing that, but left it out of the example for brevity. It always populates errorCode=exceptionClass and errorName=message. Both are `@Deprecated` in the Java API.

markelliot:
Do we know if frontends are consuming the message or the errorName? If they’re consuming the message, we can simply keep generating it the same way…

uschi2000:
errorCode and errorName don't exist at all in the wild, it's only `exceptionClass` and `message`. I think I'd like to change the wire format eventually to errorCode&errorName. Do you think that's infeasible?

markelliot:
May need an interstitial step where we continue to produce “message” in the old format to give people a chance to migrate.

Think it may be worth having a structured ‘errorId’ that allows collation of a FE error with a backend one, though. We seem to be missing that, now.

gracew:
Rob left it out of his example, but this PR includes the errorId as structured data:
```
{
  errorCode: INVALID_ARGUMENT
  exceptionClass: INVALID_ARGUMENT
  errorName: DatasetNotFound
  message: DatasetNotFound
  parameters:
    datasetRid: 123abc
    errorId: xyz
}
```

I agree that we could change the message to be consistent with the old format since it's cheap to do so here and would avoid the need for lockstep FE + BE upgrades.

uschi2000:
I think the message is consistent with the old format (assuming stacktracePropagation==false): "Refer to the server logs with this errorId: <errorId>".

See JsonExceptionMapper.

markelliot:
Cool, in that case I don’t expect this is really a break. Can we include errorId as a new structured field?

uschi2000:
in addition to the `Map<String, String> parameters`?

gracew:
I was proposing that when mapping ServiceExceptions, we still output a message field that looks like "Refer to the server logs with this errorId: ", instead of just re-using the errorName. This means backends can upgrade to http-remoting3 without any FE coordination, assuming the FE code does end up propagating this message to the user.

uschi2000:
ServiceException is a new thing, so we don't have to take care of any backcompat when mapping ServiceException.

gracew:
yes, but ServiceExceptionMapper writes out a SerializableError as the response body. in SerializableError we are maintaining the "exceptionClass" and "message" fields for backcompat purposes. I'm saying that not only are clients relying on the presence of these fields, some clients (FE ones) may be propagating them directly to users. Therefore, we might as well make the content of the message field match the previous behavior, until the client has had a chance to switch to the new fields and generate their own message using the structured parameter data.

pnepywoda:
It does match the previous behavior. As Rob said above, it'll look like
message: Refer to the server logs with this errorId: 789xyz
(with no value for stacktrace field)

markelliot:
@rfink yes, top-level next to `errorCode` and `errorName`, I'd like to have `errorId`

pnepywoda:
What's errorId?
  errorId: INVALID_ARGUMENT

or

  errorId: 400

(and what additional info is this adding?)

markelliot:
It’s the UUID we currently stuff in the message.

uschi2000:
OK, I understand Grace's point now. Once a backend upgrades to throw remoting3-style ServiceExceptions, then a remoting2 frontend will start displaying "DatasetNotFound" instead of "Refer to the blabla: errorId" to users.

I don't know what the right answer is; any takers? Paul?

markelliot:
Ok, I’d prefer to find a way not to do that, and to continue to show the original message…

uschi2000:
SerializableError#errorId field: https://github.com/palantir/http-remoting-api/pull/16

uschi2000:
updated. now passing through old-style "Refer to logs: errorId" message fields.

uschi2000:
Will merge so I can move on. Let me know if anything else is missing here.

PR #472, opened 2017-07-07
refactor tracing so that it's straightforward to run a callable with some specified trace

bavardage:
Usecase (there may be an existing way to do this)

I have a scheduled background task which makes requests to other services.

I want each run of this background task to run with a new trace id
- all calls to remote services as part of this single run have this new trace id in common
- all log statements made as part of this single run have the this new trace id in common

currently I do this via the somewhat janky

```java
void run() {
    String traceId = Tracers.randomId();
    Tracer.initTrace(Optional.empty(), traceId);
    MDC.put("traceId", traceId);
    // work of the scheduled task goes here
}
```

I should also probably remove the MDC when done (i.e. set it back to what it was before) and restore the traceid too, given potentially other scheduled tasks could in theory share this same scheduled executor service.

whereas with this PR I can do, I think
```java
void run() {
    Tracer.withTrace(Tracers.createTrace(Optional.empty(), () -> {
        //work of the scheduled task goes here
    });
}
```

Happy to consider other approaches - perhaps there exists something - or maybe this looks like a wrapper around a scheduled executor service, so that each scheduled call is automatically wrapped in a new trace.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/472)

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
@@ -30,7 +30,6 @@ import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
-import org.slf4j.MDC;
 
 @Provider
 public final class TraceEnrichingFilter implements ContainerRequestFilter, ContainerResponseFilter {
@@ -58,10 +57,6 @@ public final class TraceEnrichingFilter implements ContainerRequestFilter, Conta
                 Tracer.startSpan(operation, spanId, SpanType.SERVER_INCOMING);
             }
         }
-
-        // Give SLF4J appenders access to the trace id
-        // TODO(rfink) We should use putCloseable; when and how can we remove it though? There is no filter chain.
-        MDC.put(Tracers.TRACE_ID_KEY, Tracer.getTraceId());
 
         // Give asynchronous downstream handlers access to the trace id
         requestContext.setProperty("com.palantir.remoting2.traceId", Tracer.getTraceId());
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
@@ -41,7 +41,9 @@ public final class Tracer {
     private static final ThreadLocal<Trace> currentTrace = new ThreadLocal<Trace>() {
         @Override
         protected Trace initialValue() {
-            return createTrace(Optional.empty(), Tracers.randomId());
+            Trace trace = createTrace(Optional.empty(), Tracers.randomId());
+            MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
+            return trace;
         }
     };
 
@@ -68,7 +70,7 @@ public final class Tracer {
      * #setSampler configured sampler} returns true.
      */
     public static void initTrace(Optional<Boolean> isObservable, String traceId) {
-        currentTrace.set(createTrace(isObservable, traceId));
+        setTrace(createTrace(isObservable, traceId));
     }
 
     /**
@@ -200,6 +202,7 @@ public final class Tracer {
     public static Trace getAndClearTrace() {
         Trace trace = currentTrace.get();
         currentTrace.remove();
+        MDC.remove(Tracers.TRACE_ID_KEY);
         return trace;
     }
 
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
@@ -40,7 +40,7 @@ public final class Tracer {
     private static final ThreadLocal<Trace> currentTrace = new ThreadLocal<Trace>() {
         @Override
         protected Trace initialValue() {
-            return new Trace(sampler.sample(), Tracers.randomId());
+            return createTrace(Optional.empty(), Tracers.randomId());
         }
     };
 
@@ -51,14 +51,23 @@ public final class Tracer {
     private static TraceSampler sampler = AlwaysSampler.INSTANCE;
 
     /**
+     * Creates a new trace, but does not set it as the current trace. The new trace is {@link
+     * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent,
+     * if the {@link #setSampler configured sampler} returns true.
+     */
+    public static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
+        validateId(traceId, "traceId must be non-empty: %s");
+        boolean observable = isObservable.orElse(sampler.sample());
+        return new Trace(observable, traceId);
+    }
+
+    /**
      * Initializes the current thread's trace, erasing any previously accrued open spans. The new trace is {@link
      * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent, if the {@link
      * #setSampler configured sampler} returns true.
      */
     public static void initTrace(Optional<Boolean> isObservable, String traceId) {
-        validateId(traceId, "traceId must be non-empty: %s");
-        boolean observable = isObservable.orElse(sampler.sample());
-        currentTrace.set(new Trace(observable, traceId));
+        currentTrace.set(createTrace(isObservable, traceId));
     }
 
     /**
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
@@ -198,7 +198,7 @@ public final class Tracer {
         return currentTrace.get().getTraceId();
     }
 
-    /** Clears the current trace id and (a copy of) it. */
+    /** Clears the current trace id and returns (a copy of) it. */
     public static Trace getAndClearTrace() {
         Trace trace = currentTrace.get();
         currentTrace.remove();
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
@@ -55,7 +55,7 @@ public final class Tracer {
      * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent,
      * if the {@link #setSampler configured sampler} returns true.
      */
-    public static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
+    /*package*/ static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
         validateId(traceId, "traceId must be non-empty: %s");
         boolean observable = isObservable.orElse(sampler.sample());
         return new Trace(observable, traceId);
--- a/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 /**
  * The singleton entry point for handling Zipkin-style traces and spans. Provides functionality for starting and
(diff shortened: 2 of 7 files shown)


bavardage:
(also I didn't write tests yet, wanted to get thoughts on approaches first..)

uschi2000:
can you explain the use-case a little more? what is there an existing trace that you don't want to push down into executors/callables?

bavardage:
I'm not sure I understand your last question.

I'm running a scheduled background task via a [ScheduledExecutorService](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html)

There's no request involved (i.e. the [TraceEnrichingFilter](https://github.com/palantir/http-remoting/blob/develop/jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java) will not have run).

The only way an existing trace would be present would be by accident/pollution from some other task sharing the same executor service. I.e, without this, the behaviour is

 - the first remote call made by the scheduled task will result in the (thread local) trace id being set
 - that happens though, I think, via the thread local init, so this trace id isn't set in the mdc
    - ...and so no log statements made will actually get that trace id
    - you'll only really see it showing up in the request logs of the services the task is calling
 - all subsequent runs of the task that happen to re-use this thread will now get the same traceid
 - traceid ceases to be meaningful

maybe easier to explain in person?

markelliot:
What about Tracers.wrap(Callable) and Tracers.wrap(Runnable)?

On Jul 9, 2017, at 22:18, Ben Duffield <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm not sure I understand your last question.

I'm running a scheduled background task via a ScheduledExecutorService[docs.oracle.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__docs.oracle.com_javase_7_docs_api_java_util_concurrent_ScheduledExecutorService.html&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=GRXZQXkMkJg9eOWsw9WwtA5TjfOpcymgXUpTWrBVLRw&s=oUGq2edbdM1itknrG6smVHULDIRBcVMH6UTdYPFtTr0&e=>

There's no request involved (i.e. the TraceEnrichingFilter[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_http-2Dremoting_blob_develop_jersey-2Dservers_src_main_java_com_palantir_remoting2_servers_jersey_TraceEnrichingFilter.java&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=GRXZQXkMkJg9eOWsw9WwtA5TjfOpcymgXUpTWrBVLRw&s=zOy-l-YMji74NZpSr6Cdx_QCu6L7TMaEDIPBtI3E5CY&e=> will not have run).

The only way an existing trace would be present would be by accident/pollution from some other task sharing the same executor service. I.e, without this, the behaviour is

  *   the first remote call made by the scheduled task will result in the (thread local) trace id being set
  *   that happens though, I think, via the thread local init, so this trace id isn't set in the mdc
     *   ...and so no log statements made will actually get that trace id
     *   you'll only really see it showing up in the request logs of the services the task is calling
  *   all subsequent runs of the task that happen to re-use this thread will now get the same traceid
  *   traceid ceases to be meaningful

maybe easier to explain in person?

bavardage:
What about them?

Their behaviour is they make sure that when the callable/runnable actually runs, it runs with the traceid present at the time of construction of the callable/runnable.

This is a scheduled task - constructed at server start - and thus there is no trace id around when it is constructed.

This is sorta the opposite of `Tracers#wrap`

markelliot:
Got it. I'd think that'd work best with a wrapWithoutTrace rather than exposing the withTrace method.

On Jul 9, 2017, at 22:26, Ben Duffield <notifications@github.com<mailto:notifications@github.com>> wrote:


What about them?

Their behaviour is they make sure that when the callable/runnable actually runs, it runs with the traceid present at the time of construction of the callable/runnable.

This is a scheduled task - constructed at server start - and thus there is no trace id around when it is constructed.

This is sorta the opposite of Tracers#wrap

bavardage:
where `wrapWithoutTrace` takes care of making a new traceid? 

I was also pondering whether what I actually want here is a `TracingAwareScheduledExecutorService` which 'magically' handles all of this for you

markelliot:
There is one of those but it ultimately calls wrap, and you want not that.

On Jul 9, 2017, at 22:49, Ben Duffield <notifications@github.com<mailto:notifications@github.com>> wrote:


where wrapWithoutTrace takes care of making a new traceid?

I was also pondering whether what I actually want here is a TracingAwareScheduledExecutorService which 'magically' handles all of this for you

bavardage:
nope, there isn't for a *Scheduled*ExecutorService

bavardage:
made the changes/refactored this on top of the for-3.0.0 branch

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -51,14 +51,23 @@ protected Trace initialValue() {
     private static TraceSampler sampler = AlwaysSampler.INSTANCE;
 
     /**
+     * Creates a new trace, but does not set it as the current trace. The new trace is {@link
+     * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent,
+     * if the {@link #setSampler configured sampler} returns true.
+     */
+    public static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
```
markelliot: I don't think you need this or at least don't need it to be exposed as public anymore.

markelliot: Also instead of "isObservable" it'd be clearer to use a Boolean "useSampler", or perhaps, since we always want to use the sampler, remove thenparameter?

markelliot: And maybe even remove the traceId parameter. Ideal might be we add a clearTrace method which returns a copy of the current trace and zeros the current state, then use that method in "withTrace" instead of copy/set. At that point, this method goes away.

markelliot:
We have a WrappingScheduledExecutorService and one of the Tracers#wrap methods will configure it for many of the calls. If you want to edit which methods we wrap which way, feel free.

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
bavardage: I use `createTrace` in `Tracers#wrapWithoutTrace`

bavardage:
Will look at changing the `WrappingScheduledExecutorService` in FLUP

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
markelliot: Yes, and it could be package private, or per my 3rd comment, be removed in favor of a clearer purpose "clearTrace" method.

bavardage: made it package. I don't really understand how one would use `clearTrace` as you describe - it seems you'd have to call `initTrace` immediately before, which seems a fairly roundabout way of getting a fresh trace.

bavardage: #473 for FLUP

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -51,14 +51,23 @@ protected Trace initialValue() {
     private static TraceSampler sampler = AlwaysSampler.INSTANCE;
 
     /**
+     * Creates a new trace, but does not set it as the current trace. The new trace is {@link
+     * Trace#isObservable observable} iff the given flag is true, or, iff {@code isObservable} is absent,
+     * if the {@link #setSampler configured sampler} returns true.
+     */
+    /*package*/ static Trace createTrace(Optional<Boolean> isObservable, String traceId) {
```
markelliot: + remove params?

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -213,6 +221,9 @@ static Trace copyTrace() {
      */
     static void setTrace(Trace trace) {
         currentTrace.set(trace);
+
+        // Give SLF4J appenders access to the trace id
+        MDC.put(Tracers.TRACE_ID_KEY, trace.getTraceId());
```
markelliot, line 229: we should do this in createTrace as well?

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -195,6 +196,13 @@ public static String getTraceId() {
         return currentTrace.get().getTraceId();
     }
 
+    /** Clears the current trace id and (a copy of) it. */
+    public static Trace getAndClearTrace() {
+        Trace trace = currentTrace.get();
+        currentTrace.remove();
```
markelliot: \+ unset MDC?

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
```diff
@@ -86,26 +85,27 @@ public static Runnable wrap(Runnable delegate) {
      * will have a fresh trace.
      */
     public static <V> Callable<V> wrapWithoutTrace(Callable<V> delegate) {
```
markelliot: sorry to nit pick: `wrapWithNewTrace`

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
```diff
@@ -79,6 +79,34 @@ public static Runnable wrap(Runnable delegate) {
...
+     */
+    public static <V> Callable<V> wrapWithoutTrace(Callable<V> delegate) {
+        return () -> runWithoutTrace(delegate);
+    }
+
+    /**
+     * Runs the given callable after firstly clearing the current trace. The current
+     * trace is restored after running the callable.
+     */
+    private static <T> T runWithoutTrace(Callable<T> callable) throws Exception {
```
markelliot: maybe just inline this?

markelliot (approved):

jersey-servers/src/main/java/com/palantir/remoting2/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -59,10 +58,6 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
             }
         }
 
-        // Give SLF4J appenders access to the trace id
```
uschi2000, line 62: hopefully people upgrade jersey-servers and tracers at the same time ...

bavardage: are we doing `com.palantir.remoting3` for remoting3?

bavardage: ... I guess, not for everything

tracing/src/main/java/com/palantir/remoting2/tracing/Tracer.java
```diff
@@ -186,6 +198,14 @@ public static String getTraceId() {
         return currentTrace.get().getTraceId();
     }
 
+    /** Clears the current trace id and (a copy of) it. */
```
uschi2000: +"returns"

bavardage: done

tracing/src/main/java/com/palantir/remoting2/tracing/Tracers.java
```diff
@@ -79,6 +79,28 @@ public static Runnable wrap(Runnable delegate) {
     }
 
     /**
+     * Wraps the given {@link Callable} such that it creates a fresh {@link Trace tracing state} for its execution.
+     * That is, the trace during its {@link Callable#call() execution} is entirely separate from the trace at
+     * construction or any trace already set on the thread used to execute the callable. Each execution of the callable
+     * will have a fresh trace.
+     */
+    public static <V> Callable<V> wrapWithNewTrace(Callable<V> delegate) {
```
uschi2000, line 87: should have some tests.

uschi2000 (commented):
looks good, but would like to see test coverage for the new functionality.

uschi2000 (approved):

uschi2000:
more checkstyle, sorry. otherwise lgtm.

PR #475, opened 2017-07-19
Implement error handlers for RemoteException

uschi2000:
This change is [Reviewable](https://reviewable.io/reviews/palantir/http-remoting/475)


jaxrs-clients/src/test/java/com/palantir/remoting2/jaxrs/feignimpl/RemoteExceptionsEteTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private ExceptionsTestServer.TestService service;
+
+    @Before
+    public void before() {
+        String endpointUri = "http://localhost:" + APP.getLocalPort();
+        service = JaxRsClient.create(ExceptionsTestServer.TestService.class, "agent", createTestConfig(endpointUri));
+    }
+
+    @Test
+    public void testInvalidArgument() {
```
uschi2000, line 51: @markelliot the error handling we had discussed.

.baseline/checkstyle/checkstyle.xml
```diff
@@ -78,7 +78,6 @@
             <message key="name.invalidPattern" value="Class type name ''{0}'' must match pattern ''{1}''."/>
         </module>
         <module name="CovariantEquals"/> <!-- Java Coding Guidelines: Override ``Object#equals`` consistently -->
-        <module name="DefaultComesLast"/> <!-- Java Style Guide: The default case is present -->
```
gracew: should probably remove these from mainline baseline?

client-error-handling/build.gradle
```diff
@@ -0,0 +1,14 @@
...
+dependencies {
+    compile "com.palantir.remoting-api:errors"
+    compile "org.derive4j:derive4j-annotation"
+
+    testCompile "junit:junit"
+    testCompile "org.assertj:assertj-core"
+    testCompile "org.mockito:mockito-core"
+
+    processor "org.immutables:value"
+    processor "org.derive4j:derive4j:0.12.3"
```
gracew: hm, this didn't work for me when I ran `./gradlew eclipse` -- also the version should go in versions.props

client-error-handling/src/test/java/com/palantir/remoting2/clients/RemoteExceptionsTest.java
```diff
@@ -0,0 +1,85 @@
...
+        assertThat(RemoteExceptions.handleByErrorCode(FAILED_PRECONDITION, BY_CODE_HANDLER))
+                .isEqualTo(FAILED_PRECONDITION);
+        assertThat(RemoteExceptions.handleByErrorCode(INTERNAL, BY_CODE_HANDLER)).isEqualTo(INTERNAL);
+        assertThat(RemoteExceptions.handleByErrorCode(CUSTOM_CLIENT, BY_CODE_HANDLER)).isEqualTo(CUSTOM_CLIENT);
+        assertThat(RemoteExceptions.handleByErrorCode(CUSTOM_SERVER, BY_CODE_HANDLER)).isEqualTo(CUSTOM_SERVER);
+        assertThat(RemoteExceptions.handleByErrorCode(UNKNOWN, BY_CODE_HANDLER)).isEqualTo(UNKNOWN);
+    }
+
+    private static RemoteException createError(ErrorType type) {
+        return new RemoteException(SerializableError.forException(new ServiceException(type)), type.httpErrorCode());
```
gracew, line 83: not sure why the RemoteException constructor needs to take the http code separately now the error type is accessible via the ServiceException

gracew (commented):
can you elaborate on what implementations of `ByErrorFamilies#cases`, `ByErrorCodes#cases` you anticipate? I don't think that most services will be doing so.

jaxrs-clients would probably be the main consumer of this library, to e.g. attempt retries in the case of certain http response codes (429, 503). ErrorType currently doesn't allow that specificity of http status though.

.baseline/checkstyle/checkstyle.xml
uschi2000: already done (this is on oldish baseline). i added back the DefaultComesLast.

client-error-handling/build.gradle
uschi2000: done

client-error-handling/src/test/java/com/palantir/remoting2/clients/RemoteExceptionsTest.java
uschi2000: RemoteException is client-side; it doesn't know about ServiceException (which is server-side)

uschi2000:
re users of the Cases API:
- I wonder if retries (429/503) should be handled at the okhttp level automatically and would not show up here at all?
- Today, most services blindly propagate remote exceptions. I am hoping that this change makes it more obvious/flexible/stable how to react when observing errors. I standard behavior would be to rephrase e server-side exception in terms of my service. E.g., when a Build cannot start a Catalog transaction, it would throw "BuildFailed" instead of propagation "StartTransactionFailed".

jaxrs-clients/src/test/java/com/palantir/remoting2/jaxrs/feignimpl/RemoteExceptionsEteTest.java
```diff
@@ -0,0 +1,109 @@
...
+                        assertThat(ex.getError().errorCode()).isEqualTo(ErrorType.Code.INVALID_ARGUMENT.toString());
+                        assertThat(ex.getError().errorName()).isEqualTo("InvalidArgument");
+                        return ex;
+                    })
+                    .otherwise(() -> {
+                        fail("Did not expect the otherwise() case");
+                        return null;
+                    }));
+
+            RemoteExceptions.handleByErrorCode(e, ByErrorCodes.cases()
```
markelliot, line 67: it'd be interesting to be able to compose these with handleByErrorFamily maybe?

markelliot:
also probably want handleByErrorName?

markelliot:
and wondering if we want to be able to wrap this up one more level to handle the successful version of a call/reduce the try/catch boilerplate a bit

jaxrs-clients/src/test/java/com/palantir/remoting2/jaxrs/feignimpl/RemoteExceptionsEteTest.java
uschi2000: you can, roughly:
```
RemoteExceptions.handleByErrorCode(e, ByErrorCode.cases()
  .client(clientEx -> RemoteExceptions.handleByErrorFamily(clientEx, ...))
```
(it's not pretty, but i wager it's also not frequent)

what API did you have in mind?

uschi2000:
thought about handleByErrorName; those are not known at http-remoting-compile-time, so I think the caller is responsible for making such a case class or use vanilla java switch on the errorName?

uschi2000:
gRPC convinced/inspired me that try/catch is an acceptable API for the blocking API. was planning on doing the more general thing with dialogue's async API, where you do something like:
```
client.myEndpoint("foo", "bar", new Handler()
  .onSuccess(...)
  .onClientError(...)
  .onMaybeSomeOtherTypeOfError()
);
```

part of the rationale is also to not make the remoting 2->3 upgrade too infeasible.

markelliot:
Thinking I want to be able to write something like:

Remote.invoke(() -> service.call())
  .ok(result -> { … }) // default is identity
  .clientError(error -> { … })
  .namedError(“SomeNamedErrorWeCareAbout”, error -> { … })
  .codedError(403, error -> {…})
  .run(); // or runAsync, run returns result of the `ok` call.

uschi2000:
Those cases must be mutually exclusive, so I don't think you can mix'n'match that easily.

markelliot:
Ordered list of matchers seems fine?

uschi2000:
not a fan, seems brittle when behavior changes after reordering.

markelliot:
It’s going to be important to deal with named errors one way or another, but it seems weird not to be able to handle a bunch of named errors that are, say, client problems, and then lump all server behaviors into another bucket…

uschi2000:
OK, let's discuss in person tomorrow.

uschi2000:
Decided to punt on this for now.

PR #482, opened 2017-07-26
bump okhttp3 to avoid infinite retry bug

nziebart:
okhttp 3.4.1 has a known issue where connections can get stuck in an infinite retry loop (https://github.com/square/okhttp/issues/2756)

We saw this issue in the field with atlasdb's timelock server, and it resulted in all three timelock nodes plus two client service nodes (internal auth service) getting into a terminally unhealthy state and requiring a restart (essentially, all available application threads got into this state, and the services became unresponsive).

We've disabled okhttp-level retries for the time being (https://github.com/palantir/atlasdb/issues/2194), and are instead relying on Feign's default retry logic. However, okhttp claims that this issue is fixed in 3.4.2.


nziebart:
This is failing on 
```
 Could not find com.palantir:gradle-baseline-java-config:0.9.2.
     Searched in the following locations:
         https://jcenter.bintray.com/com/palantir/gradle-baseline-java-config/0.9.2/gradle-baseline-java-config-0.9.2.pom
         https://jcenter.bintray.com/com/palantir/gradle-baseline-java-config/0.9.2/gradle-baseline-java-config-0.9.2.zip
```
Not sure why.

uschi2000:
Have to add the bintray bucket explicitly, see baseline readme.

uschi2000:
(I think the circle cached got cleared, so all builds on develop are now failing)

uschi2000:
we're going to okhttp 3.8.1 in remoting3, so I think we can close this?

PR #499, opened 2017-08-14
Add QosExceptionMapper

uschi2000:


uschi2000:
Depends on -api being released first.

markelliot (approved):

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/QosExceptionMapper.java
```diff
@@ -0,0 +1,89 @@
...
+            throw new RuntimeException(
+                    "Internal error, unknown QosException subclass: " + exception.getClass().getSimpleName());
+        }
+    }
+
+    private Response retryAnyNow() {
+        return Response.status(307).build();
+    }
+
+    private Response retryAnyLater(Duration backoff) {
```
bavardage: yeah so, I wanna be able to send a 429 without a RETRY_AFTER header 

this is valid http...

> The response representations SHOULD include details explaining the condition, and MAY include a Retry-After header indicating how long to wait before making a new request.

bavardage: would have suggested an optional duration, but the fact that we use presence up in line 49 to decide what to do :(

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/QosExceptionMapper.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final Logger log = LoggerFactory.getLogger(QosExceptionMapper.class);
+
+    @Override
+    public Response toResponse(QosException exception) {
+        log.debug("Possible quality-of-service intervention", exception);
+        if (exception instanceof QosException.Retry) {
+            QosException.Retry retry = (QosException.Retry) exception;
+            if (!retry.getBackoff().isPresent() && !retry.getRedirectTo().isPresent()) {
+                return retryAnyNow();
+            } else {
```
bavardage: and we don't support signalling redirect to this location after this time?

bavardage: left a comment on https://github.com/palantir/http-remoting-api/pull/30/files - we shouldn't allow such a Retry then

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/QosExceptionMapper.java
bavardage: gut is that it would be cleaner to just have a subclass of QosException per response code

uschi2000: yep, can do that.

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/QosExceptionMapper.java
uschi2000: will use subclasses of QosException, should be cleaner then

uschi2000:
@bavardage ready for you

jared2501 (approved):

bavardage (approved):
lgtm!

PR #501, opened 2017-08-15
Fix optional query parameters in Retrofit: convert to null if not present

uschi2000:


uschi2000:
I guess this yields the expected behavior, but it's a little odd still: the converter docs ask us to return `null` if we cannot handle the parameter. we return `null` to indicate an empty/non-existent parameter.

bavardage (approved):

PR #502, opened 2017-08-16
Wrap the OkHttp thread pool for tracing

j-baker:
Additionally, use the same thread pool for every retrofit client


retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/Retrofit2ClientBuilder.java
```diff
@@ -108,12 +114,7 @@ private OkHttpClient createOkHttpClient(String userAgent, List<String> uris) {
         // increase default connection pool from 5 @ 5 minutes to 100 @ 10 minutes
         client.connectionPool(new ConnectionPool(100, 10, TimeUnit.MINUTES));
 
-        Dispatcher dispatcher = new Dispatcher();
-
-        dispatcher.setMaxRequests(256);
-        dispatcher.setMaxRequestsPerHost(256);
-
-        client.dispatcher(dispatcher);
+        client.dispatcher(createDispatcher());
```
uschi2000, line 114: what about the jaxrs one?

j-baker: This is only used for the async flow, which is never used with the JAX-RS clients.

uschi2000 (approved):

PR #505, opened 2017-08-17
IOException not IllegalStateException

### LINKED ISSUE #504: CompletableFuture.join hangs if connection to server fails
Hitting a retrofit endpoint that returns a `CompletableFuture` and calling `join` will block forever if the connection to server fails (invalid URL, server not running, etc). There's an exception thrown in the OkHttp Dispatcher thread, but it's not properly caught by the call handler for `CompletableFuture`.

Repro at https://github.com/palantir/http-remoting/pull/503

j-baker:
OkHttp interceptors are not allowed to throw IllegalStateExceptions.

The contract of an OkHttp call is...

```
   * @throws IOException if the request could not be executed due to cancellation, a connectivity
   * problem or timeout. Because networks can fail during an exchange, it is possible that the
   * remote server accepted the request before the failure.
```

which throwing an IllegalStateException in an interceptor breaks. This breaks assumptions
in OkHttp, and notably will cause callbacks to never be triggered.

I didn't notice this when I added the CompletableFuture retrofit
support (only noticed the other interceptor).

Fixes #504.

### DIFF
--- a/retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/MultiServerRetryInterceptor.java
+++ b/retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/MultiServerRetryInterceptor.java
@@ -88,7 +88,7 @@ public final class MultiServerRetryInterceptor implements Interceptor {
                 logger.warn("Failed to send request to {}", request.url(), e);
             }
         }
-        throw new IllegalStateException("Could not connect to any of the following servers: " + urls + ". "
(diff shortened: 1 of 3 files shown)


j-baker:
Incorporated the test from #503 also.

uschi2000 (approved):

meditativeape:
This bug is also present in remoting2. It's blocking me from upgrading another library that uses `CompletableFuture` in its Retrofit service definition. Ideally, I'd like it to be independent of the migration to remoting3. Can we port this fix to remoting2?

PR #509, opened 2017-08-21
 Use okhttp UserAgentInterceptor in Feign clients

uschi2000:


uschi2000:
Ping. Would like to make some progress here tomorrow, would appreciate feedback by EOT if possible.

j-baker (approved):

PR #510, opened 2017-08-21
Feign clients use OkHttp MultiServerRetryInterceptor instead of FailoverFeignTarget

uschi2000:


jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -88,17 +95,12 @@
...
                 .options(createRequestOptions())
                 .logger(new Slf4jLogger(JaxRsClient.class))
                 .logLevel(Logger.Level.BASIC)
                 .requestInterceptor(UserAgentInterceptor.of(userAgent))
-                .target(target);
-    }
-
-    private <T> FailoverFeignTarget<T> createTarget(Class<T> serviceClass, List<String> uris) {
-        return new FailoverFeignTarget<>(uris, serviceClass, NeverRetryingBackoffStrategy.INSTANCE);
+                .target(serviceClass, primaryUri);
```
gracew, line 101: can we use feign's EmptyTarget instead?

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -62,19 +62,26 @@
...
+     * that the {@link MultiServerRetryInterceptor} replaces this URI with a random URI from the client configuration
+     * before making the request.
+     */
+    private final String primaryUri;
 
     AbstractFeignJaxRsClientBuilder(ClientConfiguration config) {
-        this.config = config;
         Preconditions.checkArgument(config.maxNumRetries() == 0,
                 "Connection-level retries are not supported by %s", JaxRsClient.class.getSimpleName());
+        Preconditions.checkArgument(!config.uris().isEmpty(), "Must provide at least one service URI");
```
gracew, line 75: don't think we can require this as per your earlier email saying that initial uri lists may be empty

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
uschi2000: I don't think so, it throughs `java.lang.UnsupportedOperationException: Request with non-absolute URL not supported with empty target` before hitting the okhttp code path that could inject a URL.

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
uschi2000: i think we'll have to do the filtering for valid/invalid configurations on the witchcraft side. this preconditions here only makes an existing hidden constraint more visible: https://github.com/palantir/http-remoting/blob/8b1f898639ad1b7eaeb854be9dc6f84b0a44baa3/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/feignimpl/FailoverFeignTarget.java#L133 fails when the list of servers is empty.

uschi2000:
Ping. Would like to make some progress here tomorrow, would appreciate feedback by EOT if possible.

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -62,19 +62,26 @@
...
+        this.config = config;
+        this.primaryUri = config.uris().get(new Random().nextInt(config.uris().size()));
```
j-baker: if this isn't going to be used, then we should just pick the first one, just cos it's irrelevant (the reader would assume this is in some way relevant).

j-baker:
Is there any logging that'd be affected by this? Not sure what level the 'i made a call' logs appear at, but they could easily be pointing at the wrong service url now.

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
uschi2000: done

uschi2000:
Re logging: that's a valid concern. We inject an Slf4j logger which logs something like this at the DEBUG level:
```
    log(configKey, "---> %s %s HTTP/1.1", request.method(), request.url());
```
That `request.url()` is going to have this `.get(0)` URL in it that might not actually be used at the OkHttp layer to really make the request. Not sure what to do about this :/

uschi2000:
One option is to turn off feign logging and use okhttp logging only.

j-baker:
We should probably inject an OkHttp logging interceptor, then (unless one already exists)

uschi2000:
Yep, agree

uschi2000:
See https://github.com/palantir/http-remoting/pull/513 . That should unblock the logging concern in this PR.

gracew:
lgtm with the addition of the logging PR

PR #511, opened 2017-08-21
retry if http code is 307

jamesthomp:
As part of an internal spec, http code 307 should be retried.
This PR contains the tests and a minimal implementation of this part of the spec.


uschi2000:
I'd like to implement this at the okhttp level. Actually just started looking into this, so happy to take over work for it.

jamesthomp:
go for it - I had implemented these tests anyway to check for myself that http-remoting wasn't retrying, and when I discovered that the fix was 8LOC, I thought I might as well just PR it for consideration.

PR #514, opened 2017-08-23
Disable Feign retry logic

uschi2000:
This is a follow-up to #510: We implement retry logic in OkHttp and disable all retrying
in Feign. In particular, without this change Feign would treat all IOExceptions thrown by OkHttp
as retryable.


uschi2000:
FYI, @gracew

PR #516, opened 2017-08-23
Feign/JaxRs clients use OkHttp-level error handling instead of Feign error decoder

uschi2000:


uschi2000:
note that error decoding still has full test coverage via SerializableErrorErrorDecoderTests and SerializableErrorInterceptorTest

markelliot (approved):

PR #518, opened 2017-08-27
Factor out UrlSelector from MultiServerRetryInterceptor

uschi2000:


uschi2000:
FYI @divath

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        return new UrlSelectorImpl(ImmutableList.copyOf(canonicalUrls.build()));
+    }
+
+    @Override
+    public Optional<HttpUrl> redirectTo(HttpUrl current, String redirectBaseUrl) {
+        return redirectTo(current, HttpUrl.parse(redirectBaseUrl));
+    }
+
+    private Optional<HttpUrl> redirectTo(HttpUrl current, HttpUrl redirectBaseUrl) {
+        return baseUrlFor(redirectBaseUrl).flatMap(baseUrl -> {
```
markelliot, line 57: Can we simplify a bunch of this if we require all urls to have the same base path at construction time?

uschi2000: I think you still need to do the prefix matching, so it probably doesn't make a big difference?

markelliot: Is it that where we do all this work we no longer know the distinction between the application's requested path and the URL we're adding it to?

uschi2000: yeah.

uschi2000: we construct the okhttp interceptor with the discovery-consumed service URLs. we then intercept a feign/retrofit - generated URL that contains richer paths, parameters, query parameters, etc. redirecting here means replacing its base path by another one.

markelliot (approved):

PR #519, opened 2017-08-28
UserAgentInterceptor sends http-remoting version and API version if available

uschi2000:
…vailable

### DIFF
--- a/gradle/publish-base.gradle
+++ b/gradle/publish-base.gradle
@@ -9,7 +9,7 @@ jar {
         attributes(
                 "Implementation-Title"   : project.name,
                 "Implementation-Version" : project.version,
-                "Implementation-Vendor"  : "Palantir Technologies, Inc.")
+                "Implementation-Vendor"  : "Palantir Technologies Inc.")
     }
 }
 
--- a/gradle/publish-base.gradle
+++ b/gradle/publish-base.gradle
@@ -4,6 +4,15 @@ apply plugin: 'nebula.maven-resolved-dependencies'
 apply plugin: 'nebula.javadoc-jar'
 apply plugin: 'nebula.source-jar'
 
+jar {
+    manifest {
+        attributes(
+                "Implementation-Title"   : project.name,
+                "Implementation-Version" : project.version,
+                "Implementation-Vendor"  : "Palantir Technologies, Inc.")
+    }
+}
+
 bintray {
     user = System.env.BINTRAY_USERNAME
     key = System.env.BINTRAY_PASSWORD
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
@@ -16,8 +16,11 @@
 
 package com.palantir.remoting3.okhttp;
 
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.Pattern;
 import okhttp3.Interceptor;
 import okhttp3.Request;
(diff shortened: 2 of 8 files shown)


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
```diff
@@ -28,14 +31,31 @@
...
+
+    private String getVersionString(Class<?> serviceClass) {
+        List<String> versions = new ArrayList<>(2);
+        String maybeRemotingVersion = this.getClass().getPackage().getImplementationVersion();
+        if (maybeRemotingVersion != null) {
+            versions.add("http-remoting " + maybeRemotingVersion);
+        }
+        String maybeServiceVersion = serviceClass.getPackage().getImplementationVersion();
+        if (maybeServiceVersion != null) {
+            versions.add(String.format("Service API %s %s", serviceClass.getSimpleName(), maybeServiceVersion));
```
schlosna: Would try to avoid String.format on this hot path as we could just do String concat and avoid the expensive format parsing

uschi2000: Can do, but it's really not that hot: This executes once when making a JaxRsClient.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
```diff
@@ -28,14 +31,31 @@
     private static final Pattern VALID_USER_AGENT = Pattern.compile("[A-Za-z0-9()\\-#;/.,_\\s]+");
     private final String userAgent;
 
-    private UserAgentInterceptor(String userAgent) {
+    private UserAgentInterceptor(String userAgent, Class<?> serviceClass) {
```
markelliot, line 34: We decided to get rid of service class version parsing in favor of passing explicit service+version at config time. Don't think we want to add this back.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
```diff
@@ -28,14 +31,31 @@
...
+        this.userAgent = version.isEmpty()
+                ? userAgent
+                : String.format("%s (%s)", userAgent, version);
+    }
+
+    private String getVersionString(Class<?> serviceClass) {
+        List<String> versions = new ArrayList<>(2);
+        String maybeRemotingVersion = this.getClass().getPackage().getImplementationVersion();
+        if (maybeRemotingVersion != null) {
+            versions.add("http-remoting " + maybeRemotingVersion);
```
markelliot: To make analysis simpler we probably want this to come second.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
uschi2000: That's a different beast:
- In WC: version of the server/product
- Here: version of http-remoting and version of the (remote) API service class

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
markelliot: I see, ok.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UserAgentInterceptor.java
```diff
@@ -28,14 +31,31 @@
...
+        List<String> versions = new ArrayList<>(2);
+        String maybeServiceVersion = serviceClass.getPackage().getImplementationVersion();
+        if (maybeServiceVersion != null) {
+            versions.add("Service API " + serviceClass.getSimpleName() + " " + maybeServiceVersion);
+        }
+        String maybeRemotingVersion = this.getClass().getPackage().getImplementationVersion();
+        if (maybeRemotingVersion != null) {
+            versions.add("http-remoting " + maybeRemotingVersion);
+        }
+        return Joiner.on(",").join(versions);
```
markelliot: Based on https://msdn.microsoft.com/en-us/library/ms537503(v=vs.85).aspx I think we want to use semicolons.

markelliot: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent describes a slightly different convention (but still use semicolons)

gradle/publish-base.gradle
```diff
@@ -4,6 +4,15 @@ apply plugin: 'nebula.maven-resolved-dependencies'
 apply plugin: 'nebula.javadoc-jar'
 apply plugin: 'nebula.source-jar'
 
+jar {
+    manifest {
+        attributes(
+                "Implementation-Title"   : project.name,
+                "Implementation-Version" : project.version,
+                "Implementation-Vendor"  : "Palantir Technologies, Inc.")
```
gracew: no comma, as per https://github.com/palantir/gradle-baseline/pull/206?

gracew (approved):

gradle/publish-base.gradle
uschi2000: ah, cheers.

PR #520, opened 2017-08-28
WIP: wrap OkHttp Call to intercept and retry QoS exceptions

uschi2000:


uschi2000:
goal here is to catch 429-type status codes (basically anything that requires retry after backoff) and not block the thread in order to wait for a retry time slot. this goal rules out any Interceptor-based solution, I think.

maybe the QosIoException thing is a bit roundabout.

uschi2000:
Will re-open when it's ready.

PR #526, opened 2017-08-31
Finish QosIoExceptionInterceptor and add tests

uschi2000:


uschi2000:
This seems uncontroversial, merging into rfink/qos

PR #528, opened 2017-08-31
SerializableErrorInterceptor should not count response 101 as error

jared2501:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/101


uschi2000 (approved):

qinfchen:
should the interceptor treat all 1xx as non-error status since they are all informational?

Issue #527, opened 2017-08-31
MultiServerRetryInterceptor should not use HttpUrl


jared2501:
So that MultiServerRetryInterceptor can be used with websocket URIs that follow the format "wss://localhost/path"

jared2501:
Hmm actually reading the code a bit more, not sure you can do this

jared2501:
https://github.com/square/okhttp/issues/1652

PR #529, opened 2017-08-31
Add ws:// and wss:// suport to UrlSelectorImpl

jared2501:


jared2501:
@uschi2000 - thoughts on this as a fix to #527? Looks like OKHttp does [this in Request.Builder#url](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Request.java#L135).

See the discussion on https://github.com/square/okhttp/issues/1652 and https://github.com/square/okhttp/pull/1663.

If it looks good I'll write some tests.

jared2501:
Note gave this a quick test in a project I'm working on and looks like it works.

jared2501:
Went ahead and wrote a unit test - not sure if any other tests should be added. Perhaps we should add an end-to-end web socket test some where?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/UrlSelectorTest.java
```diff
@@ -123,6 +126,17 @@ public void testRedirectToNext() throws Exception {
...
 
+    @Test
+    public void testWorksWithWebSockets() throws Exception {
+        Request wsRequest = new Request.Builder()
+                .url("wss://foo/a")
+                .build();
+        UrlSelectorImpl selector = UrlSelectorImpl.create(ImmutableList.of("wss://foo/", "wss://bar/"));
+
+        // Silently replace web socket URLs with HTTP URLs. See https://github.com/square/okhttp/issues/1652.
+        assertThat(selector.redirectToNext(wsRequest.url())).isEqualTo(Optional.of(parse("https://bar/a")));
```
uschi2000, line 137: Where is the "ws" used at all? I find it a bit funny to return "http://foo" as the "next" url for an initial url "ws://foo".

jared2501: Technically you're not doing this since the URL on the Request class is `http://foo` since OkHTTP converts transparently under the hood.

jared2501: See the code I linked above

uschi2000:
OK.

jared2501:
Woo! ready to merge? any chance of an release/rc with this in it?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -38,7 +38,7 @@ private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls) {
     static UrlSelectorImpl create(Collection<String> baseUrls) {
```
uschi2000, line 38: can add docs about the baseUrls, now that this is somewhat non-obvious?

jared2501: done

uschi2000 (approved):

PR #532, opened 2017-09-03
Implement AsyncQosIoExceptionHandler for Unavailable and Throttle errors

uschi2000:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
```diff
@@ -44,8 +48,19 @@
         return qosIoException.getQosException().accept(new QosException.Visitor<ListenableFuture<Response>>() {
             @Override
             public ListenableFuture<Response> visit(QosException.Throttle exception) {
-                // TODO(rfink): Implement.
-                return Futures.immediateFailedFuture(qosIoException);
+                Optional<Duration> duration = exception.getRetryAfter();
```
uschi2000: @chrisalice @j-baker do you think this backoff logic is reasonable?

j-baker: for an obvious style thing, I'd try to do something more like

```java
Optional<Duration> backoff = exception.getRetryAfter().isPresent() ? exception.getRetryAfter() : backoffStrategy.nextBackOff()
```

and then only have one nested if.

However, I think I'm going to have to spend some more time grokking the general picture.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -44,6 +47,12 @@
      */
     private static final ExecutorService executorService = Tracers.wrap(new Dispatcher().executorService());
 
+    /**
+     * The {@link ScheduledExecutorService} used for scheduling call retries.
+     */
+    private static final ScheduledExecutorService scheduledExecutorService =
+            Tracers.wrap(new ScheduledThreadPoolExecutor(0, Util.threadFactory("OkHttp Scheduler", false)));
```
j-baker: we should indicate that this is a remoting threadpool and not an OkHttp threadpool.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
```diff
@@ -56,11 +68,12 @@
...
             public ListenableFuture<Response> visit(QosException.Unavailable exception) {
-                if (!backoffStrategy.nextBackoff().isPresent()) {
+                Optional<Duration> backoff = backoffStrategy.nextBackoff();
+                if (!backoff.isPresent()) {
                     return Futures.immediateFailedFuture(qosIoException);
                 } else {
-                    // TODO(rfink): Use duration and schedule retry for later.
-                    return executorService.submit(() -> call.clone().execute());
+                    return executorService.schedule(
+                            () -> call.clone().execute(), backoff.get().toMillis(), TimeUnit.MILLISECONDS);
```
alicederyn, line 76: Consider adding the QoS exception as a suppressed exception if the cloned call fails

uschi2000: How would you imagine this should work? I could inject it into the QosIoExceptionAwareCall call when cloning, but is that really worth it?

alicederyn: Yes, that'd work. It might be useful when things are going south to be able to see how the service was responding before the client gave up, and how the client reacted. Up to you though.

alicederyn: You'd probably need to timestamp the exceptions too

uschi2000: OK, I think it'd be more manageable (and possibly useful) to log suitable information and piece things together with our log analysis tooling afterwards. Will add log statements liberally.

PR #533, opened 2017-09-03
Support async calls in QosIoExceptionAwareCall

uschi2000:


j-baker:
My only thing with this stuff is that I think the place you're _meant_ to do this kind of retry logic is in an interceptor - the effect is that you e.g. don't need to write custom code for async. What was the reason for this?

uschi2000:
My understanding of the OkHttp flow is that in the context of an interceptor I'm trapped in the thread that's currently executing the request. For async calls, this means that N calls that are waiting for a later retry are consuming all N OkHttp threads. Hence the circumnavigation. I don't particularly like it and would love to hear about better alternatives, but I currently don't know one.

uschi2000:
Moved tests around a bit and added end-to-end tests for Retrofit (sync and async) and Feign clients, verifying that 503 get retried.

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
```diff
@@ -79,17 +79,26 @@ public void testConsecutiveCalls() throws Exception {
...
                         .build());
-        assertThat(proxy.string(), is("foo"));
+        assertThat(bogusHostProxy.string(), is("foo"));
         assertThat(server1.getRequestCount(), is(1));
     }
+
+    @Test
+    public void testQosError_performsFailover() throws Exception {
+        server1.enqueue(new MockResponse().setResponseCode(503));
+        server1.enqueue(new MockResponse().setBody("\"foo\""));
```
alicederyn, line 99: Do we not guarantee not retrying the same server twice?

uschi2000: No.

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -125,6 +133,29 @@ public void doesNotShareBackoffStateBetweenCalls() throws Exception {
...
+            public void onFailure(Call call, IOException ioException) {
+                future.completeExceptionally(ioException);
+            }
+
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                future.complete(response.body().string());
+            }
+        });
+        assertThat(future.get()).isEqualTo("pong");
```
alicederyn, line 155: Generally best to use the calls that time out, as otherwise a code failure becomes a hard-to-diagnose hung test

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/QosIoExceptionInterceptorTest.java
```diff
@@ -60,7 +60,7 @@ public void test308WithoutLocation() throws Exception {
 
     @Test
     public void test308WithLocation() throws Exception {
-        URL url = new URL("http://foo");
+        URL url = new URL("http://127.0.0.1");
```
alicederyn, line 63: Should we not test with a different URL than the request?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
uschi2000: good idea.

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/QosIoExceptionInterceptorTest.java
uschi2000: done

uschi2000: https://github.com/palantir/http-remoting/pull/534

j-baker:
Rob, I don't quite get it. With this implementation, aren't you blocking two threads?

j-baker:
Ah, I figured it out :) It's because if you're using a ScheduledExecutorService you're not just doing the equivalent of like, Thread.sleep.

uschi2000:
Yep. The construction is a bit crazy ("scheduled recursion"?), bit I think it should be OK? Would love to do something simpler if you have a suggestion...

alicederyn:
Scheduled callback to the same function makes sense to me FWIW

alicederyn:
Though, as I mentioned before, it seems weird to have a thread sleeping + doing nothing

uschi2000:
Sorry, which thread exactly?

alicederyn:
https://github.com/palantir/http-remoting/blob/9bba3ac5d8e24ec20b2687a70beee08bbc980c40/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/QosIoExceptionAwareCall.java#L45

uschi2000:
But this is unavoidable -- it's a synchronous call that's coming in, we'll have to block a thread somewhere, no?

alicederyn:
Sure, but that thread could handle all the sleeping + retrying etc rather than a separate executor pool.

PR #541, opened 2017-09-11
OkHttp interceptors memorize the current target URL

uschi2000:


retrofit2-clients/src/test/java/com/palantir/remoting3/retrofit2/TestService.java
```diff
@@ -50,9 +50,6 @@
     @GET("relative")
     Call<String> getRelative();
 
-    @GET("/absolute")
```
uschi2000, line 53: Absolute URLs don't work with the redirect mechanism. I suspect we don't really need absolute URLs anyway for any of our RPC?

uschi2000:
ping

markelliot:
not really sure I understand what you're changing here

uschi2000:
"not really sure I understand what you're changing here"
with this change the client memorizes the target URL across calls. e.g., after a 308 redirect, subsequent calls also go to the redirected node rather than the original one

PR #543, opened 2017-09-12
record response metrics

ellisjoe:
these are generally useful, but the explicit goal here is to start recording enough information such that we can determine when interactions with services we depend on are unhealthy.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
@@ -18,7 +18,9 @@ package com.palantir.remoting3.okhttp;
 
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.SharedMetricRegistries;
 import java.io.IOException;
+import java.util.Optional;
 import okhttp3.Interceptor;
 import okhttp3.Response;
 
@@ -34,7 +36,7 @@ public final class InstrumentedInterceptor implements Interceptor {
     private final Meter serverError;
     private final Meter other;
 
-    public InstrumentedInterceptor(MetricRegistry registry, String name) {
+    InstrumentedInterceptor(MetricRegistry registry, String name) {
         informational = registry.meter(MetricRegistry.name(name, "response", "family", "informational"));
         successful    = registry.meter(MetricRegistry.name(name, "response", "family", "successful"));
         redirection   = registry.meter(MetricRegistry.name(name, "response", "family", "redirection"));
@@ -69,5 +71,11 @@ public final class InstrumentedInterceptor implements Interceptor {
         }
 
         return response;
+    }
+
+    static InstrumentedInterceptor withDefaultMetricRegistry(String name) {
+        MetricRegistry registry = Optional.ofNullable(SharedMetricRegistries.tryGetDefault())
+                .orElseGet(MetricRegistry::new);
+        return new InstrumentedInterceptor(registry, name);
     }
 }
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting3.okhttp;
+
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import java.io.IOException;
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * Records metrics about the response codes of http requests.
+ */
+public final class InstrumentedInterceptor implements Interceptor {
+
+    private final Meter informational;
+    private final Meter successful;
+    private final Meter redirection;
+    private final Meter clientError;
+    private final Meter serverError;
+    private final Meter other;
+
+    public InstrumentedInterceptor(MetricRegistry registry, String name) {
+        informational = registry.meter(MetricRegistry.name(name, "response", "code", "1xx"));
+        successful    = registry.meter(MetricRegistry.name(name, "response", "code", "2xx"));
+        redirection   = registry.meter(MetricRegistry.name(name, "response", "code", "3xx"));
+        clientError   = registry.meter(MetricRegistry.name(name, "response", "code", "4xx"));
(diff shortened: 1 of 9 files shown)


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
+import java.io.IOException;
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * Records metrics about the response codes of http requests.
+ */
+public final class InstrumentedInterceptor implements Interceptor {
```
schlosna, line 31: Curious to see what granularity you're tracking these -- per target service interface or per individual endpoint.

schlosna (commented):
We might want to pull this out into its own module to avoid pushing the metrics dependency on all the other modules.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
ellisjoe: currently interested in service level

ellisjoe:
@schlosna what's the concern with the metrics dependency?

versions.props
```diff
@@ -10,6 +10,7 @@ com.palantir.remoting-api:* = 1.2.1
 com.palantir.safe-logging:safe-logging = 0.1.3
 com.squareup.okhttp3:* = 3.8.1
 com.squareup.retrofit2:* = 2.3.0
+io.dropwizard.metrics:metrics-core = 3.2.3
```
schlosna, line 13: 3.2.4 is out now, see https://github.com/dropwizard/metrics/releases

@ellisjoe No specific concern right now, more just that it's another transitive library dependency that we're defining and potential diamond dependency conflicts in the future for folks that might have the metrics version imposed by Dropwizard server. I realize for most non-DW servers we'll specify it at the server level and set it in the BOM so should be fine.

ellisjoe: cool so I think the only way to pull this out into a separate module is to have remoting expose some way of injecting something that looks like an interceptor. I know when I talked to @uschi2000 awhile back we wanted to avoid that (reduce api surface), but maybe now we would be happy defining our own interceptor interface? or just something that is able to get a read-only view of requests/responses?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final Meter informational;
+    private final Meter successful;
+    private final Meter redirection;
+    private final Meter clientError;
+    private final Meter serverError;
+    private final Meter other;
+
+    public InstrumentedInterceptor(MetricRegistry registry, String name) {
```
schlosna: Its not hooked up to the OkHttp client right now, how are you envisioning wiring in the metric registry?

ellisjoe: either plumbing it through the JaxRs.create(..) method or using the default shared metric registry: https://github.com/dropwizard/metrics/blob/4.0-development/metrics-core/src/main/java/com/codahale/metrics/SharedMetricRegistries.java

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+        clientError   = registry.meter(MetricRegistry.name(name, "response", "code", "4xx"));
+        serverError   = registry.meter(MetricRegistry.name(name, "response", "code", "5xx"));
+        other         = registry.meter(MetricRegistry.name(name, "response", "code", "other"));
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Response response = chain.proceed(chain.request());
+
+        switch (javax.ws.rs.core.Response.Status.Family.familyOf(response.code())) {
```
uschi2000, line 55: import?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/InstrumentedInterceptorTest.java
```diff
@@ -0,0 +1,102 @@
...
+        registry = new MetricRegistry();
+        interceptor = new InstrumentedInterceptor(registry, "client");
+    }
+
+    @Test
+    public void testInformational() throws IOException {
+        responseWithCode(100);
+        interceptor.intercept(chain);
+
+        assertThat(registry.getMeters().get("client.response.code.1xx").getCount()).isEqualTo(1);
```
uschi2000: hmm. maybe name this `client.response.family.informational`? (imagine we'd later want to also explicitly capture, say 204; then, does the 2xx count include the 204s? you get around this problem by capturing the family independently of the aggregated codes.)

versions.props
uschi2000: wouldn't you want to instrument all clients anyway? or just those in WC?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/InstrumentedInterceptorTest.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Test
+    public void testSuccessful() throws IOException {
```
uschi2000: if you do the family thing suggested above, then you can do something like
```
for (int code : ImmutableList.of(100, 200, ...)) {
  ..
}
```

versions.props
ellisjoe: Ideally instrument all clients if we're happy accepting the diamond dependency. I think they've been good about backcompat so not too worried.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
ellisjoe: already imported the okhttp3.Response

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/InstrumentedInterceptorTest.java
ellisjoe: done

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/InstrumentedInterceptorTest.java
ellisjoe: done. need to map code -> family name as well.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+    private final Meter clientError;
+    private final Meter serverError;
+    private final Meter other;
+
+    public InstrumentedInterceptor(MetricRegistry registry, String name) {
+        informational = registry.meter(MetricRegistry.name(name, "response", "family", "informational"));
+        successful    = registry.meter(MetricRegistry.name(name, "response", "family", "successful"));
+        redirection   = registry.meter(MetricRegistry.name(name, "response", "family", "redirection"));
+        clientError   = registry.meter(MetricRegistry.name(name, "response", "family", "client-error"));
+        serverError   = registry.meter(MetricRegistry.name(name, "response", "family", "server-error"));
```
ellisjoe, line 47: @schlosna do we have naming guidelines for metrics? are hyphens the standard here?

ellisjoe:
@uschi2000 wired in the interceptor. any other concerns?

uschi2000 (approved):

PR #544, opened 2017-09-14
log the uri template as the trace operation

ellisjoe:


jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -28,17 +28,27 @@
...
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
+import org.glassfish.jersey.server.ExtendedUriInfo;
 
 @Provider
 public final class TraceEnrichingFilter implements ContainerRequestFilter, ContainerResponseFilter {
     public static final TraceEnrichingFilter INSTANCE = new TraceEnrichingFilter();
 
+    @Context
+    private ExtendedUriInfo uriInfo;
```
pnepywoda, line 42: how is this different from `requestContext.getUriInfo()` ?

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -28,17 +28,27 @@
...
+
     // Handles incoming request
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        String operation = requestContext.getMethod() + " /" + requestContext.getUriInfo().getPath();
+        String path = Optional.ofNullable(uriInfo)
+                .flatMap(info -> Optional.ofNullable(info.getMatchedModelResource()))
+                .flatMap(resource -> Optional.ofNullable(resource.getPath()))
```
pkoenig10: You can just use `Optional#map`. It returns an empty optional when the mapping function returns null.

ellisjoe: nice

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
ellisjoe: that returns `UriInfo` rather than `ExtendedUriInfo` which doesn't have the matched model resource

uschi2000 (approved):

PR #546, opened 2017-09-15
make sure all log messages are safe

pnepywoda:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RetryInterceptor.java
```diff
@@ -60,7 +61,7 @@ public Response intercept(Chain chain) throws IOException {
                 }
             }
             if (!successful) {
-                log.error("Request to url {} failed on attempt {}", request.url(), numTries);
+                log.error("Request to url {} failed on attempt {}", request.url(), SafeArg.of("numTries", numTries));
```
markelliot: log message should get cleaned up to:

    "Failed request to url"

tracing/src/main/java/com/palantir/remoting3/tracing/AsyncSpanObserver.java
```diff
@@ -75,7 +76,7 @@ public void onFailure(Throwable error) {
             });
         } else {
             log.trace("Failed to notify span observer since the maximum number of allowed concurrent observations was "
-                    + "exceeded: {}", maxInflights);
+                    + "exceeded: {}", SafeArg.of("maxInflights", maxInflights));
```
markelliot: Log message should get cleaned up to remove braces.

markelliot (changes requested):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RetryInterceptor.java
pnepywoda: why? don't like the unsafe url being logged?

tracing/src/main/java/com/palantir/remoting3/tracing/AsyncSpanObserver.java
pnepywoda: why? it'll log the safe arg there

markelliot: per the spec, the substitution never happens/the args are named and shown along side.

markelliot (approved):

uschi2000 (approved):

uschi2000:
try `ag '{}'`, there's still a few bad ones

pnepywoda:
Sure but they'll be unsafe args

PR #547, opened 2017-09-18
Upgrade remoting-api, add ClientConfiguration#backoffSlotSize

uschi2000:


http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
```diff
@@ -64,6 +64,12 @@
     /** The maximum number of times a failed request is retried. */
     int maxNumRetries();
 
+    /**
+     * The size of one backoff time slot for call retries. For example, an exponential backoff retry algorithm may
+     * choose a backoff time in {@code [0, backoffSlotSize^c]} for the c-th retry.
```
bavardage, line 69: same thing again, the 2^c

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -40,6 +40,7 @@
     private static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
     private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(10);
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(10);
+    private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(500);
```
bavardage, line 43: meh, sure - if we were to change this value, how 'breaking' a change would this be? Think likely will want to tweak this as we learn more...

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
uschi2000: done

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
uschi2000: Well, we have to set some default. How we are going to change it over time ... I don't know. It could be considered a breaking change, but users can always opt to give an explicit value.

PR #548, opened 2017-09-18
OkHttpClients uses ExponentialBackoff instead of NoDelayBackoff

uschi2000:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/ExponentialBackoff.java
```diff
@@ -16,23 +16,34 @@
...
-        return numTimesRetried <= maxNumRetries
-                ? Optional.of(Duration.ZERO)
-                : Optional.empty();
+        retryNumber += 1;
+        if (retryNumber > maxNumRetries) {
+            return Optional.empty();
+        }
+
+        int upperBound = (int) Math.pow(2, retryNumber);
+        return Optional.of(backoffSlotSize.multipliedBy(random.nextInt(upperBound)));
```
schlosna, line 51: do we want [decorrelated jitter](https://www.awsarchitectureblog.com/2015/03/backoff.html)?

bavardage: > Of the jittered approaches, “Equal Jitter” is the loser. It does slightly more work than “Full Jitter”, and takes much longer. The decision between “Decorrelated Jitter” and “Full Jitter” is less clear. The “Full Jitter” approach uses less work, but slightly more time. Both approaches, though, present a substantial decrease in client work and server load.

so this, as implemented, is full jitter, right?

schlosna: roughly yes. mostly just want us to define/document the intended algorithm. I think in most cases we're probably happy to do more work if that reduces time.

uschi2000: i think in the grander scheme of things, it probably doesn't matter for today. i added some docs.

bavardage (approved):

PR #549, opened 2017-09-19
ensure metrics are registered

ellisjoe:


okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -0,0 +1,53 @@
...
+        SslConfiguration sslConfig = SslConfiguration.of(Paths.get("src/test/resources/trustStore.jks"));
+        ClientConfiguration conf = ClientConfigurations.of(
+                ImmutableList.of("http://localhost"),
+                SslSocketFactories.createSslSocketFactory(sslConfig),
+                SslSocketFactories.createX509TrustManager(sslConfig));
+        OkHttpClients.create(conf, "test", OkHttpClients.class);
+
+        SortedMap<String, Meter> meters = SharedMetricRegistries.getDefault().getMeters();
+
+        assertThat(meters.get("com.palantir.remoting3.okhttp.OkHttpClients.response.family.informational")).isNotNull();
```
ellisjoe: @uschi2000 main point of this is to figure out if these metric names are the right thing

schlosna: My guess is that you'll probably want to extract out the class portion of the metric name unless we're just doing blind search replace of `s/remotingN/remotingN+1/g` across the project each release

ellisjoe: switched to `java.lang.String` since the class name doesn't matter for the test

uschi2000 (approved):

PR #560, opened 2017-09-25
make ProxySelector creator util method public

onursatici:
Given a `PemX509Certificate`, there is no convenient way of creating a `JaxRsClient` with proxy configuration without re-implementing `createProxySelector`

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
+++ b/http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
@@ -86,7 +86,7 @@ public final class ClientConfigurations {
                 .build();
     }
 
-    private static ProxySelector createProxySelector(Optional<ProxyConfiguration> proxyConfig) {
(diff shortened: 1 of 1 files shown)


http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -86,6 +86,28 @@ public static ClientConfiguration of(
                 .build();
     }
 
+    /**
+     * Creates a new {@link ClientConfiguration} instance from the given SSL configuration, URIs and proxy
+     * configuration, filling in all other configuration with the defaults specified as constants in this class.
+     */
+    public static ClientConfiguration of(
```
uschi2000: can you not use the builder?

onursatici: Yeah I can also use the ClientConfiguration builder if the `createProxySelector` is public. I can either make that public, or add something like this

onursatici:
@uschi2000 ready for review

uschi2000 (approved):

PR #571, opened 2017-10-04
Generate only 63 bit random IDs in tracing

### LINKED ISSUE #570: Only generate 63 bits of traceID
Generating random IDs for tracing currently generates an entire 64-bit long. This isn't as portable as generating 63 random bits which is agnostic of signedness

https://github.com/palantir/http-remoting/blob/develop/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java#L35

chewr:
Generating a random ID without a sign bit ensures portability across consumers that may expect signed or unsigned IDs from traces and spans.

Fixes #570

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
@@ -32,7 +32,10 @@ public final class Tracers {
 
     /** Returns a random ID suitable for span and trace IDs. */
     public static String randomId() {
-        return longToPaddedHex(ThreadLocalRandom.current().nextLong());
+        long signedLong = ThreadLocalRandom.current().nextLong();
+
+        // clear out sign bit
+        return longToPaddedHex(signedLong & Long.MAX_VALUE);
     }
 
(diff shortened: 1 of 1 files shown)


schlosna:
Curious what specific consumers may not handle a full 64-bit value that we have to drop the MSB.

chewr:
zipkin-go-opentracing for one. TraceID is exposed as a signed int64 here https://github.com/openzipkin/zipkin-go-opentracing/blob/master/thrift/gen-go/zipkincore/zipkinCore.go#L798

You can see here that only the 63 less significant bits are generated randomly for IDs involved in tracing: https://github.com/openzipkin/zipkin-go-opentracing/blob/master/util.go#L18

markelliot:
Signed int64s still reasonably hold all the bits, it’s not clear why the openzipkin implementation would drop the MSB. Does the spec say that it should be unsigned? If so, the change here is wrong, if not, the change here is irrelevant. Either way, suspect a bug in openzipkin.

uschi2000:
Is this still relevant?

chewr:
We worked around this in the zapper library

PR #575, opened 2017-10-12
Basic fix for tracing data redaction issues

j-baker:
Given that the name is unsafe here, and the name will be correct in the spans on the server, we might get away with just omitting it here?

### DIFF
--- a/tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
+++ b/tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
@@ -32,7 +32,7 @@ public enum OkhttpTraceInterceptor implements Interceptor {
     @Override
     public Response intercept(Chain chain) throws IOException {
         Request request = chain.request();
-        OpenSpan span = Tracer.startSpan(request.method() + " " + request.url(), SpanType.CLIENT_OUTGOING);
+        OpenSpan span = Tracer.startSpan("remote call to redacted url", SpanType.CLIENT_OUTGOING);
         Request.Builder tracedRequest = request.newBuilder()
                 .addHeader(TraceHttpHeaders.TRACE_ID, Tracer.getTraceId())
                 .addHeader(TraceHttpHeaders.SPAN_ID, span.getSpanId())
--- a/tracing-okhttp3/src/test/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptorTest.java
+++ b/tracing-okhttp3/src/test/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptorTest.java
@@ -145,7 +145,6 @@ public final class OkhttpTraceInterceptorTest {
         OkhttpTraceInterceptor.INSTANCE.intercept(chain);
         verify(observer).consume(spanCaptor.capture());
         Span okhttpSpan = spanCaptor.getValue();
-        assertThat(okhttpSpan.getOperation()).isEqualTo("GET http://localhost/");
         assertThat(okhttpSpan).isNotEqualTo(outerSpan);
     }
 


markelliot:
I think the fix we originally had made for this on the receiving side was to use the unsubstituted url, since that's known to be safe and conveys a meaningful amount of info.

markelliot:
not sure if that's available at this point in the request chain

uschi2000:
We'd have to implement this at the Feign level, not the OkHttp level.

j-baker:
Cool yeah, so basically talked to @ellisjoe about this and my understanding was: We can't really make much benefit from tracing code at the moment because the logs might contain unsafe information, and both @ellisjoe and @uschi2000 investigated it and couldn't find a 'nice' place that wouldn't involve replicating logic in Feign and Retrofit etc.

Figured we should just merge this pending finding the right place to do it, because tracing > no tracing.

This information should exist on the server side of the span anyway.

markelliot:
More than once have wanted to know at the OkHttp level what the URL template was. Does OkHttp have a way for us to pass that in as metadata on the request object?

On Oct 16, 2017, at 23:06, James Baker <notifications@github.com<mailto:notifications@github.com>> wrote:


Cool yeah, so basically talked to @ellisjoe[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_ellisjoe&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=Km3h6mvmBmFVuTrOZqmJrOXJ7ge84xKMmE03qNnjNU4&s=CJU8gdrSZRR5DJ-qnzsOhN9pO2xuAtSmmwTN6fAqLzM&e=> about this and my understanding was: We can't really make much benefit from tracing code at the moment because the logs might contain unsafe information, and both @ellisjoe[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_ellisjoe&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=Km3h6mvmBmFVuTrOZqmJrOXJ7ge84xKMmE03qNnjNU4&s=CJU8gdrSZRR5DJ-qnzsOhN9pO2xuAtSmmwTN6fAqLzM&e=> and @uschi2000[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_uschi2000&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=Km3h6mvmBmFVuTrOZqmJrOXJ7ge84xKMmE03qNnjNU4&s=bASVgYmovqg7_SPZP8s-G4m_B777nPsKBFHp_-FIRvM&e=> investigated it and couldn't find a 'nice' place.

Figured we should just merge this pending finding the right place to do it, because tracing > no tracing.

uschi2000:
It does, but as far as I remember you need to unwrap/rewrap a lot of Feign code to access the bit where the OkHttp call is constructed.

uschi2000:
I can probably finish this today or tomorrow.

uschi2000 (approved):

uschi2000:
Would to really hack Feign (either wrap all proxies or inject&remove HTTP headers) to do the right thing. We're removing the name from the client spans with this PR and rely on the fact that the server span contains the target API name via TraceEnrichingFilter.

uschi2000:
Thanks, James.

tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
```diff
@@ -32,7 +32,7 @@
     @Override
     public Response intercept(Chain chain) throws IOException {
         Request request = chain.request();
-        OpenSpan span = Tracer.startSpan(request.method() + " " + request.url(), SpanType.CLIENT_OUTGOING);
+        OpenSpan span = Tracer.startSpan("remote call to redacted url", SpanType.CLIENT_OUTGOING);
```
gracew, line 35: I thought in our internal spec we agreed for this to be the method type, e.g. GET, POST, etc.

uschi2000: fixed

PR #576, opened 2017-10-17
Remove the client span creation

k-simons:
- Happy to close this if we want a more full-fledged solution as mentioned at the bottom of https://github.com/palantir/http-remoting/pull/575
- However curious to hear thoughts on instead of trying to determine the unsubstituted path at the interceptor layer, just no longer create a span on out going client requests
- We will still propagate the tracing headers for the current open span if it exists
- On the incoming server side, determine the unsubstituted path. I believe that currently it will still be an unsafe span name that has substituted path params
- Main thought here is that instead of creating a span for rpc requests on both the client and the incoming server, just create it at the incoming service. There is a very good change that an open span will already exist on the client service. 
- Lose some fidelity in being able to cleanly view network overhead between rpc calls and also lose originating service on rpc call if there is not a currently open span
- More than happy to close this pr if this is the wrong approach, however would still want to salvage some changes to TraceEnrichingFilter.java to close out https://github.com/palantir/http-remoting/issues/566 . Did not fix tests

### DIFF
--- a/jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
+++ b/jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
@@ -17,12 +17,14 @@
 package com.palantir.remoting3.servers.jersey;
 
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import com.palantir.remoting.api.tracing.Span;
 import com.palantir.remoting.api.tracing.SpanType;
 import com.palantir.remoting.api.tracing.TraceHttpHeaders;
 import com.palantir.remoting3.tracing.Tracer;
 import com.palantir.remoting3.tracing.Tracers;
 import java.io.IOException;
+import java.util.List;
 import java.util.Optional;
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerRequestFilter;
@@ -33,6 +35,7 @@ import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
 import org.glassfish.jersey.server.ExtendedUriInfo;
 import org.glassfish.jersey.server.model.Resource;
+import org.glassfish.jersey.uri.UriTemplate;
 
 @Provider
 public final class TraceEnrichingFilter implements ContainerRequestFilter, ContainerResponseFilter {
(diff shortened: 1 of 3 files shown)


jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
```diff
@@ -49,7 +52,8 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                 .map(Resource::getPath)
                 .orElse("(unknown)");
 
-        String operation = requestContext.getMethod() + " " + path;
+        String operation = requestContext.getMethod() + " "
```
uschi2000, line 55: I don't understand this change. `path` is now unused?

tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
```diff
@@ -32,22 +29,10 @@
...
-
-        Response response;
-        try {
-            response = chain.proceed(tracedRequest.build());
-        } finally {
-            Tracer.completeSpan();
-        }
-
-        return response;
+        Request.Builder tracedRequest = request.newBuilder();
```
uschi2000, line 32: This change means that we're losing the root span, don't think that's the desired behavior.

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java
k-simons: Ah yes I missed https://github.com/palantir/http-remoting/pull/544, closing the PR

tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
k-simons: To clarify, if the first server already had in an incoming request (which most would), that span would be the root span, so you would be losing the span between the root span the first span on the 2nd service? Not the root span?

uschi2000: If the if the server has one: yes. But not all traces originate from servers.

PR #577, opened 2017-10-17
Feign propagated URI templates for tracing

markelliot:


uschi2000:
Yep, this is the hack I had discussed with James, too. Mark, are you sending this is a proof of concept, or do you you think we should implement this? In the latter case, also need Retrofit.

markelliot:
Had some time this morning and implemented the hack you proposed.

markelliot:
(so that we could merge)

markelliot:
Retrofit doesn't have a great way to intercept the call before it's passed to OkHttp, so what I'd propose instead of including the solution here, is that we use the `@Headers` annotation to where we generate Retrofit interfaces such that the path is added as a header in the 'normal' retrofit scheme. The OkHttp client change in this PR will result in that, then, safely being included in the trace logs.

Cool?

uschi2000:
Uh, that's quite convoluted :( Is that really worth the effort, given that we have the API endpoint name in the corresponding server span already?

markelliot:
The vast majority of the callers we use are Feign-based, even with Retrofit this is a strict improvement for those callers and no worse for Retrofit. If the Retrofit workaround is too gross, perhaps we can take this just as upside on Feign, for now.

markelliot:
...am still digging around Retrofit, think maybe there's an option to fix this

markelliot:
...j/k, there's not really a good pre-request Retrofit interceptor, so we'd have to duplicate and slightly tweak a fairly chunky bit of code.

If you don't want to take the Feign upside alone, you can close the PR.

uschi2000:
OK, let's do it. Want to add a simple test?

tracing-okhttp3/src/main/java/com/palantir/remoting3/tracing/okhttp3/OkhttpTraceInterceptor.java
```diff
@@ -29,10 +29,20 @@
 public enum OkhttpTraceInterceptor implements Interceptor {
     INSTANCE;
 
+    private static final String HTTP_REMOTING_HEADER = "hr-path-template";
```
uschi2000: let's call this PATH_TEMPLATE_HEADER, and reuse in the contract

uschi2000: (not sure you can reuse...)

markelliot:
ok, all set.

markelliot:
something with GH/circle seems messed up, though

uschi2000 (approved):

j-baker:
Depressingly this does not work. Basically because the template URL is in a template header, Feign just templates everything into it.

PR #582, opened 2017-10-24
A few perf improvments for Tracing

j-baker:
I want Tritium to trace by default, but the overhead seemed a little excessive (700ns per method invocation).

Here, we have 3 improvements:

First, don't iterate over a hashmap.

Second, handcoded 'toString' method.

Third, 'completeTraceFast' which avoids a bunch of computation if I'm
not going to use it (e.g. a System.nanoTime call which is 100 cycles
plus).

For changes (with the Tritium benchmark):

Unchanged: 711ns
With iterating over the list not the hashmap: 666ns
With the custom toHexString: 588ns

With the shortcircuit and a 1% sampling collector:

After all changes but shortcircuit: 256ns
With shortcircuit only: 220ns
After all changes: 198ns

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
@@ -18,8 +18,6 @@ package com.palantir.remoting3.tracing;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
 import com.palantir.remoting.api.tracing.OpenSpan;
 import com.palantir.remoting.api.tracing.Span;
 import com.palantir.remoting.api.tracing.SpanObserver;
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
@@ -17,12 +17,16 @@
 package com.palantir.remoting3.tracing;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.palantir.remoting.api.tracing.OpenSpan;
 import com.palantir.remoting.api.tracing.Span;
 import com.palantir.remoting.api.tracing.SpanObserver;
 import com.palantir.remoting.api.tracing.SpanType;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.slf4j.Logger;
@@ -48,8 +52,8 @@ public final class Tracer {
         return trace;
     });
 
-    // Thread-safe Map implementation
-    private static final Map<String, SpanObserver> observers = Maps.newConcurrentMap();
+    private static final Map<String, SpanObserver> observers = new HashMap<>();
+    private static List<SpanObserver> observersList = ImmutableList.of();
 
     // Thread-safe since stateless
     private static TraceSampler sampler = AlwaysSampler.INSTANCE;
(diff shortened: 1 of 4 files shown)


tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -48,8 +52,8 @@ private Tracer() {}
         return trace;
     });
 
-    // Thread-safe Map implementation
-    private static final Map<String, SpanObserver> observers = Maps.newConcurrentMap();
+    private static final Map<String, SpanObserver> observers = new HashMap<>();
```
schlosna: `@GuardedBy("this")`

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -48,8 +52,8 @@ private Tracer() {}
...
+    private static List<SpanObserver> observersList = ImmutableList.of();
```
schlosna: `observersList` needs to be volatile as we're currently only inducing barrier on write side but not reads. overhead of volatile read should be fine on x86

j-baker: To me it's surprising that we'd expect people to register subscribers concurrently

j-baker: I went with the convention of this class, which is to not do this (e.g. the sampling mode) - but sure, will fix both cases.

j-baker: and yeah, looks like volatile reads are very cheap in x86 provided little contention (which we don't expect here).

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
j-baker: that annotation is currently not on the classpath :( I added a comment to the effect given that GuardedBy isn't really read by anything.

schlosna: That's sad as error-prone will do some validations when annotated: https://github.com/google/error-prone/blob/master/docs/bugpattern/GuardedBy.md

j-baker:
Btw, those overheads include the cost of the dynamic proxy - the additional overhead of the tracing are now low as opposed to high.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -123,23 +126,40 @@ private static OpenSpan startSpanInternal(String operation, SpanType type) {
...
      * Like {@link #completeSpan()}, but adds {@code metadata} to the current span being completed.
      */
     public static Optional<Span> completeSpan(Map<String, String> metadata) {
-        return completeSpanInternal(metadata);
+        return completeSpanInternal(metadata, true);
     }
 
-    private static Optional<Span> completeSpanInternal(Map<String, String> metadata) {
+    private static Optional<Span> completeSpanInternal(Map<String, String> metadata, boolean mustConstructResult) {
         Optional<OpenSpan> maybeOpenSpan = currentTrace.get().pop();
```
markelliot: so let's make:

    private static Optional<OpenSpan> popCurrentSpan() {
        return currentTrace.get().pop();
    }

Then:

    private static void fastCompleteSpan() {
        popCurrentSpan();
    }

and `completeSpan` and `completeSpanInternal` should largely merge back together.

markelliot: this also lets you make `fastCompleteSpan(Map<String, String> garbage)` something that can ignore `garbage`.

markelliot: also, why does `fastCompleteSpan(Map<String, String>)` exist as a thing in this construct?

j-baker: fastCompleteSpan doesn't just want to do this though - it also needs to talk to subscribers

j-baker: however your meta-point makes sense and I'll refactor the code into something that makes more sense

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -123,46 +126,71 @@ private static OpenSpan startSpanInternal(String operation, SpanType type) {
...
+
+    /**
+     * Like {@link #fastCompleteSpan()}, but adds {@code metadata} to the current span being completed.
+     */
+    public static void fastCompleteSpan(Map<String, String> metadata) {
+        popCurrentSpan()
+                .filter(openSpan -> currentTrace.get().isObservable())
+                .map(openSpan -> toSpan(openSpan, metadata))
+                .ifPresent(span -> {
+                    for (SpanObserver observer : observersList) {
```
markelliot: turn this for loop into a function?

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -48,11 +50,12 @@ private Tracer() {}
         return trace;
     });
 
-    // Thread-safe Map implementation
-    private static final Map<String, SpanObserver> observers = Maps.newConcurrentMap();
+    // Only access in a class-synchronized fashion
+    private static final Map<String, SpanObserver> observers = new HashMap<>();
+    private static volatile List<SpanObserver> observersList = ImmutableList.of();
```
uschi2000: add comment here to explain why we keep the observersList

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -36,7 +37,24 @@ public static String randomId() {
     }
 
     static String longToPaddedHex(long number) {
```
uschi2000: worth adding simple docs now that the implementation is more cryptic than `padStart(toHexString)`

markelliot (approved):

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -218,7 +218,6 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
 
     @Test
     public void testTraceIdGeneration() throws Exception {
-        assertThat(Tracers.randomId()).hasSize(16); // fails with p=1/16 if generated string is not padded
```
markelliot: Don’t we want to keep this?

uschi2000: yep, let's keep it. functionality should be covered by the deterministic ones below, but it's good to keep  of as a piece of documentation

markelliot (approved):

PR #583, opened 2017-10-24
Actually fix tracing...

j-baker:
@markelliot @uschi2000


j-baker:
The code actually still logs unsafe params before this change

ellisjoe (approved):

PR #584, opened 2017-10-24
canonical UserAgent

ellisjoe:


http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+                .addAgents(agent)
+                .build();
+    }
+
+    @Value.Immutable
+    @ImmutablesStyle
+    interface Agent {
+        Pattern SERVICE_NAME_REGEX = Pattern.compile("([a-zA-Z][a-zA-Z0-9\\-]*)");
+        Pattern INSTANCE_ID_REGEX = SERVICE_NAME_REGEX;
+        Pattern VERSION_REGEX = Pattern.compile("([0-9][a-z0-9\\-\\.]*)");
```
ellisjoe: unsure what we want to restrict these to exactly

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+                .serviceName(serviceName)
+                .version(version)
+                .build();
+        return ImmutableUserAgent.builder()
+                .addAgents(agent)
+                .build();
+    }
+
+    @Value.Default
+    default UserAgent append(String serviceName, String version) {
```
lynchmattj-zz: Wonder if this should be a static utility so the use is more clear.  Append in this case could suggest that this could mutate the original.  Could lead to bugs in consumers.

lynchmattj-zz: so something like: `UserAgent.concat(UserAgent originalAgent, String serviceName, String version)`

ellisjoe: I'm personally in favor of this given that the `UserAgent` is immutable. Willing to change though if another person finds `UserAgent.append(userAgent, service, version)` more clear than `userAgent.append(service, version)`.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+ * Constructs, validates, and formats a canonical User-Agent header. Because the http header spec
+ * (https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) requires headers to be joined on commas, individual
+ * {@link Agent} header strings must never contain commas.
+ */
+@Value.Immutable
+@ImmutablesStyle
+public interface UserAgent {
+    List<Agent> agents();
+
+    @Value.Lazy
```
markelliot: I don’t know that we actually want this to be lazy

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+    default String headerFormat() {
+        return Joiner.on(", ").join(Lists.transform(agents(), Agent::headerFormat));
```
markelliot: I think we may want to sort this?

lynchmattj-zz: The intent here is to join them in the order they were added.  How would you sort them?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
lynchmattj-zz: should we use this? https://github.com/palantir/sls-packaging/blob/develop/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+        Pattern SERVICE_NAME_REGEX = Pattern.compile("([a-zA-Z][a-zA-Z0-9\\-]*)");
+        Pattern INSTANCE_ID_REGEX = SERVICE_NAME_REGEX;
+        Pattern VERSION_REGEX = Pattern.compile("([0-9][a-z0-9\\-\\.]*)");
+
+        String serviceName();
+        Optional<String> instanceId();
+        String version();
+
+        @Value.Lazy
+        default String headerFormat() {
```
uschi2000, line 94: should we implement toString instead?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
uschi2000: what's the purpose of having multiple agent headers? one for "http-remoting (3.3.3)" and one for "foo-service/node123 (5.6.7)"? In that case, not sure sorting has a lot of upside / it's unclear how to sort.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
uschi2000: can use the regex, but please don't add a dependency

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,109 @@
...
+            StringBuilder builder = new StringBuilder(serviceName());
+
+            instanceId().ifPresent(id -> builder.append("/").append(id));
+
+            builder.append(" (").append(version()).append(")");
+            return builder.toString();
+        }
+
+        @Value.Check
+        default void check() {
```
diogoholanda, line 104: The error messages here should probably contain the regexes.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ash211: Sometimes calls go through a layer of libraries and it's useful to have versions of each encoded.  For example right now we have something like: `contour (1.2.3); foundry-spark (4.5.6); foundry-fsclient (7.8.9)`

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
diogoholanda: Personally i wouldn't even have this method. I do think Matt has a point around appending having a mutability sound to it.

ellisjoe: done

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: done

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: changed to `Derived`

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: leaving as is

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: could change this to an abstract class and do that. not sure it makes things more clear though?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: not including regexes matches what we do elsewhere:
https://github.com/palantir/resource-identifier/blob/develop/src/main/java/com/palantir/ri/ResourceIdentifier.java#L246
https://github.com/palantir/sls-packaging/blob/develop/sls-versions/src/main/java/com/palantir/slspackaging/versions/SlsProductVersions.java

ellisjoe:
@uschi2000 I stuck this in the same package as the client configuration. might be worth moving into  the remoting-api repo though?

uschi2000 (approved):

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -0,0 +1,123 @@
...
+            checkArgument(SERVICE_NAME_REGEX.matcher(serviceName()).matches(),
+                    "Illegal service name format: %s", serviceName());
+            if (instanceId().isPresent()) {
+                checkArgument(INSTANCE_ID_REGEX.matcher(instanceId().get()).matches(),
+                        "Illegal instance id format: %s", instanceId().get());
+            }
+            checkArgument(isOrderableVersion(version()), "Illegal version format: %s", version());
```
uschi2000, line 111: calling out that depending on where/how this class will get used, failing here might to too harsh and it could be more appropriate to log a warning and fill in a dummy version (0.0.0).

ash211: I think this will break integration tests that run with versions of 1.0.0-1-g123abc since that's not orderable.  So this wouldn't be usable in integration tests

PR #586, opened 2017-10-25
add tags to client metrics

ellisjoe:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -16,26 +16,58 @@
...
+    }
+
+    public static Optional<Meter> getMeter(
+            MetricRegistry registry, String serviceName, String hostname, String family) {
+        SortedMap<String, Meter> meters = registry.getMeters((name, metric) -> {
+            TaggedMetric taggedMetric = TaggedMetric.from(name);
+            return taggedMetric.name().equals("client.response")
+                    && taggedMetric.tags().get(SERVICE_NAME_KEY).equals(serviceName)
+                    && taggedMetric.tags().get(HOSTNAME_KEY).equals(hostname)
+                    && taggedMetric.tags().get(FAMILY_KEY).equals(family);
```
schlosna: is this method just `@VisibleForTesting`?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -16,26 +16,58 @@
...
+import com.palantir.tritium.tags.TaggedMetric;
+import java.util.Map;
+import java.util.Optional;
+import java.util.SortedMap;
 
 public final class HostMetrics {
 
+    private static final String SERVICE_NAME_KEY = "service-name";
+    private static final String HOSTNAME_KEY = "hostname";
+    private static final String FAMILY_KEY = "family";
```
schlosna: should we expose these as either public constants or via static method (to allow changes so the strings aren't embedded in dependencies)?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/metrics/HostMetricsRegistryTest.java
```diff
@@ -29,8 +29,8 @@ public void test() {
         HostMetricsRegistry hostRegistry = new HostMetricsRegistry(registry, "service");
 
         // TODO(jellis): check different meter when host is added as a tag
```
schlosna: do you want to add the test for alternate host tag?

ellisjoe: yeah the test was right. removed the todo

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
ellisjoe: yeah. it's probably fine to expose publicly but if we end up adding another tag at some point this could potentially break

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
ellisjoe: exposed publicly now

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -16,26 +16,58 @@
...
+    private static String name(String serviceName, String hostname, String family) {
+        Map<String, String> tags = ImmutableMap.<String, String>builder()
+                .put(SERVICE_NAME_TAG, serviceName)
+                .put(HOSTNAME_TAG, hostname)
+                .put(FAMILY_TAG, family)
+                .build();
+        return TaggedMetric.toCanonicalName("client.response", tags);
+    }
+
+    public static Optional<Meter> getMeter(
```
uschi2000: this is used in tests only, right? would prefer to move it into the test scope

ellisjoe: moved into `HostMetricsTest`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -16,26 +16,58 @@
...
 
-    public HostMetrics(MetricRegistry registry, String name, String hostname) {
-        // TODO(jellis): #581 add hostname as a tag
-        informational = registry.meter(MetricRegistry.name(name, "response", "family", "informational"));
-        successful    = registry.meter(MetricRegistry.name(name, "response", "family", "successful"));
-        redirection   = registry.meter(MetricRegistry.name(name, "response", "family", "redirection"));
-        clientError   = registry.meter(MetricRegistry.name(name, "response", "family", "client-error"));
-        serverError   = registry.meter(MetricRegistry.name(name, "response", "family", "server-error"));
-        other         = registry.meter(MetricRegistry.name(name, "response", "family", "other"));
+    public HostMetrics(MetricRegistry registry, String serviceName, String hostname) {
```
uschi2000: add some docs please: `Constructs a metrics registry that records metrics for calls from the given service to the given host.`

uschi2000: nit: camelCase is inconsistent between serviceName and hostname

uschi2000 (approved):

PR #587, opened 2017-10-26
Cancel requests when thread is interrupted (for JaxRS)

uschi2000:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MultiServerRetryInterceptor.java
```diff
@@ -46,6 +47,8 @@ public Response intercept(Chain chain) throws IOException {
         // Try original request first.
         try {
             return chain.proceed(chain.request());
+        } catch (InterruptedIOException e) {
```
uschi2000: where would this exception get thrown?

CRogers:
Synchronous conversation in #pr-587

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MultiServerRetryInterceptor.java
CRogers: This gets thrown by OkHttp if you interrupt the thread before OkHttp tries to connect to the socket. Currently, http-remoting will not release the thread, which is sadness. However, the window for this to happen is very small - I've updated the PR to reflect a more serious/likely scenario, which is when the thread is interrupted while waiting for a socket response. See the failing test.

CRogers: (Rethrowing the InterruptedIOExceptions does not solve the "interrupted while waiting for a socket repsonse" problem unfortunately - diving into the code, it looks like it's blocking on socket operations by OkHttp)

CRogers:
@uschi2000 Any chance you can take a look at this?

CRogers:
Looks like some of the tests are hanging (such as `Java8OptionalAwareDecoderTest.testOptionalThrowsNotAuthorized`). Seems like the async callbacks aren't being triggered even deeper than that changes that this PR (`QosIoExceptionAwareCall` for example). My suspicion is the async `enqueue` code path is now being triggered where as before we only used the synchronous `execute` code path. I reckon it may not have been tested that much before and might not work?

CRogers:
Talked to @j-baker and apparently you must throw `IOException`s from OkHttp `Interceptor`s if you want the async codepath to work. `SerializableErrorToExceptionConverter` and friends throw `RuntimeException`s, which is probably causing this to all go wrong.

CRogers:
See the [handle-interrupts-more-threads](https://github.com/palantir/http-remoting/compare/handle-interrupts-more-threads?expand=1#diff-07eb3e409f7215a8a8b71594d6a0b47bR32) branch for a method that uses an extra thread to continue using the synchronous code path, however I'm not so confident in this code.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -178,7 +178,8 @@ private QosIoExceptionAwareOkHttpClient(OkHttpClient delegate, Supplier<QosIoExc
 
         @Override
         public Call newCall(Request request) {
-            return new QosIoExceptionAwareCall(getDelegate().newCall(request), handlerFactory.get());
+            return new InterruptibleCall(
+                    new QosIoExceptionAwareCall(getDelegate().newCall(request), handlerFactory.get()));
```
markelliot, line 182: do we need to layer these the other way so that we exit from the interrupted thread earlier?

markelliot (commented):
Might be missing it, where does the Retrofit setup get adjusted so that the `RemoteIoException` is unwrapped in the same it's unwrapped for Feign setups?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
CRogers: I think since we call the async codepath on the `QosIoExceptionAwareCall` from inside `InterruptibleCall`, then catch the `InterruptedException` while waiting on a future for that async work to finish, it will handle the interruption just as soon as if the layering was the other way. I tested layering them the other way and it does still pass the test, though.

CRogers:
@markelliot I think the retrofit code uses this [`AyncCallTag` hack](https://github.com/palantir/http-remoting/blob/ff589049a5b73e800a835ed90fa552f8033fb19c/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncCallTag.java#L20) to handle the same thing that `RemoteIoException` is handling. @iamdanfox @j-baker [said here](https://github.com/palantir/http-remoting/pull/592#discussion_r147692498) we can get rid of `AsyncCallTag` with `RemoteIoException`s and proposed doing it in a later pure refactor PR.

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
```diff
@@ -44,6 +45,9 @@
     @Rule
     public ExpectedException expectedException = ExpectedException.none();
 
+    @Rule
+    public Timeout timeout = Timeout.millis(2000);
```
j-baker, line 49: why is this rule required?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InterruptibleCall.java
```diff
@@ -0,0 +1,63 @@
...
+                future.set(response);
+            }
+        });
+
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            getDelegate().cancel();
+            Thread.currentThread().interrupt();
+            throw new InterruptedIOException("Thread was interrupted, cancelling call");
```
j-baker, line 53: can we set e as the cause?

CRogers: Doesn't take a cause unfortunately https://docs.oracle.com/javase/8/docs/api/java/io/InterruptedIOException.html

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
CRogers: Looks like https://github.com/palantir/http-remoting/pull/592/commits/8c1d3df0d6863fc64912686ed8715bf53f92d479 "2second timeout to make unit tests fail fast" by @iamdanfox, not sure if we still want it.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InterruptibleCall.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (ExecutionException e) {
+            throw (IOException) e.getCause();
```
j-baker, line 55: anything we can do here with logging to make sure the original stack trace appears? I'd in general suggest doing e.getCause().addSuppressed(e) but this causes issues for logback...

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
CRogers: I think we want to keep it as otherwise the way these tests fail is hanging forever. When this happens, you don't get any info about what failed on circle.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InterruptibleCall.java
CRogers: What issues does this cause for logback? Some other options are to 1) log the ExecutionException, throw the cause or 2) Wrap the ExecutionException in an IOException.

markelliot:
for the AsyncCallTag hack, it sort of sounds like the expectation is the SerializableError bubbles up the stack, but at the OkHttp level that exception is now being wrapped -- the hack will cause the exception to bubble up, but it'll be the wrong one, no?

markelliot:
happy to be wrong, any chance you can point me at the test?

CRogers:
Honestly not entirely sure how the `AsyncCallTag` stuff works - @j-baker and @iamdanfox have more insight.

CRogers:
Do you mean the test for async call tags? Looks like there is a [`SerializableErrorInterceptorTest`](https://github.com/palantir/http-remoting/blob/ff589049a5b73e800a835ed90fa552f8033fb19c/retrofit2-clients/src/test/java/com/palantir/remoting3/retrofit2/SerializableErrorInterceptorTest.java)

markelliot:
I think we want to check that if we make a retrofit call and it throws that we get back a SerializableError – that test would lead me to believe that’s not what we’ll get back.

CRogers:
I've looked around but not sure what the behaviour we want around SerializableErrors. Would be happy for someone else to write the test or to iterate with someone who knows in person.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InterruptibleCall.java
nziebart: Fwiw, okhttp already properly handles interrupts when using http/2

iamdanfox:
@melliot I've opened a pure refactor PR that adds some test coverage to the retrofit codepath as you requested: https://github.com/palantir/http-remoting/pull/623.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InterruptibleCall.java
j-baker: stack overflow exception!

j-baker:
This is done now, right? (i.e. we can close this?)

iamdanfox:
Yes, let's close this out as @uschi2000's okhttp client refactor includes the 'future and immediately get' workaround: https://github.com/palantir/http-remoting/pull/624/files#diff-f2745018f6d1c9428e36cd8a027b43fcR85

PR #591, opened 2017-10-26
Fix bug in the async codepath

j-baker:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -83,7 +84,7 @@ static QosIoExceptionAwareOkHttpClient withStableUris(
     }
 
     private static Supplier<QosIoExceptionHandler> createQosHandler(ClientConfiguration config) {
-        return () -> new AsyncQosIoExceptionHandler(scheduledExecutorService,
+        return () -> new AsyncQosIoExceptionHandler(scheduledExecutorService, Executors.newSingleThreadExecutor(),
```
schlosna: should this be a different executor?

nziebart:
Note that this is not really ideal, because you will take up 1 thread per retry.

If you retry 10 times over the course of 1 minute, due to exponential backoffs or whatever, you will block 10 threads over that course of time.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
```diff
@@ -43,10 +42,15 @@
...
-    private final ListeningScheduledExecutorService executorService;
+    private final ListeningScheduledExecutorService scheduledExecutorService;
+    private final ListeningExecutorService executorService;
     private final BackoffStrategy backoffStrategy;
 
-    AsyncQosIoExceptionHandler(ScheduledExecutorService executorService, BackoffStrategy backoffStrategy) {
+    AsyncQosIoExceptionHandler(
+            ScheduledExecutorService scheduledExecutorService,
+            ExecutorService executorService,
+            BackoffStrategy backoffStrategy) {
```
markelliot, line 53: preconditions check that executor services must not be the same

markelliot (approved):

uschi2000 (approved):

uschi2000 (approved):

markelliot (approved):

PR #592, opened 2017-10-27
Handle interrupts preserve exception API

iamdanfox:
This is just one possible approach to allow @CRogers's "Cancel requests when thread is interrupted" PR (#587) to maintain http-remoting's current API.

### DIFF
--- a/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/RemoteIoExceptionClient.java
+++ b/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/RemoteIoExceptionClient.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting3.jaxrs;
+
+import com.palantir.remoting3.okhttp.RemoteIoException;
+import feign.Client;
+import feign.Request;
+import feign.Response;
+import java.io.IOException;
+
+/**
+ * This client's sole purpose is to turn our special {@link RemoteIoException} into a RuntimeException.
+ * <p>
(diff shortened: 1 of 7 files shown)


jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/RemoteIoExceptionClient.java
```diff
@@ -0,0 +1,47 @@
...
+    public RemoteIoExceptionClient(Client delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public Response execute(Request request, Request.Options options) throws IOException {
+        try {
+            return delegate.execute(request, options);
+        } catch (RemoteIoException e) {
+            throw e.getRuntimeExceptionCause();
```
j-baker, line 44: as much as possible I'd prefer we don't blindly rethrow exceptions; can we do something here to make sure that the stack trace is the right place? Can we leverage the fact we know that this is a RemoteException and wrap it maybe with another identical RemoteException?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/SerializableErrorInterceptor.java
```diff
@@ -40,7 +40,9 @@ public Response intercept(Chain chain) throws IOException {
         if (!(response.isSuccessful() || response.code() == 101)) {
             Collection<String> contentTypes = response.headers("Content-Type");
             InputStream body = response.body().byteStream();
-            throw SerializableErrorToExceptionConverter.getException(contentTypes, response.code(), body);
+            RuntimeException runtimeException = SerializableErrorToExceptionConverter.getException(
+                    contentTypes, response.code(), body);
+            throw new RemoteIoException(runtimeException);
```
j-baker, line 45: yay! this is great. You can actually remove a chunk more complexity here - if you delete AsyncCallTag you can just do something very similar to the rethrow code in the Java8 call adapter and all the pain we've had in the last day disappears.

iamdanfox: Yeah getting rid of AsyncCallTag seems like it would be a nice refactor... might put this in a separate pure-refactor PR later I think

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/RemoteIoExceptionClient.java
iamdanfox: OK so just to confirm I'm understanding: re-throwing the internal 'cause' would lead to confusing stacktraces because it would would omit lines for all this RemoteIoException handling stuff and just point at the original constuction location... you'd prefer that we actually construct a new exception here with the same semantic purpose.

iamdanfox:
I'm gonna merge this into @crogers' branch so that we have one place to iterate and can address all comments there!

PR #593, opened 2017-10-27
Revert "Revert "add tags to client metrics" (#590)"

ellisjoe:
This reverts commit d1e472729a8eb71c93a5bfdaf8c761a2c493e799.


ellisjoe:
opening this to bring back tags. should hold off on merging until we finalize how we're going to serialize tags

markelliot:
Also want to block merging until we have a full release of tritium.

On Oct 27, 2017, at 10:30, Joe Ellis <notifications@github.com<mailto:notifications@github.com>> wrote:


opening this to bring back tags. should hold off on merging until we finalize how we're going to serialize tags

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -18,26 +18,49 @@
...
+        redirection   = registry.meter(name(serviceName, hostname, "redirection"));
+        clientError   = registry.meter(name(serviceName, hostname, "client-error"));
+        serverError   = registry.meter(name(serviceName, hostname, "server-error"));
+        other         = registry.meter(name(serviceName, hostname, "other"));
+    }
+
+    private static MetricName name(String serviceName, String hostname, String family) {
+        return MetricName.builder()
+                .safeName(CLIENT_RESPONSE_METRIC_NAME)
+                .putSafeTags(SERVICE_NAME_TAG, serviceName)
```
ellisjoe: should be appended to `client.response.`

jaxrs-clients/versions.lock
```diff
@@ -189,6 +203,25 @@
...
             ]
         },
+        "com.palantir.tritium:tritium-api": {
+            "locked": "0.7.0",
+            "transitive": [
+                "com.palantir.tritium:tritium-core",
+                "com.palantir.tritium:tritium-metrics"
+            ]
+        },
+        "com.palantir.tritium:tritium-core": {
```
uschi2000: what's the point of tritium-api if I need to depend on this whole transitive beast?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/InstrumentedInterceptor.java
```diff
@@ -16,25 +16,20 @@
...
  * Records metrics about the response codes of http requests.
  */
 public final class InstrumentedInterceptor implements Interceptor {
 
-    private static final Logger log = LoggerFactory.getLogger(InstrumentedInterceptor.class);
-
     private final HostMetricsRegistry hostMetrics;
 
-    InstrumentedInterceptor(MetricRegistry registry, String serviceName) {
+    InstrumentedInterceptor(TaggedMetricRegistry registry, String serviceName) {
```
uschi2000, line 33: TaggedMetricRegistry should really be an interface and part of the tritium-api so that we can avoid transitive dependencies on all the tritium implementation details. http-remoting already being a shared library, those dependencies make me a little uneasy. in theory, after this merges, tritium can no longer change any of its public APIs.

uschi2000: @schlosna , what would it take to make that change?

ellisjoe: unless we plumb the registry all the way through the remoting apis we're still going to have the issue. explicitly the issue is that we use `TaggedMetricRegistry.getDefault()` to get the global registry at the moment.

markelliot: Can do the serviceloader thing to pull the implementation if we need to.

ellisjoe: ugh

uschi2000:
The only hope for that pattern is slf4j style discovery.

uschi2000:
I know it's pedantic, but these chained dependencies can really lead to a world of pain, so I'd prefer taking the smallest possible dependency.

ellisjoe:
In this instance this feels a bit over engineered. The registry is just a wrapper around a map and I don't expect that to change. I'm happy with moving the registry into it's own dependency-less project, but service loading seems like overkill.

ellisjoe:
Also happy to split out an interface, just don't think there's a need to avoid depending on the impl for now.

markelliot:
How do you get the meters and what not without the impl?

uschi2000:
OK, if you're confident that this won't break, go ahead. Can you please add some pretty loud warning signs in tritium that those APIs must not change unless the Java package is incremented?

ellisjoe:
blocked on https://github.com/palantir/tritium/pull/53

schlosna:
[Tritium 0.8.0 released](https://bintray.com/palantir/releases/tritium/0.8.0) with TaggedMetricRegistry interface change and minimized dependencies in `tritium-registry` module

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/InstrumentedInterceptorTest.java
```diff
@@ -42,27 +45,36 @@
...
 
     @Test
     public void testResponseFamilyMetrics() throws IOException {
+        successfulRequest(REQUEST_A);
         interceptor.intercept(chain);
 
-        Meter meter = registry.getMeters().get("client.response.family.successful");
-        assertThat(meter.getCount()).isEqualTo(1);
+        Meter meterA = registry.meter(name("hosta"));
+        assertThat(meterA.getCount()).isEqualTo(1);
```
uschi2000, line 70: nit: check that meter for hostb has count 0

uschi2000: actually, I guess you do that below.

uschi2000 (approved):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -18,26 +18,47 @@
...
+        successful    = registry.meter(name(serviceName, hostname, "successful"));
+        redirection   = registry.meter(name(serviceName, hostname, "redirection"));
+        clientError   = registry.meter(name(serviceName, hostname, "client-error"));
+        serverError   = registry.meter(name(serviceName, hostname, "server-error"));
+        other         = registry.meter(name(serviceName, hostname, "other"));
+    }
+
+    private static MetricName name(String serviceName, String hostname, String family) {
+        return MetricName.builder()
+                .safeName(serviceName + CLIENT_RESPONSE_METRIC_NAME_SUFFIX)
```
schlosna, line 52: why aren't we putting service name in a tag?

uschi2000:
So the diamond dependency danger turns out to be pretty real.
https://github.com/palantir/tritium/issues/54

PR #595, opened 2017-11-01
Refactor UserAgent and UserAgents

uschi2000:


http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
```diff
@@ -26,19 +26,19 @@
...
 
     @Test
     public void testCorrectHeaderFormatWithoutInstanceId() {
         UserAgent baseUserAgent = UserAgent.of("service", "1.0.0");
-        assertThat(baseUserAgent.headerFormat()).isEqualTo("service (1.0.0)");
+        assertThat(UserAgents.format(baseUserAgent)).isEqualTo("service (1.0.0)");
 
-        UserAgent derivedAgent = UserAgent.append(baseUserAgent, "remoting", "2.0.0");
-        assertThat(derivedAgent.headerFormat()).isEqualTo("service (1.0.0), remoting (2.0.0)");
+        UserAgent derivedAgent = UserAgents.merge(baseUserAgent, UserAgent.of("remoting", "2.0.0"));
```
markelliot: This syntax seems way more clunky

markelliot:
also wondering if we should switch serviceName to generally be 'service address' (which would help us locate the origin node)

markelliot:
and in terms of reverse-parsing, if the use of parentheses is the best scheme

markelliot:
sorry for spamming with comments, last thing is we may want to format slightly more like the standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

which expects:

    <product> / <product-version> <comment>

common format for web browsers:

    <browser>/<version> (<system-information>) <platform> (<platform-details>) <extensions>

translating, that might result in:

    <calling service>/<calling service version> [<other detail>/<other detail version>]*

e.g.:

    1.main.foo-service.stack1/1.2.3 http-remoting/3.4.0 client-jar/2.3.4

http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
uschi2000: can make it `base.merge(other)`

uschi2000:
will address the syntax proposal in a separate PR.

http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
uschi2000: done

uschi2000:
@ellisjoe main change here is that version syntax isn't strictly enforced.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -55,69 +49,41 @@ static UserAgent of(String serviceName, String instanceId, String version) {
...
-    static UserAgent append(UserAgent userAgent, String serviceName, String version) {
-        Agent agent = ImmutableAgent.builder()
-                .serviceName(serviceName)
-                .version(version)
-                .build();
+    /**
+     * Returns the {@link UserAgent} comprising all {@link UserAgent#agents} from the left and all {@link
+     * UserAgent#agents} from the right given {@link UserAgent}s.
+     */
+    default UserAgent merge(UserAgent other) {
```
ellisjoe: @lynchmattj @diogoholanda didn't like this the first time around. I'm still in favor of this version

http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
```diff
@@ -26,19 +26,19 @@
     @Test
     public void testCorrectHeaderFormatWithInstanceId() {
         UserAgent baseUserAgent = UserAgent.of("service", "instanceId", "1.0.0");
-        assertThat(baseUserAgent.headerFormat()).isEqualTo("service/instanceId (1.0.0)");
+        assertThat(UserAgents.format(baseUserAgent)).isEqualTo("service/instanceId (1.0.0)");
 
-        UserAgent derivedAgent = UserAgent.append(baseUserAgent, "remoting", "2.0.0");
-        assertThat(derivedAgent.headerFormat()).isEqualTo("service/instanceId (1.0.0), remoting (2.0.0)");
+        UserAgent derivedAgent = baseUserAgent.merge(UserAgent.of("remoting", "2.0.0"));
+        assertThat(UserAgents.format(derivedAgent)).isEqualTo("service/instanceId (1.0.0), remoting (2.0.0)");
```
ellisjoe: why not have the formatting method in `UserAgent`?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
```diff
@@ -34,18 +31,15 @@
 @ImmutablesStyle
 public interface UserAgent {
 
-    List<Agent> agents();
+    String DEFAULT_VERSION = "0.0.0";
```
ellisjoe: `unknown`?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
markelliot: am fine with either, just want to keep the `base.addstuff(<stuff>)` syntax over `UserAgents.merge(base, other)`.

uschi2000: well, it's incomplete because it was missing the `append(service, node, version)` variant.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
uschi2000: need to special-case parsing then. can do.

http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
uschi2000: just want to keep the data objects small

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
markelliot: 0.0.0 is probably fine tbh

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: I don't think `append(service, node, version)` makes sense to append. when would a user agent have more than one `node`?

@markelliot the thing they explicitly didn't like was the non static version of the method

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
ellisjoe: because it doesn't match the expected regex?

markelliot: we're putting a string here to mean "the version is unknown", the word "unknown" and the string "0.0.0" are about equivalent

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
markelliot: oh I see, because we want the agent construction to become more structured.

the non-static version makes sense to me, guess we can wait for one of them to chime in?

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
uschi2000: back to 0.0.0

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgent.java
uschi2000: i'd like to move on here since the qos work is blocking on it and there's still 3-4 PRs and releases that need to trickle through before I can pick up and make those changes.

ellisjoe: Regardless of the static method issue, I'm not sure you actually want to be able to merge two UserAgents. I don't know what having multiple nodeIds means in that case.

uschi2000: ok. then the type should be 
```
class UserAgent {
  Optional<String> node;
  List<Agent> libraries;
}
```

uschi2000: used as:
```
UserAgent agent = UserAgent.of("my-service", "192.168.0.1", "1.2.3")
  .addLibrary(MyApi.class);
  .addLibrary("my-other-lib", "4.5.5");
```
or similar

uschi2000: pretty close to what we had, i guess

ellisjoe: sounds great to me. there was some push back on having UserAgent accept a `Class` and do the manifest business. may just want to have the `add(name, version)` for now

uschi2000: deal

markelliot: sorry, the other option here would be:

    Agent primary;
    List<Agent> informational;

markelliot: ...which also eliminates the dependency on the list ordering

uschi2000: ok, try again.

markelliot (approved):

ellisjoe (approved):

PR #596, opened 2017-11-01
in Tracers add support for wrapping a StreamingOutput

bavardage:
image

(from https://books.google.com/books?id=wzkYAgAAQBAJ&pg=PA76&lpg=PA76&dq=streamingoutput+which+thread&source=bl&ots=dnrF-pOxow&sig=te0F0Aw9qGJ5NYh10-KiRvXqhpE&hl=en&sa=X&ved=0ahUKEwiSocTNn5zXAhVL8CYKHW3VAOAQ6AEIQjAE#v=onepage&q=streamingoutput%20which%20thread&f=false + observed behaviour)

you don't necessarily have control over which thread is being used to stream the response, so can be handy to wrap up your streaming outputs!

This introduces a dependency in the `tracing` project on `javax.ws.rs:javax.ws.rs-api`, but seemed the cleanest place to put it (the tracing stuff needed is package private).


bavardage:
I'm unable to assign a reviewer - @uschi2000 can you take a look?

bavardage:
also, someone able to give me sufficient perms on this repo such that I can assign reviewers?

markelliot (commented):
I think we may want to split out to JaxrsTracers and a separate project so that the tracing library doesn’t pull Jersey.

bavardage:
and we're ok with making `#copyTrace` and `#setTrace` public?

markelliot:
ugh, not really. What if we make an abstract class?

    public abstract class DeferredTracer {

        private final Trace trace;

        public DeferredTracer() {
            this.trace = Tracer.copyTrace();
        }

        public <T> T traced(Callable<T> inner) {
            Trace originalTrace = Tracer.copyTrace();
            Tracer.setTrace(trace);
            do {
                return innner.call();
            } finally {
                Tracer.setTrace(originalTrace);
            }
        }
    }

Then you could, in another package:

    public class TracingAwareStreamingOutput extends DeferredTracer implements StreamingOutput {
        private final StreamingOutput delegate;

        public TracingAwareStreamingOutput(StreamingOutput delegate) {
            super();
            this.delegate = delegate;
        }

        @Override
        public void write(OutputStream output) throws IOException, WebApplicationException {
            traced(() -> delegate.write(output));
        }
    }

It's a little clunky, maybe can rework a bit more.

bavardage:
did what you suggested, but w/o the inheritance

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -177,19 +171,38 @@ public V call() throws Exception {
...
         @Override
         public void run() {
+            deferredTracer.withTrace(() -> {
+                delegate.run();
+                return null;
+            });
+        }
+    }
+
+    public interface ThrowingCallable<T, E extends Throwable> {
```
bavardage: am I being silly/is there some existing interface like this?

markelliot: sadly nothing exists

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -177,19 +171,42 @@ public V call() throws Exception {
...
+        T call() throws E;
+    }
+
+    public static final class DeferredTracer {
```
markelliot: this is great, should likely add a basic javadoc though that shows example usage?

markelliot (approved):
perfect, would be ideal to add javadoc but won't block on it

markelliot (approved):

markelliot:
👍 

also fixed it so you're a collaborator/should be able to assign reviewers, merge, etc.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -177,19 +171,61 @@ public V call() throws Exception {
...
+     * });
+     *
+     * </code>
+     * </pre>
+     */
+    public static final class DeferredTracer {
+        private final Trace trace;
+
+        public DeferredTracer() {
+            this.trace = Tracer.copyTrace();
```
uschi2000: can we inject the trace here rather than grabbing it from static state?

markelliot: the whole point is that the static state is private

markelliot: ...this avoids making Tracer.copyTrace() world public

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -177,19 +171,61 @@ public V call() throws Exception {
...
+     * deferredTracer.withTrace(() -> {
+     *     doThings();
+     *     System.out.println(Tracer.getTraceId()); // prints trace id at time of construction of deferred tracer
+     *     return null;
+     * });
+     *
+     * </code>
+     * </pre>
+     */
+    public static final class DeferredTracer {
```
uschi2000: let's pull this out into it's own file if it's public

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
uschi2000: oh, i see. cool.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
bavardage: fair, done

PR #597, opened 2017-11-01
Use canonical UserAgent format

uschi2000:


uschi2000:
As proposed.

http-clients/src/test/java/com/palantir/remoting3/clients/UserAgentTest.java
```diff
@@ -24,21 +24,21 @@
 public final class UserAgentTest {
 
     @Test
-    public void testCorrectHeaderFormatWithInstanceId() {
-        UserAgent baseUserAgent = UserAgent.of(UserAgent.Agent.of("service", "1.0.0"), "instanceId");
-        assertThat(UserAgents.format(baseUserAgent)).isEqualTo("service/instanceId (1.0.0)");
+    public void testCorrectHeaderFormatWithNodeId() {
+        UserAgent baseUserAgent = UserAgent.of(UserAgent.Agent.of("service", "1.0.0"), "nodeId");
```
markelliot: can we use a nodeId string that's not `nodeId` so that we know this actually gets rendered?

uschi2000: sure

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgents.java
```diff
@@ -42,27 +44,40 @@
...
+     * Formats the given agent in the form {@code name/version (key:value; key:value)}, where the ()-block of comments
+     * is omitted if zero comments are provided.
+     */
+    private static String formatSingleAgent(UserAgent.Agent agent, Map<String, String> comments) {
+        StringBuilder formatted = new StringBuilder()
+                .append(agent.name())
+                .append("/")
+                .append(agent.version());
+
+        String formattedComments = COLON_SEMICOLON_JOINER.join(comments);
```
ellisjoe, line 67: should validate that keys/values don't contain parenthesis at a minimum

uschi2000: this is a private code path, key is always "nodeId" and value goes through through regex.

ellisjoe: ok, seems a bit fragile

uschi2000: will leave a comment.

http-clients/src/main/java/com/palantir/remoting3/clients/UserAgents.java
```diff
@@ -30,8 +31,9 @@
     private static final Logger log = LoggerFactory.getLogger(UserAgents.class);
 
     private static final Joiner COMMA_JOINER = Joiner.on(", ");
```
markelliot, line 33: in the MDN spec, the separator is only space and not comma

ellisjoe: https://tools.ietf.org/html/rfc7231#section-5.5.3

Issue #585, opened 2017-10-24
Remoting clients are not ideal for leader-aware services


nziebart:
Leader-aware services would like a way to redirect clients to the leader and keep them there. However, a given service node will not always know the URI of the leader.

Some example scenarios would be:
(a) A node is partitioned away from the others, and cannot contact any of them to determine who the leader is
(b) A leader election is in progress, and no one is the leader
(c) A quorum of nodes is not yet available, and a leader election cannot make progress until they become available
(d) A node is a follower, and knows who the leader is

Case (d) is pretty easy; we can just issue a 308. However the other cases are not well supported by the current retry logic.

One option would be to always issue a 308 to some random node in any of the above cases. However, this is not great for (b) and (c), because it does not result in any backoff.

If we simply modify the retry logic to try a random node on 503, that would probably suffice.

markelliot:
Perhaps we should instead make our leader aware proxy simply forward requests to the leader?

nziebart:
That would incur an extra network hop, and in rare cases like (a) is not possible.

It would make the client side logic simpler but, in any scenario where we could reliably forward to the leader, it seems like we could also issue a 308 / redirect.

Though I guess we could just return 503 until we know who the leader is, at which point we redirect to it. Though again, it does not solve (a).

nziebart:
Spoke with @uschi2000 offline - proposed fix is just to retry on a random node on 503

PR #599, opened 2017-11-01
Retry on other nodes with 503s

jeremyk-91:
This PR addresses #585 by retrying on other nodes when receiving a 503.

Main concerns:
* I equated "random node" in the discussion on 585 with `UrlSelector.redirectToNext()` which is in an undefined order according to the interface (though it's not in the readme nor in practice). Is this sufficient?
* There were some machinations involved in ensuring that the `UrlSelector` used by the handler and the other filters used by a given okhttp client were the same. Was this actually necessary?
* I don't really like the circular-ish dependency this creates between the `QosIo...Client` and the handler (which needs to know about the Client so it can create different `Call`s), but didn't see a much better way here.
* There isn't quite an analogue to AtlasDB's "fast failover" yet, where we speed through all nodes - we still follow the backoff strategy. Given that AtlasDB TimeLock often is run in 3 or 5 node setups, I think this is fine for now, but if work on GraphLock or other scalability work goes ahead we might need to think of a different solution. I'm separately tracking https://github.com/palantir/atlasdb/issues/2591 for that, though.
* I ran `./gradlew check`, is there something else I should do to verify things work?

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncQosIoExceptionHandler.java
@@ -17,6 +17,7 @@
 package com.palantir.remoting3.okhttp;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
@@ -31,6 +32,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import okhttp3.Call;
+import okhttp3.Request;
 import okhttp3.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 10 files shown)


jeremyk-91:
I don't seem to have permissions to add reviewers - @uschi2000 and @nziebart could you take a look please?

uschi2000:
Thanks, Jeremy. I see how this works, but it's so complicated :( I wonder if we need to rethink the code paths and how they're implemented with OkHttp before making this change. Jeremy, do you have a suggestion for how we could make the end-to-end simpler?

jeremyk-91:
Right, agree that while this works it's pretty complex - I'm not the happiest with this solution either. I think the main thing we need is a way of creating `Call`s or `Request`s apart from making an equal clone (which the call currently knows how to do).

There are 2 routes I can think of that I can explore:
- Have another retry interceptor like the RetryOther one that uses a UrlSelector to determine redirects, though we also need to build the backoff into it. Main concerns are that we want the scheduled executor for retries to have a finite number of threads - we could share them, but things might get messy.
- We know the calls aren't just regular OkHttp calls but are some subclass of them. If we can somehow make a `MultiServerAwareRedirectableCall` which has not just the ability to clone but also to create a copy pointed at a different node, then the exception handler on Unavailables can just use the call's ability to copy itself. Might be a bit messy, but we get to keep the `AsyncQosIoExceptionHandler` a fair bit cleaner.

Feedback on how/if the other two might work would be great! Currently Atlas is having a team surge, but I'll see if I can try out one of these later this week / early next week.

uschi2000:
alternatively, how about wrapping the call from the outside only? at the moment, some stuff is done from the outside, and some stuff is done from the inside (with interceptors). i.e., we'd just look at the Response instead of introspecting some crap in the middle and doing the crazy exception-trigger-executor-scheduled-recursion thing. (this is my bad.)

nziebart:
Yes, I think we can do this by throwing a `QosException.RetryOther` in `QosRetryLaterInterceptor`, and moving all the retry logic into the `Call` wrapper. Proof of concept of this here: https://github.com/palantir/http-remoting/tree/nziebart/503-retry-other (note it's built on top of https://github.com/palantir/http-remoting/pull/604 rather than develop)

nziebart:
Though, we need to think carefully about the relationship between `MultiServerRetryInterceptor`, okhttp's `RetryAndFollowUpInterceptor`, and the new `CallRetryer`. Don't want to get into a situation where we redirect infinitely if, e.g., one node tries to 308/redirect to another node that is down.

(may be an argument for moving `MultiServerRetryInterceptor` to the `Call` level as well)

nziebart:
Actually, re-reading your comment above @uschi2000 I now understand what you mean. Yeah, we should probably have all the logic at the call level.

uschi2000:
The refactor merged and this implementation is now very stale. Before doing more work on this, let's discuss whether this is still relevant, I suspect it is not.

PR #603, opened 2017-11-02
More lenient nodeId syntax for UserAgents

uschi2000:


ellisjoe (approved):

ellisjoe:
what do we expect nodeId to be normally?

uschi2000:
A host name or an IP address

PR #604, opened 2017-11-03
Don't block threads unnecessarily during backoff & retry

nziebart:
Previously we fixed an issue where the backoff & retry logic would recursively submit tasks onto the same executor until it was full, but the solution was to give it a different executor with many threads (the okhttp threadpool). It still recursively submits tasks, so will consume more threads than needed when retrying.

This PR refactors the backoff & retry logic and handles the blocking and async cases separately:
- for blocking calls, just sleep in a loop since the caller is already blocking anyway
- for async calls, schedule an async retry and do not block any threads

The tests got a bit weird as I wanted to test the sync and async codepaths for all cases - happy to work on these more if others feel it's needed.


okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -314,16 +264,59 @@ public void interceptsAndHandlesQos_endToEnd_memorizedCurrentUrlBetweenCalls() t
...
+
+        CallExecutor BLOCKING = call -> call.execute();
+
+        CallExecutor ASYNC = call -> {
+            FutureCallback callback = new FutureCallback();
+            call.enqueue(callback);
+            try {
+                return callback.get();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
schlosna, line 301: Do you want to reset interrupt?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -314,16 +264,59 @@ public void interceptsAndHandlesQos_endToEnd_memorizedCurrentUrlBetweenCalls() t
...
+            } catch (ExecutionException e) {
+                throw (Exception) e.getCause();
```
schlosna, line 303: What if cause is an Error?

schlosna: Or general Throwable?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
nziebart: meh, it's just a test. i'll add it though

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
nziebart: This is a hack because checkstyle doesn't let you declare `throws Throwable` =/

nziebart: But also, the callback interface only accepts `IOException`, so this will always be safe

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -167,28 +135,10 @@ public void throwsProperRemoteExceptionAfterRetry() throws Exception {
...
-                OkHttpClientsTest.class,
-                () -> retryingHandler);
+        OkHttpClient client = createRetryingClient(5);
 
         Call call = client.newCall(new Request.Builder().url(url).build());
-        assertThatThrownBy(call::execute).isInstanceOf(RemoteException.class);
-    }
-
-    @Test
-    public void interceptsAndHandlesQosIoExceptions_wrapsRuntimeExceptionsAsIoExceptions() throws Exception {
```
nziebart, line 181: note: I removed this test, because we want to propagate `RemoteException` as-is and not wrap it. This means that normal `RuntimeException` are also propagated as-is, and I don't think there's any problem with that.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RetryingCall.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.palantir.remoting3.okhttp;
+
+import java.io.IOException;
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.Response;
+
+public final class RetryingCall extends ForwardingCall {
```
nziebart, line 24: Guess there's not even really a reason for this to be a separate class anymore, could just merge with CallRetryer

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/QosAwareCallRetryerTest.java
```diff
@@ -0,0 +1,340 @@
...
+        @Override
+        public AbstractThrowableAssert<?, ? extends Throwable> verifyNoRetryAndThrow(QosIoException exception) {
+            FutureCallback response = new FutureCallback();
+            call.enqueue(response);
+            scheduler.tick(100, TimeUnit.SECONDS);
+            assertThat(response.isDone()).isTrue();
+
+            verify(delegateCall, times(1)).enqueue(any());
+
+            verifyNoMoreInteractions(clonedDelegateCall);
```
nziebart, line 267: Note to self: figure out why this fails if I include 'delegateCall' in this verify

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -49,16 +48,7 @@
...
-     * The {@link ScheduledExecutorService} used for scheduling call retries. This thread pool is distinct from OkHttp's
-     * internal thread pool and from the thread pool used by {@link #executorService}.
-     * <p>
-     * Note: In contrast to the {@link java.util.concurrent.ThreadPoolExecutor} used by OkHttp's {@link
-     * #executorService}, {@code corePoolSize} must not be zero for a {@link ScheduledThreadPoolExecutor}, see its
-     * Javadoc.
-     */
-    private static final ScheduledExecutorService scheduledExecutorService = Tracers.wrap(
-            new ScheduledThreadPoolExecutor(5, Util.threadFactory("http-remoting/OkHttp Scheduler", false)));
+    private static final ScheduledExecutorService asyncRetryExecutor = Executors.newSingleThreadScheduledExecutor();
```
iamthomaspowell, line 51: Does this mean all call retries due to `QosIoException` are scheduled on the same thread?

uschi2000:
I think this may now be obsolete, @nziebart . Feel free to reopen against develop if it's still relevant.

Issue #594, opened 2017-10-30
okhttp clients does not declare dependency on JaxRS


gatesn:
https://github.com/palantir/http-remoting/blob/d1e472729a8eb71c93a5bfdaf8c761a2c493e799/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java#L42

This therefore doesn't raise as a version conflict, and I'm getting a runtime NoSuchMethod exception:

```
Caused by: java.lang.NoSuchMethodError: javax.ws.rs.core.Response$Status$Family.familyOf(I)Ljavax/ws/rs/core/Response$Status$Family;
	at com.palantir.remoting3.okhttp.metrics.HostMetrics.record(HostMetrics.java:42)
	at com.palantir.remoting3.okhttp.metrics.HostMetricsRegistry.record(HostMetricsRegistry.java:48)
```

uschi2000:
I'm pretty sure this project has a transitive dependency on ws.rs through the `error-handling` dependency.

justinuang:
To clarify, the actual issue is that we depend on jersey-core:1.9 transitively through hadoop, which packages some of the classes declared in javax.ws.rs:javax.ws.rs-api. The Response$Status$Family class provided in jersey-core doesn't have the familyOf method.

Unfortunately, I don't think we have a choice regarding the jersey-core dependency since it's still in the latest hadoop here: https://hadoop.apache.org/docs/r2.8.2/hadoop-project-dist/hadoop-hdfs/dependency-analysis.html

Could we try inlining that method and see if it fixes our build? The hypothesis is that jersey-core brings in classes that are roughly the same, but only differ on a small subset of methods.

uschi2000:
closed by https://github.com/palantir/http-remoting/pull/605

justinuang:
Thanks! Can we get a release? or is there a way to test a snapshot build

dansanduleac:
@uschi2000 should actually close this issue :)

justinuang:
Confirmed that this fixes our flakes =) Thanks again!

PR #605, opened 2017-11-03
Rfink/do not use ws rs family

uschi2000:


uschi2000:
fixes #594

justinuang:
I think it's in several other places as well.

screen shot 2017-11-03 at 6 17 07 pm

uschi2000:
done

uschi2000:
all the other libraries on the classpath can obviously still run this code path, so it's not clear to me that fixing it in http-remoting is going to fully resolve your problem.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -40,11 +40,11 @@
     /** Creates a metrics registry for calls from the given service to the given host. */
     public HostMetrics(TaggedMetricRegistry registry, String serviceName, String hostname) {
         informational = registry.meter(name(serviceName, hostname, "informational"));
-        successful    = registry.meter(name(serviceName, hostname, "successful"));
-        redirection   = registry.meter(name(serviceName, hostname, "redirection"));
-        clientError   = registry.meter(name(serviceName, hostname, "client-error"));
-        serverError   = registry.meter(name(serviceName, hostname, "server-error"));
-        other         = registry.meter(name(serviceName, hostname, "other"));
```
ellisjoe, line 47: lame

ellisjoe (commented):
sad. what did they break exactly?

uschi2000:
unclear who broke what, but in Hadoop-reality there are conflicting versions of this class.

uschi2000:
sort of a miracle that none of the other libraries would hit this, but the sacrifice in http-remoting is small if it helps.

uschi2000:
+1, @ellisjoe ?

ellisjoe (approved):

PR #606, opened 2017-11-03
Add UserAgents#REMOTING_AGENT_NAME

uschi2000:


http-clients/src/main/java/com/palantir/remoting3/clients/UserAgents.java
```diff
@@ -34,6 +34,13 @@
 
 public final class UserAgents {
 
+    /**
+     * The {@link com.palantir.remoting3.clients.UserAgent.Agent#name agent name}identifying the http-remoting library
```
ellisjoe: space after `}`

uschi2000: done

PR #610, opened 2017-11-07
Add MeshProxyInterceptor to allow envoy use over SSL

jared2501:


http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
```diff
@@ -61,6 +64,15 @@
     /** The credentials to use for the proxy selected by {@link #proxy}. */
     Optional<BasicCredentials> proxyCredentials();
 
+    /**
+     * Rewrites all request URLs to the mesh proxy, but keeps the "Host" (or HTTP2 :authority pseudo-header) as the
+     * authority from the original request URL.
+     * <p>
+     * Note that if this option is set, then the {@link #maxNumRetries} must also be set to 0 since the mesh proxy is
+     * expected to handle all retry logic.
```
markelliot: The proxy retries even for streamed requests?

jared2501: Ah no it doesn't - but we really shouldn't be retrying for streamed requested in http-remoting either most likely and should let the application retry. This is since the Request object in OkHttp is an InputStream.

jared2501: Request's body*

jared2501: Ah actually MultipartBody has a `writeTo(BufferedSink sink)` method.

jared2501: It's questionable if we should transparently retry streams imo - it could lead to some pretty performance problems.

markelliot: How do retries of very large requests work? Say a request that’s 2MB?

jared2501: I am not too sure how this works with HTTP 1.1 services, but in HTTP 2 the upstream can actually respond before the client has written the body of the request. Since envoy does HTTP 2 streams between all the envoy nodes I imagine they'd implement it so that the rate limiter responds immediately before the request body arrives.

jared2501: Will have to do a bit of investigation on our test stack. [Here's the envoy documentation for retries
](https://www.envoyproxy.io/envoy/configuration/http_filters/router_filter.html#config-http-filters-router-x-envoy-retry-on) - it's reasonably sparse about the details.

markelliot: I think it’s optimistic to outlaw retries usage here, I guess.

jared2501: The issue if you don't disable retries is then you get n * m retries. Perhaps we want to not require that people specify a number here so that we can tweak it later, and start with a default of 0. I'll confirm the retry logic is what we expect on our test stack.

Another change we can do at a later date is to support annotating methods as idempotent, and only retrying on idempotent methods. Or, perhaps, assuming idempotent as default.

markelliot: Sure, forget it.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+    public Response intercept(Chain chain) throws IOException {
+        HttpUrl originalUri = chain.request().url();
+        return chain.proceed(chain.request()
+                .newBuilder()
+                .url(originalUri.newBuilder()
+                        .host(meshProxy.getHostText())
+                        .port(meshProxy.getPort())
+                        .build())
+                // TODO(jnewman): add HTTP2 :authority pseudo-header as well? Or does OkHttp handle this for us?
+                .header(HttpHeaders.HOST, String.format("%s:%s", originalUri.host(), originalUri.port()))
```
jared2501: We should only write the port on the Host if it's set - the port is part of the authority

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
```diff
@@ -61,6 +64,15 @@
     /** The credentials to use for the proxy selected by {@link #proxy}. */
     Optional<BasicCredentials> proxyCredentials();
 
+    /**
+     * Rewrites all request URLs to the mesh proxy, but keeps the "Host" (or HTTP2 :authority pseudo-header) as the
```
uschi2000: this is quite correct. if set, then all requests are send to the configured proxy instead of the target URL. more concretely:
- all requests are sent to the configured mesh proxy instead of the configured {@link #url}
- an additional "Host" / ":authority" header is injected
- maxNumRetries must be 0
- only one URL should be specified (I think?)

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
```diff
@@ -70,6 +82,13 @@
      */
     Duration backoffSlotSize();
 
+    @Value.Check
+    default void check() {
+        if (meshProxy().isPresent()) {
+            checkArgument(maxNumRetries() == 0, "If meshProxy is configured then maxNumRetries must be 0");
```
uschi2000, line 90: I think we have some constraints regarding the target URI? (e.g., exactly one? at most one?)

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -60,11 +60,30 @@ public static ClientConfiguration of(ServiceConfiguration config) {
                 .enableGcmCipherSuites(config.enableGcmCipherSuites().orElse(DEFAULT_ENABLE_GCM_CIPHERS))
                 .proxy(config.proxy().map(ClientConfigurations::createProxySelector).orElse(ProxySelector.getDefault()))
                 .proxyCredentials(config.proxy().flatMap(ProxyConfiguration::credentials))
+                .meshProxy(meshProxy(config.proxy()))
                 .maxNumRetries(config.maxNumRetries().orElse(config.uris().size()))
                 .backoffSlotSize(config.backoffSlotSize().orElse(DEFAULT_BACKOFF_SLOT_SIZE))
                 .build();
     }
 
+    public static ClientConfiguration of(
```
uschi2000: would really to not add more static constructors. can we do without this?

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -96,13 +115,22 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
...
         }
 
         throw new IllegalStateException("Failed to create ProxySelector for proxy configuration: " + proxyConfig);
     }
 
+    private static Optional<HostAndPort> meshProxy(Optional<ProxyConfiguration> proxy) {
+        if (!proxy.isPresent() || proxy.get().type() != ProxyConfiguration.Type.MESH) {
+            return Optional.empty();
+        }
+        return Optional.of(HostAndPort.fromString(proxy.get().hostAndPort().get()));
```
uschi2000: think I'd invert this:
```
if (A && B) {
  return positive;
} else {
  return negative;
}
```

http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
```diff
@@ -67,4 +71,26 @@ public void testFromParameters_fillsInDefaults() {
         assertThat(actual.enableGcmCipherSuites()).isFalse();
         assertThat(actual.proxy().select(URI.create("https://foo"))).containsExactly(Proxy.NO_PROXY);
     }
+
+    @Test
+    public void meshProxy_maxRetriesSetTo0() throws Exception {
```
uschi2000: nit: maxRetriesMustBe0

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.net.HostAndPort;
+import com.google.common.net.HttpHeaders;
+import java.io.IOException;
+import okhttp3.HttpUrl;
+import okhttp3.Interceptor;
+import okhttp3.Response;
+
+/**
+ * An {@link Interceptor} that changes the {@link okhttp3.Request#url Request URL's} authority to a given {@link
+ * HostAndPort}, but adds a "Host" header that is set to the original Request's authority. This allows the use of a L4
```
uschi2000: s/but/and/

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/MeshProxyInterceptorTest.java
```diff
@@ -0,0 +1,78 @@
...
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+public final class MeshProxyInterceptorTest {
```
uschi2000, line 35: I'd like to see a test that sanity-checks the interplay of this interceptors with the other interceptors that change URLs. (This is mostly my bad: the interceptor-based logic here is really a bit messed up and complicated :( )

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
jared2501: updated, with slight nit: the "Host" header is always set, but OkHttp will normally set it for you ([14.23 of RF2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)).

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
jared2501: added + added tests

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
jared2501: removed

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
jared2501: but baseline says you should do it the other way!

jared2501: done though :)

http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
jared2501: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
jared2501: done

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/MeshProxyInterceptorTest.java
jared2501: Hmm any suggestions on how to add this test / what it should test exactly? Considering that `ClientConfiguration` disallows retries, it effectively disables all other interceptors from rewriting URIs.

http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
```diff
@@ -67,4 +71,39 @@ public void testFromParameters_fillsInDefaults() {
...
+                .proxy(ProxyConfiguration.mesh("localhost:1234"))
+                .maxNumRetries(0)
+                .build());
+        assertThat(validConfig.meshProxy()).isEqualTo(Optional.of(HostAndPort.fromParts("localhost", 1234)));
+        assertThat(validConfig.maxNumRetries()).isEqualTo(0);
+    }
+
+    @Test
+    public void meshProxy_exactlyOneUri() throws Exception {
+        ServiceConfiguration invalidServiceConfig = ServiceConfiguration.builder()
```
uschi2000: I'm fine with making a test-local factory method here to keep the tests shorter. (didn't want to bloat the public API main/ with another factory)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        HttpUrl originalUri = chain.request().url();
+        return chain.proceed(chain.request()
+                .newBuilder()
+                .url(originalUri.newBuilder()
+                        .host(meshProxy.getHostText())
+                        .port(meshProxy.getPort())
+                        .build())
+                // TODO(jnewman): add HTTP2 :authority pseudo-header as well? Or does OkHttp handle this for us?
```
uschi2000: can you make and link an Issue?

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/MeshProxyInterceptorTest.java
uschi2000: ja, probably not much to do here. all good

http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
jared2501: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
jared2501: done

PR #611, opened 2017-11-08
Better defaults for maxNumRetries

a-k-g:
The motivation is to have more number of retries when there's just 1 uri and also limit the maximum number of retries in case there's a lot of uris.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
+++ b/http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
@@ -42,6 +42,8 @@ public final class ClientConfigurations {
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(10);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(500);
     private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
+    private static final int DEFAULT_MAX_NUM_RETRIES_LOWER_LIMIT = 3;
+    private static final int DEFAULT_MAX_NUM_RETRIES_UPPER_LIMIT = 10;
 
     private ClientConfigurations() {}
 
@@ -60,7 +62,7 @@ public final class ClientConfigurations {
                 .enableGcmCipherSuites(config.enableGcmCipherSuites().orElse(DEFAULT_ENABLE_GCM_CIPHERS))
                 .proxy(config.proxy().map(ClientConfigurations::createProxySelector).orElse(ProxySelector.getDefault()))
                 .proxyCredentials(config.proxy().flatMap(ProxyConfiguration::credentials))
-                .maxNumRetries(config.maxNumRetries().orElse(config.uris().size()))
+                .maxNumRetries(config.maxNumRetries().orElseGet(() -> getMaxNumRetries(config.uris().size())))
                 .backoffSlotSize(config.backoffSlotSize().orElse(DEFAULT_BACKOFF_SLOT_SIZE))
                 .build();
     }
@@ -81,7 +83,7 @@ public final class ClientConfigurations {
                 .enableGcmCipherSuites(DEFAULT_ENABLE_GCM_CIPHERS)
                 .proxy(ProxySelector.getDefault())
                 .proxyCredentials(Optional.empty())
-                .maxNumRetries(uris.size())
+                .maxNumRetries(getMaxNumRetries(uris.size()))
                 .backoffSlotSize(DEFAULT_BACKOFF_SLOT_SIZE)
                 .build();
     }
@@ -113,5 +115,14 @@ public final class ClientConfigurations {
             @Override
             public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {}
         };
+    }
+
+    private static Integer getMaxNumRetries(int numUris) {
+        if (numUris > DEFAULT_MAX_NUM_RETRIES_UPPER_LIMIT) {
+            return DEFAULT_MAX_NUM_RETRIES_UPPER_LIMIT;
+        } else if (numUris < DEFAULT_MAX_NUM_RETRIES_LOWER_LIMIT) {
+            return DEFAULT_MAX_NUM_RETRIES_LOWER_LIMIT;
+        }
+        return numUris;
     }
 }
--- a/http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
+++ b/http-clients/src/test/java/com/palantir/remoting3/clients/ClientConfigurationsTest.java
@@ -26,6 +26,9 @@ import java.net.Proxy;
 import java.net.URI;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.X509TrustManager;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -42,6 +42,8 @@
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(10);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(500);
     private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
+    private static final int DEFAULT_MAX_NUM_RETRIES_LOWER_LIMIT = 3;
```
jared2501, line 45: Hmmm, not sure this is a wise limit given #610. Also, not sure why you'd want to retry 3 times if there are just 2 URLs, especially given we apply exponential backoff with a pretty low retry delay. If you upped the retry delay, then you're going to get some really bad p99 delays.

jared2501: Basically, think trying a request each time against each URI is the best you can do - if all URIs are down, then you're out of luck, you better fail the request.

a-k-g: A URI can often be "down" when let's say the service is in a stale state due to which it is unable to serve the request and retuns a 503. In such a case, even if there are just 1 or 2 URLs, you'd want the request to retry as it's expected that the stale state is very brief.

uschi2000: yep, it's fine to retry more than #urls times. but I don't think we should impose a lower bound.

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -114,4 +116,13 @@ private static ProxySelector fixedProxySelectorFor(Proxy proxy) {
             public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {}
         };
     }
+
+    private static Integer getMaxNumRetries(int numUris) {
```
uschi2000, line 120: not a fan of silently overwriting the user-supplied config. would prefer no limits.

diogoholanda: This is not overwriting any user-supplied config though. This is only used if the maxNumRetries is not provided right?

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
diogoholanda: Wouldn't people still be able to overwrite this though?
We are strictly talking about good defaults here. Doesn't a minimum greater than one seem more appropriate?

uschi2000: oh, I didn't read carefully enough, sorry.

jared2501: It's not clear that a retry of > 1 but less than the number of URIs actually is more appropriate. In particular, I think without a global timeout then we will end up being less resilient due to retry explosions deeper in a service call tree. [This link](https://developers.redhat.com/blog/2017/06/01/microservices-patterns-with-envoy-proxy-part-ii-timeouts-and-retries/) describes this and how envoy solves it.

jared2501: Going to talk with @uschi2000 offline an report back here!

jared2501: Correction: It's not clear that a retry of > 1 AND greater than the number of URIs is more appropriate**

diogoholanda: @uschi2000 @jared2501 any conclusions from your talks?

jared2501:
We haven't talked yet

jared2501:
Rob's out of the office today. Likely will chat tonorrow.

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
a-k-g: No conclusion here?

jared2501: Not yet!

jared2501: This is just a change of defaults - would be surprised if this is blocking anything (you can always change your configuration defaults if you so desire).

a-k-g: This isn't blocking sure, but changing the defaults in the product's shipped configuration isn't really a desirable solution. There's multiple issues there:
- If we put a default value in the shipped config, it'll no longer grow with the number of available uris
- If we do end up changing the behavior here, that would mean removing `max-num-retries` from the product's shipped config. Now at that point if people have already overridden the product default (maybe if they have more URIs), their override will become unnecessary. 

The alternative would be to encode the logic that I have in this PR in my code, but the way the service configuration is used along with internal libraries makes that code quite fragile.

Regarding the retry explosion issue that you brought up, doesn't it already exist for cases when number of URIs > 1? I'm not sure why the problem would happen only if retry > greater than number of URIs.

jared2501: @a-k-g - 

The main thing is that if retries > URLs then you're going to retry URLs that you know have failed. Currently, retrying URLs that you know have failed currently won't significantly increase the chance of success after they've been tried already since the default backoff is quite small and the number of URLs that are tried is also quite small. I.e. only milliseconds will have passed between retries.

If we were to increase the period that we wait before retring already tried URLs, then we start to have issues with threadpools building up and clogging up other requests that are coming in that won't necessarily fail. Other RPC systems appear to solve this problem by some combination of: a) a global deadline for an RPC chain, b) circuit breaking c) making the RPC calls non-blocking, so that the queues for outgoing RPCs can be much larger.

Without some combination of the above, I doubt changing the default behaviour to try a URL more than once will offer any upside, and in fact may actually be harmful.

However, I'd like to chat with @uschi2000 about this.

uschi2000:
I'd propose to punt on this for now: Main reason is that's it's not likely that increasing the number of retries without more advanced global throttling, timeouts, and circuit breaking is going to improve the performance of the overall system. We're going to implement this in a more first-class way in the Mesh and I think it's our best bet to leave the behavior unchanged until we have better data to base a decision on.

diogoholanda:
I think that's fine. I just wanted to comment here about the actual use case that motivated the change:
In general, what @jared2501 said about retrying the same node being likely to fail again and the retry being undesirable is true. However, this is not the case for QoS exceptions. The service is literally asking us to just wait a little bit and retry later, but if the number of URLs is 1, then our request will simply fail. Maybe we should special case the retries for QoS then?

jared2501:
I'd say we probably want to put a lot of this retry logic into the mesh
later rather than spending too much time trying to get it right in the http
remoting layer. The mesh layer is not far off! We already have it deployed
to our internal platform release stack, and have it comfortably sitting
between a couple services.

diogoholanda:
I am fine with the decision here. Just wanted to expose the actual problem we had to make sure it will be addressed in the mesh solution.

a-k-g:
Thanks for getting this resolved.

PR #616, opened 2017-11-09
Remove service name from metric name

ellisjoe:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/metrics/HostMetrics.java
```diff
@@ -26,7 +26,8 @@
  */
 public final class HostMetrics {
 
-    public static final String CLIENT_RESPONSE_METRIC_NAME_SUFFIX = ".client.response";
+    public static final String CLIENT_RESPONSE_METRIC_NAME = ".client.response";
```
uschi2000: why is there a prefix `.`?

ellisjoe: oops. fixed

uschi2000 (approved):

PR #624, opened 2017-11-15
[BREAK] Refactor okhttp-clients: handle all custom behavior by wrapping calls in RemotingOkHttpCall

uschi2000:
In particular, removes all non-local behavior from interceptors


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.concurrent.TimeoutException;
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.HttpUrl;
+import okhttp3.Request;
+import okhttp3.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class RemotingOkHttpCall extends ForwardingCall {
```
uschi2000: start your review here

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.time.Duration;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.Supplier;
+import okhttp3.Call;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+
+final class RemotingOkHttpClient extends ForwardingOkHttpClient {
```
uschi2000: start your review here

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                callback.onFailure(call, new IOException("Failed to handle request, this is an http-remoting bug."));
+            }
+        });
+    }
+
+    @Override
+    public Response execute() throws IOException {
+        SettableFuture<Response> future = SettableFuture.create();
```
iamdanfox: Why guava SettableFuture over java8 CompletableFuture?

uschi2000: It's mostly a religious question.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                future.set(response);
+            }
+        });
+
+        try {
+            return future.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.interrupted();
```
iamdanfox: `Thread.interrupted();` just returns a boolean - not sure this is doing what you want:

https://github.com/palantir/http-remoting/pull/587/files#diff-07eb3e409f7215a8a8b71594d6a0b47bR52

uschi2000: ouch. thanks, fixed.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -167,7 +152,25 @@ private static QosIoExceptionAwareOkHttpClient createInternal(
         // dispatcher with static executor service
         client.dispatcher(createDispatcher());
 
-        return new QosIoExceptionAwareOkHttpClient(client.build(), handlerFactory);
+        return new RemotingOkHttpClient(
+                client.build(),
+                () -> new ExponentialBackoff(config.maxNumRetries(), config.backoffSlotSize(), new Random()),
```
iamdanfox, line 157: Could we make this not a supplier? Even if the config live reloads internally, I think it would be nice to not have to pass it around as an arg to `newCallWithMutableState`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+                        client.newCallWithMutableState(redirectedRequest, backoffStrategy, maxNumRelocations - 1)
+                                .enqueue(callback);
+                    }
+                }
+            }
+
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                // Relay successful responses
+                if (response.code() / 100 == 2) {
```
iamdanfox: haha surely we just want to explicitly check the valid codes!!!

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+                    callback.onResponse(call, response);
+                }
+
+                // Handle to handle QoS situations: retry, failover, etc.
+                Optional<QosException> qosError = qosHandler.handle(response);
+                if (qosError.isPresent()) {
+                    qosError.get().accept(new QosException.Visitor<Void>() {
```
iamdanfox: I think it would help readability to pull this chunk into a method, that way the `onResponse` will read very clearly as: success=>earlyreturn, hand qos, handle remoteexception, handle anything else,

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+import okhttp3.Request;
+import okhttp3.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class RemotingOkHttpCall extends ForwardingCall {
+
+    private static final Logger log = LoggerFactory.getLogger(RemotingOkHttpCall.class);
+
+    private static final ResponseHandler<RemoteException> remoteExceptionHandler =
```
iamdanfox: :+1: This is much clearer than the old stuff

retrofit2-clients/src/test/java/com/palantir/remoting3/retrofit2/Retrofit2ClientApiTest.java
```diff
@@ -71,9 +71,9 @@ public void testOptionalStringHandling() throws IOException, InterruptedExceptio
 
         server.enqueue(new MockResponse().setBody("\"pong\""));
         assertThat(service.getGuavaOptionalString(guavaOptional("p"), guavaEmptyOptional()).execute().body())
-            .isEqualTo(guavaOptional("pong"));
+                .isEqualTo(guavaOptional("pong"));
         assertThat(server.takeRequest().getPath())
-            .isEqualTo("/getGuavaOptionalString/p/");
+                .isEqualTo("/getGuavaOptionalString/p/");
```
iamdanfox, line 88: @rfink could you copy in the extra tests I wrote to cover retrofit error handling?  It was pretty under-specified before: https://github.com/palantir/http-remoting/pull/623/files#diff-506462de9dcfed10a388f16cf93d6c96R136

- completableFuture_should_throw_RemoteException_for_server_serializable_errors
- sync_retrofit_call_should_throw_RemoteException_for_server_serializable_errors
- async_retrofit_call_should_throw_RemoteException_for_server_serializable_errors
- completableFuture_should_throw_normal_IoException_for_client_side_errors

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+        });
+
+        try {
+            return future.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new IOException("Call was interrupted during execution");
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof IoRemoteException) {
+                throw ((IoRemoteException) e.getCause()).getWrappedException();
```
iamdanfox: I think this is a bit sneaky - all 10 other exceptions thrown in this file are IOExceptions but this one is a RuntimeException.

I think I'd prefer this unwrapping to happen conspicuously out in the RemotingOkHttpClient, so that it can be clearly called out as the sync equivalent of the stuff in AsyncSerializableErrorCallAdapterFactory.

As a minimum I think your IoRemoteException really needs some explanatory [javadoc](https://github.com/palantir/http-remoting/pull/623/files#diff-413ccc6ce7a71f7570f59237fcf103b7R25) to describe why we're wrapping things

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+            public void onFailure(Call call, IOException exception) {
+                if (!backoffStrategy.nextBackoff().isPresent()) {
+                    callback.onFailure(call, new IOException("Failed to complete the request due to an "
+                            + "IOException", exception));
+                } else {
+                    Optional<HttpUrl> redirectTo = urls.redirectToNext(request().url());
+                    if (!redirectTo.isPresent()) {
+                        callback.onFailure(call, new IOException("Failed to determine valid failover URL"
+                                + "for '" + request().url() + "' and base URLs " + urls.getBaseUrls()));
+                    } else {
```
iamdanfox: Could we try turning these nested if/else blocks into early returns? I think it would make it more readable and less deeply nested.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+                                .enqueue(callback);
+                    }
+                }
+            }
+
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                // Relay successful responses
+                if (response.code() / 100 == 2) {
+                    callback.onResponse(call, response);
```
iamdanfox: Also why isn't there a `return` after you call the onResponse handler?  the current impl looks like it could call both `onResponse` and `onFailure`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    @Override
+    public void enqueue(Callback callback) {
+        doEnqueueAndHandleErrorsAndRetries(callback);
+    }
+
+    @Override
+    public RemotingOkHttpCall doClone() {
```
iamdanfox: @j-baker objected to these doClone methods before... can we just make ForwardingCall an interface with lots of `default` methods, missing just `clone` and `getDelegate` implementations. Gets rid of a bit of inheritance :)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+        super.enqueue(new Callback() {
+            @Override
+            public void onFailure(Call call, IOException exception) {
+                if (!backoffStrategy.nextBackoff().isPresent()) {
+                    callback.onFailure(call, new IOException("Failed to complete the request due to an "
+                            + "IOException", exception));
+                } else {
+                    Optional<HttpUrl> redirectTo = urls.redirectToNext(request().url());
+                    if (!redirectTo.isPresent()) {
+                        callback.onFailure(call, new IOException("Failed to determine valid failover URL"
```
iamdanfox: I'm a bit concerned about the readability of stacktraces here... there are so many identical IOExceptions thrown in this class, with just a string interpolated method to differentiate them.  What about subclassing to distinguish a few of them?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/AsyncCallTag.java
```diff
@@ -1,50 +0,0 @@
...
- * So, we have this interface. The intended use is that all calls are tagged with an {@link AsyncCallTag}. If the
- * request is to be called in an async fashion, {@link #setCallAsync()} is called.
- * <p>
- * The interceptor will call {@link #isAsyncCall()}, which will return true iff it was registered in this way. In this
- * case, the interceptor is a no-op.
- * <p>
- * The {@code AsyncSerializableErrorCallAdapterFactory} (in the retrofit2-clients project) then handles filling the
- * exception into the response.
- */
-public final class AsyncCallTag {
```
iamdanfox, line 40: 🥇

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.io.CharStreams;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+import okhttp3.Response;
+
+enum IoExceptionResponseHandler implements ResponseHandler<IOException> {
```
iamdanfox, line 28: I think this usage of enums upsets error-prone.  Can we just have a private constructor and a `public static final INSTANCE` instead?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+
+                // Handle responses that correspond to RemoteExceptions / SerializableErrors
+                Optional<RemoteException> httpError = remoteExceptionHandler.handle(response);
+                if (httpError.isPresent()) {
+                    callback.onFailure(call, new IoRemoteException(httpError.get()));
+                    return;
+                }
+
+                // Catch-all: handle all other responses
+                Optional<IOException> ioException = catchAllExceptionHandler.handle(response);
```
iamdanfox: Seems a bit surprising that the 'catchAllExceptionHandler' doesn't actually catch all exceptions and lets some stuff fall through to that last case.  Can we rename/change behaviour to make this more obvious?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,238 @@
...
+            RemoteExceptionResponseHandler.INSTANCE;
+    private static final ResponseHandler<IOException> catchAllExceptionHandler = IoExceptionResponseHandler.INSTANCE;
+    private static final ResponseHandler<QosException> qosHandler = QosExceptionResponseHandler.INSTANCE;
+
+    private final BackoffStrategy backoffStrategy;
+    private final UrlSelector urls;
+    private final RemotingOkHttpClient client;
+    private final ScheduledExecutorService schedulingExecutor;
+    private final ExecutorService executionExecutor;
+    private final Duration timeout;
```
iamdanfox: might be worth elaborating on the purpose of this timeout as there are so many timeouts involved with http rpc already

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
uschi2000: it has to be a factory since each call needs a fresh, stateful Backoff instance. (this has nothing to do with config reload)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: which ones? all 2xx should be considered successful.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
```diff
@@ -0,0 +1,53 @@
...
+    INSTANCE;
+
+    @Override
+    public Optional<IOException> handle(Response response) {
+        if (response.isSuccessful() || response.code() == 101) {
```
iamdanfox: Could we give this magic number a name since it's not present in the usual `javax.ws.rs.core.Response.Status`?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
iamdanfox: Ah OK, I didn't clock the interior mutability in the ExponentialBackoff implementation of `BackoffStrategy`.  Given there's only one implementation right now, what do you think about collapsing that interface? (It could obviously return if we needed it again)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: `200 <= response.code() && responseCode < 300` ? 

The division is visually pretty similar to a modulo operation

retrofit2-clients/src/test/java/com/palantir/remoting3/retrofit2/Retrofit2ClientApiTest.java
uschi2000: done, except for async_retrofit_call_should_throw_RemoteException_for_server_serializable_errors , see https://github.com/palantir/http-remoting/issues/625

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -167,7 +151,25 @@ private static QosIoExceptionAwareOkHttpClient createInternal(
...
+                UrlSelectorImpl.create(config.uris(), randomizeUrlOrder),
+                schedulingExecutor,
+                executionExecutor,
+                largestOf(config.connectTimeout(), config.readTimeout(), config.writeTimeout())
+        );
+    }
+
+    @VisibleForTesting
+    static Duration largestOf(Duration... durations) {
```
iamdanfox, line 166: `Arrays.stream(durations).max()` ?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemoteExceptionResponseHandler.java
```diff
@@ -27,54 +27,50 @@
...
+                body = toString(response.body().byteStream());
+            } catch (IOException e) {
+                return Optional.empty();
+            }
+            try {
+                SerializableError serializableError = MAPPER.readValue(body, SerializableError.class);
+                return Optional.of(new RemoteException(serializableError, response.code()));
             } catch (Exception e) {
-                String message = String.format(
+                log.warn("Failed to deserialize JSON: {}", String.format(
```
iamdanfox: This log.warn is a bit strange... the String.format gives us a double message.  Also if we used normal `{}` then you could probably safeArg the response.code() at least

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+                throw new IOException("Failed to execute call", e);
+            }
+        } catch (TimeoutException e) {
+            throw new IOException("Call timed out after: " + timeout.toString(), e);
+        }
+    }
+
+    @Override
+    public void enqueue(Callback callback) {
+        doEnqueueAndHandleErrorsAndRetries(callback);
```
iamdanfox: thoughts on inlining this? or if you want the descriptive method name, what about also using it in the sync `execute()` method?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+            @Override
+            public void onResponse(Call call, Response response) throws IOException {
+                future.set(response);
+            }
+        });
+
+        try {
+            return future.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
iamdanfox: Could we copy callum's test that verifies this works: https://github.com/palantir/http-remoting/pull/587/files#diff-39cfad0395ff8616ffa04724e0c47221R29

iamdanfox: Also why aren't we calling `getDelegate().cancel()`?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+            throw new IOException("Call was interrupted during execution");
```
iamdanfox: what about an `InterruptedIOException`?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                }
+                return null;
+            }
+
+            @Override
+            public Void visit(QosException.Unavailable exception) {
+                Optional<Duration> backoff = backoffStrategy.nextBackoff();
+                if (!backoff.isPresent()) {
+                    log.debug("Max number of retries exceeded, failing call");
```
iamdanfox, line 237: I bet it would be helpful for debugging if we could log /throw how many retries we've attempted.  Seeing this message "failing call after 0 retries" would be an immediate red flag, as would "failing call after 49000 retries"

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+                } else {
+                    Optional<HttpUrl> redirectTo = urls.redirectToNext(request().url());
+                    if (!redirectTo.isPresent()) {
+                        callback.onFailure(call, new IOException("Failed to determine valid failover URL"
+                                + "for '" + request().url() + "' and base URLs " + urls.getBaseUrls()));
+                    } else {
+                        Request redirectedRequest = request().newBuilder()
+                                .url(redirectTo.get())
+                                .build();
+                        client.newCallWithMutableState(redirectedRequest, backoffStrategy, maxNumRelocations - 1)
```
iamdanfox: I think this backoffStrategy argument is the one that confuses me - the method says we're mutating stuff, but it appears that we're passing the same thing in again.  If it's internally mutable, we might as well leave it where it is - otherwise, I'd expect to be calling `.backoffMore()` or something

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+
+    @Override
+    public void enqueue(Callback callback) {
+        doEnqueueAndHandleErrorsAndRetries(callback);
+    }
+
+    @Override
+    public RemotingOkHttpCall doClone() {
+        return new RemotingOkHttpCall(getDelegate().clone(),
+                backoffStrategy,
```
iamdanfox: are we OK with passing the original `backoffStrategy` to the cloned thing?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,245 @@
...
+                callback.onFailure(call, new IOException("Failed to handle request, this is an http-remoting bug."));
+            }
+        });
+    }
+
+    private QosException.Visitor<Void> createQosVisitor(Callback callback, Call call) {
+        return new QosException.Visitor<Void>() {
+            @Override
+            public Void visit(QosException.Throttle exception) {
+                Optional<Duration> backoff = exception.getRetryAfter().isPresent()
```
nziebart: Do we want to give the backoff strategy a hint (the retry-after header), but ultimately allow it to decide what the backoff should be? Currently a service that constantly returns retry-after will cause us to retry indefinitely

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: yep, fair. will tackle this in later PR to keep this one somewhat contained. https://github.com/palantir/http-remoting/issues/626

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: fixed, thanks

uschi2000: (added 2 tests here, too)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: future PR. https://github.com/palantir/http-remoting/issues/627

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: yeah, can do. again, will do in a future PR to keep this manageable, OK?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
uschi2000: error-prone is fine with this since the INSTANCE is immutable

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: Do you have a proposal for how to do this? The handlers basically implement a switch statements over HTTP codes and at the bottom is the `default` case. Not sure how to check at compile-time that these handlers are exhaustive.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: added.

uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: inlined

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: calling cancel now, thanks

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: maybe. part of the other TODO, ok?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: added the test, but it's not clear to me what exactly that thing is testing?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: valid request, but maybe not part of this refactor PR. added TODO

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: BackoffStrategy.nextBackoff mutates the internal state.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: yep: the backoff strategy carries the state regarding number of previous attempts.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: good call. added behavior and tests.

uschi2000:
@nziebart @iamdanfox ready for you again.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: @uschi2000 Callum and I wrote that test because after trouble with our product on production, a jstack and metrics showed many more hanging threads than open connections.

The test times out on http-remoting develop, demonstrating that if an incoming dropwizard request handler thread is interrupted (because the client aborts), any in-flight http-remoting network calls ignore this interruption and stay open for (~10 minutes).

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: Not sure why we'd want to defer a one line change - are there downsides?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: OK. with this refactor, the test is a bit bogus: it passes even when the thread interrupt line is removed.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoRemoteException.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting3.okhttp;
+
+import java.io.IOException;
+
+// TODO(rfink): Can we hide this as an implementation detail of RemotingOkHttpCall?
```
iamdanfox: I think this wrapper class warrants some Javadoc justifying the subtle reasons we had to implement it.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: ah, you mean the existing one, sure.

uschi2000: done

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoRemoteException.java
uschi2000: done. also changed the wrapped type.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: @uschi2000 even better, let's just use `boolean Response#isSuccessful()`:

https://square.github.io/okhttp/3.x/okhttp/okhttp3/Response.html#isSuccessful--

> Returns true if the code is in [200..300), which means the request was successfully received, understood, and accepted.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,256 @@
...
+
+                // Handle to handle QoS situations: retry, failover, etc.
+                Optional<QosException> qosError = qosHandler.handle(response);
+                if (qosError.isPresent()) {
+                    qosError.get().accept(createQosVisitor(callback, call));
+                    return;
+                }
+
+                // Handle responses that correspond to RemoteExceptions / SerializableErrors
+                Optional<RemoteException> httpError = remoteExceptionHandler.handle(response);
```
iamdanfox, line 163: I noticed both this `RemoteExceptionResponseHandler` and `IoExceptionResponseHandler` check for `Response.isSuccessful`, which will never be true if we're already filtering this out at the top of this method. Shall we remove these?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: We could rename this handler to say `catchEverythingExceptionHttp101` so that the name is consistent with it's signature, or if we want to keep it named `catchAll`, maybe we should just make this specific one not implement the `ResponseHandler` interface and just always return an IOException?

As an aside, what is the desired behaviour for a http 101 code?  At the moment it'll zoom all the way down the bottom here and trigger `new IOException("Failed to handle request, this is an http-remoting bug.")`.  Do remoting-servers return 101?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -0,0 +1,256 @@
...
+                return null;
+            }
+        };
+    }
+
+
+    // TODO(rfink): Consider removing RemotingOkHttpCall#doClone method, #627
+    @Override
+    public RemotingOkHttpCall doClone() {
+        return new RemotingOkHttpCall(getDelegate().clone(), backoffStrategy, urls, client, schedulingExecutor,
```
iamdanfox, line 255: If `backoffStrategy` has interior mutability, I don't think it's really fair to call this `clone` anymore because if you clone `remotingOkHttpCallA` to get `remotingOkHttpCallB` then they'd both share the same mutable `BackoffStrategy` instance - calls from either instance (presumably on different threads) would then mutate the same object.  This seems kinda antithetical to the normal 'clone' contract.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: a 101 should really never make it to this stage.

uschi2000: okhttp internals should catch it and trigger a protocol upgrade

uschi2000: @jared2501 , is that correct?

uschi2000: will now also relay 1xx responses as successful.

uschi2000: renamed exception handler

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: better safe than sorry?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: i think this is compatible with the Call contract:   
```
/**
   * Create a new, identical call to this one which can be enqueued or executed even if this call
   * has already been.
   */
```
, as well as with a general clone contract for mutable objects.

added some javadoc

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: Hmm, no I don't think it's desirable to have a branch that we know will never be hit.  If someone breaks the desired happy path somehow, I'd prefer things to fail as fast as possible (and get fixed), not be obscured by some kind of 'defense in depth' if statements.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: OK sounds like we're not aligned on this one, personally I'd write a bit more code to eliminate the last little bit of mutability and redo that BackoffStrategy so that it's immutable, but it's your PR not mine.

uschi2000: where would you prefer to keep state that is mutable across retries?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: OK. How do you propose we write this instead then?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
iamdanfox: If we reframed the concept of backoff handling to be captured by something like:

```java
interface BackoffSupplier {
    Optional<Duration> currentBackoffDuration(); //doesn't mutate anything
    BackoffSupplier increaseBackoff(); // constructs and returns a new instance
}
```

Then your retry call could look like this:
```java
client.newCallWithMutableState(
        redirectedRequest, backoff.increaseBackoff(), maxNumRelocations - 1);
```

iamdanfox:
The overall approach looks like it'll be a huge improvement in maintainability - I can't speak for the exact QOS handling that's been rewritten, but everything I understood looks great :)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
uschi2000: Don't think that's better since it means that you always need to remember to call two methods on the backoff object.

iamdanfox (approved):

uschi2000:
cool, thanks for taking the time to review, Dan.

uschi2000:
now, the big question: do we need to call this 4.0.0?

PR #632, opened 2017-11-16
Enable wrapping a ScheduledExecutorService with a new trace for every execution

diogoholanda:


diogoholanda:
@uschi2000 @bavardage

bavardage:
only gotcha here, is if you wrapWithNewTrace a ScheduledExecutorService which is already wrapped with the other method (which may be common in a 'standard' way of constructing these) then it won't act as expected

uschi2000:
diogo, can you add a java comment explaining ben's point? (this should be added to both wrapping functions.)

diogoholanda:
Something like that? Let me know if you want to rephrase it

bavardage:
nope, not like this - I think it happens the other way around.

The outer wrapping has no effect

bavardage:
actually no, you're correct - don't mind me!

bavardage (approved):

uschi2000 (approved):

PR #633, opened 2017-11-16
Track down connection leaks

j-baker:
Internally, we saw a service seemingly lock up (during a leader election), and when a GC happened then OkHttp logged a lot of exceptions claiming that we had leaked connection objects. The most obvious place for this to happen is here, and I have fixed it.

If we change a response into a thrown exception, we need to close the response.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/QosExceptionResponseHandler.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/QosExceptionResponseHandler.java
@@ -39,16 +39,22 @@ enum QosExceptionResponseHandler implements ResponseHandler<QosException> {
 
     @Override
     public Optional<QosException> handle(Response response) {
+        final Optional<QosException> result;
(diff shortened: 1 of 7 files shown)


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/QosExceptionResponseHandler.java
```diff
@@ -39,16 +39,22 @@
...
             case 503:
-                return Optional.of(handle503());
+                result = Optional.of(handle503());
+                break;
+            default:
+                result = Optional.empty();
         }
-
-        return Optional.empty();
+        result.ifPresent(exception -> response.close());
```
uschi2000: looks like we have to do this in all code paths that consumer the response.

j-baker: the other ones all try-with-resources it, so it's unnecessary.

PR #642, opened 2017-11-29
Remove family and hostname tags from metrics

ellisjoe:

PR #650, opened 2017-12-06
Reapply "Remove family and hostname tags from metrics (#642)"

ellisjoe:


ellisjoe:
`OkHttpClients` is the only class that had conflicts

iamdanfox (approved):

PR #653, opened 2017-12-13
Fix response buffering in RemotingOkHttpCall

uschi2000:


uschi2000:
@j-baker could you review this asap?

j-baker:
What has this actually changed?

gmaretic:
@j-baker Since each of the handlers could read the response body, there could be nothing left to read for subsequent handlers. As the test that's failing on the base branch is showing, if the `remoteExceptionHandler`  fails to deserialize a json into a `RemoteException`, `ioExceptionHandler` will create an exception with an empty body.
We now recreate the response on every call to a handler to avoid this.

j-baker:
Ok, that's something like what I assumed (my working assumption when I wrote this was that a handler wasn't allowed to read off the stream unless it knew it would return something of that type)

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -150,39 +157,39 @@ public void onFailure(Call call, IOException exception) {
...
+                if (response.code() / 100 <= 2) {
+                    callback.onResponse(call, response);
                     return;
                 }
 
                 // Buffer the response into a byte[] so that multiple handler can safely consume the body.
-                // The buffer call consumes and closes the original unbufferedResponse body.
-                final Response response;
+                // The buffer call consumes and closes the original response body.
+                byte[] bodyBytes;
```
j-baker: instead of forcing the downstreams to deal with the notion of response and its body, maybe just have something like

```
Supplier<Response> errorResponseSupplier;
try {
    byte[] body = response.body().bytes();
    // so error handlers can read the body without breaking subsequent handlers
    errorResponseSupplier = () -> buildFrom(response, bodyBytes);
}
```
?

and then it's just errorResponseSupplier.get()?

j-baker: actually, possibly just have a method

```
/**
 * Error handlers might need to read the body, copy it memory so subsequent handlers can read the same data.
 */
Supplier<Response> copyResponse(Response response);
```

which does that?

j-baker (approved):

PR #658, opened 2018-01-18
Proper infinite timeouts

iamdanfox:
After upgrading to http-remoting 3.13.0 our data upload product has been throwing timeout exceptions when an upload takes longer than 10 seconds.

This failure was introduced because okhttp treats zero as infinity but our new `largestOf` method treats zero as zero.  In addition, we had a `future.get(timeoutMillis)` call which also did not treat zero as infinity.

### DIFF
--- a/okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
+++ b/okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
@@ -500,9 +500,14 @@ public final class OkHttpClientsTest extends TestBase {
 
     @Test
     public void largestOf_sanity() throws Exception {
-        assertThat(OkHttpClients.largestOf()).isEqualTo(Duration.ZERO);
         assertThat(OkHttpClients.largestOf(Duration.ofMinutes(1), Duration.ofSeconds(20), Duration.ofHours(7)))
                 .isEqualTo(Duration.ofHours(7));
+    }
+
+    @Test
+    public void largestOf_treats_zero_as_infinity() throws Exception {
+        assertThat(OkHttpClients.largestOf(Duration.ZERO, Duration.ofSeconds(20), Duration.ofHours(7)))
+                .isEqualTo(Duration.ZERO);
     }
 
     private OkHttpClient createRetryingClient(int maxNumRetries) {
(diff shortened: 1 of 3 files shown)


okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -442,13 +446,70 @@ public void handlesInterruptedThreads() throws Exception {
...
+        server.enqueue(new MockResponse()
+                .setBodyDelay(Duration.ofSeconds(11).toMillis(), TimeUnit.MILLISECONDS)
+                .setBody("Hello, world!"));
+
+        OkHttpClient client = OkHttpClients.withStableUris(
+                ClientConfiguration.builder()
+                        .from(createTestConfig(url))
+                        .connectTimeout(Duration.ZERO)
+                        .readTimeout(Duration.ZERO)
+                        .writeTimeout(Duration.ZERO) // want to allow unlimited time for uploads
```
jmcampanini, line 460: if the writeTimeout is the timeout for the client writing to the server, then we aren't testing that in this case, since it's a sync-request

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -442,13 +446,70 @@ public void handlesInterruptedThreads() throws Exception {
...
+
     @Test
     public void largestOf_sanity() throws Exception {
-        assertThat(OkHttpClients.largestOf()).isEqualTo(Duration.ZERO);
         assertThat(OkHttpClients.largestOf(Duration.ofMinutes(1), Duration.ofSeconds(20), Duration.ofHours(7)))
                 .isEqualTo(Duration.ofHours(7));
     }
 
+    @Test
+    public void largestOf_treats_zero_as_infinity() throws Exception {
```
jmcampanini, line 508: is the throwing the illegal argument new behavior?

CRogers (approved):

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
iamdanfox: Yeah the illegal argument is intentional because this function is only ever called in one place (above) where we know there are three arguments.

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
iamdanfox: True, this test just verifies that slow downloads work.  In practise, since the timeout is at a high enough level (it can't distinguish between uploads vs downloads) I'm confident this will also be fine for uploads.

PR #660, opened 2018-01-22
Test IoException metric recording

ellisjoe:


uschi2000 (approved):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -119,7 +119,7 @@ private static RemotingOkHttpClient createInternal(
         client.sslSocketFactory(config.sslSocketFactory(), config.trustManager());
 
         // Intercept calls to augment request meta data
-        client.addNetworkInterceptor(InstrumentedInterceptor.create(registry, hostMetrics, serviceClass));
+        client.addInterceptor(InstrumentedInterceptor.create(registry, hostMetrics, serviceClass));
```
ellisjoe, line 122: connecting to a host occurs before the network interceptors are invoked which means some IOExceptions won't be seen by network interceptors. think moving it here is fine because we'll only record metrics for the responses the application actually sees.

uschi2000: would a test fail if this were a network interceptor?

ellisjoe: yes. the one added to OkHttpClientsTest

PR #662, opened 2018-01-25
Don't stick userinfo in Host header

sfackler:


okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/MeshProxyInterceptorTest.java
```diff
@@ -75,4 +75,17 @@ public void intercept_portNotInAuthority() throws Exception {
         assertThat(request.getValue().header(HttpHeaders.HOST)).isEqualTo("foo.com");
     }
 
+    @Test
+    public void intercept_userinfoNotInHost() throws Exception {
```
uschi2000: let's call this intercept_stripsUserinfoFromAuthority

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/MeshProxyInterceptor.java
```diff
@@ -39,14 +39,21 @@
     @Override
     public Response intercept(Chain chain) throws IOException {
         HttpUrl originalUri = chain.request().url();
+
+        StringBuilder host = new StringBuilder(originalUri.uri().getHost());
```
uschi2000: can you leave a comment here: 
```
// Not using URL#getAuthority since we are removing the userinfo portion. 
// Note that userinfo is not allowed in http2 :authority headers: https://tools.ietf.org/html/rfc7540#section-8.1.2.3
```

uschi2000 (approved):
cheers

uschi2000:
build failed

sfackler:
Fixed

PR #663, opened 2018-01-25
fix race in UrlSelector resulting in bad failover behaviour under any concurrent load

bavardage:
concurrent requests on the same client results in 'failover' to the same url you just failed on

the `UrlSelector` seems to be shared client-wide, so if you have multiple requests hitting a host which is down and a fairly small number of urls total, you end up with `urls.redirectToNext(request().url());` in `RemotingOkHttpCall:148` giving you the same url you just failed on, at which point you naturally fail again, then having exhausted all retries - give up!

### DIFF
--- a/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
+++ b/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
@@ -21,8 +21,13 @@ import static org.hamcrest.Matchers.startsWith;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 
+import com.google.common.collect.Lists;
 import com.palantir.remoting3.clients.ClientConfiguration;
 import feign.RetryableException;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import org.junit.Before;
@@ -56,6 +61,24 @@ public final class JaxRsClientFailoverTest extends TestBase {
         server2.enqueue(new MockResponse().setBody("\"foo\""));
 
         assertThat(proxy.string(), is("foo"));
+    }
+
+    @Test
+    public void testConnectionError_performsFailover_testHarder() throws Exception {
+        ExecutorService executorService = Executors.newFixedThreadPool(10);
+
+        server1.shutdown();
+        for (int i = 0; i < 100; i++) {
+            server2.enqueue(new MockResponse().setBody("\"foo\""));
+        }
+
+        List<Future<String>> things = Lists.newArrayListWithCapacity(100);
+        for (int i = 0; i < 100; i++) {
+            things.add(executorService.submit(() -> proxy.string()));
+        }
+        for (int i = 0; i < 100; i++) {
+            assertThat(things.get(i).get(), is("foo"));
+        }
     }
 
     @Test
--- a/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
+++ b/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
@@ -64,19 +64,19 @@ public final class JaxRsClientFailoverTest extends TestBase {
     }
 
     @Test
-    public void testConnectionError_performsFailover_testHarder() throws Exception {
-        ExecutorService executorService = Executors.newFixedThreadPool(10);
+    public void testConnectionError_performsFailover_concurrentRequests() throws Exception {
+        ExecutorService executorService = Executors.newFixedThreadPool(2);
 
         server1.shutdown();
-        for (int i = 0; i < 100; i++) {
+        for (int i = 0; i < 10; i++) {
             server2.enqueue(new MockResponse().setBody("\"foo\""));
         }
 
-        List<Future<String>> things = Lists.newArrayListWithCapacity(100);
-        for (int i = 0; i < 100; i++) {
+        List<Future<String>> things = Lists.newArrayListWithCapacity(10);
+        for (int i = 0; i < 10; i++) {
             things.add(executorService.submit(() -> proxy.string()));
         }
-        for (int i = 0; i < 100; i++) {
+        for (int i = 0; i < 10; i++) {
             assertThat(things.get(i).get(), is("foo"));
         }
     }
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
@@ -98,8 +98,9 @@ final class UrlSelectorImpl implements UrlSelector {
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl current) {
+        Optional<Integer> currentIndex = indexFor(current);
         int index = currentUrl.updateAndGet(
-                (currentUrlIndex) -> (indexFor(current).orElse(currentUrlIndex) + 1) % baseUrls.size());
+                (currentUrlIndex) -> (currentIndex.orElse(currentUrlIndex) + 1) % baseUrls.size());
         return redirectTo(current, baseUrls.get(index));
     }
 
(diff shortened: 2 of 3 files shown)


jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
```diff
@@ -58,6 +63,24 @@ public void testConnectionError_performsFailover() throws Exception {
         assertThat(proxy.string(), is("foo"));
     }
 
+    @Test
+    public void testConnectionError_performsFailover_testHarder() throws Exception {
+        ExecutorService executorService = Executors.newFixedThreadPool(10);
```
bavardage: n.b. test passes if this is set to 1

bavardage: (and fails reliably locally with this set to 2)

bavardage: or to be explicit, adding:
`System.out.println("failed on : " + call.request().url() + " new url is : " + redirectTo);` just after `RemotingOkHttpCall:148` prints:

```
failed on : http://localhost:60106/string new url is : Optional[http://localhost:60106/string]
failed on : http://localhost:60106/string new url is : Optional[http://localhost:60105/string]
failed on : http://localhost:60106/string new url is : Optional[http://localhost:60106/string]
failed on : http://localhost:60105/string new url is : Optional[http://localhost:60105/string]
failed on : http://localhost:60106/string new url is : Optional[http://localhost:60106/string]
failed on : http://localhost:60106/string new url is : Optional[http://localhost:60106/string]
```

bavardage:
will try to come up with a less awful way of fixing this, not convinced we want to be sharing the url selector across the whole client in the way we are right now

bavardage:
@uschi2000 would gladly get your thoughts here

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -99,7 +99,13 @@ private static String switchWsToHttp(String url) {
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl current) {
         int index = currentUrl.updateAndGet(operand -> (operand + 1) % baseUrls.size());
-        return redirectTo(current, baseUrls.get(index));
+        Optional<HttpUrl> newUrl = redirectTo(current, baseUrls.get(index));
+
+        if (newUrl.map(url -> url.equals(current)).orElse(false) && baseUrls.size() > 1) {
+            return redirectToNext(current);
```
uschi2000: this potentially unbounded recursion is fishy. I understand that the underlying problem is that multiple callers may concurrently call redirectToNext and thus independently increment the currentUrl counter?

bavardage: yep this fix is awful (commit message is "bad fix"), I just wanted to see if it did fix it

yep exactly - so you end up with the retry url being the same as the one you just failed on. It's not easy to instead do it with thread locals because a call bounces around between some number of threads.

also wonder if we should instead just have a url selector impl per call, but maybe there was a reason we didn't do that to start with?

uschi2000: it's tricky because you do want some information shared between calls:
```
client.foo(); // fails on server1, then tries server2 and succeeds
client.foo(); // should try server2 only
```
I think that's desired behavior? maybe round-robin is too naive an implementation here and instead we should keep per-url failure counts that age age off over time.

uschi2000: pseudo code:
```
// select URL
map<url, failedCount> urls = ...;  // failedCounts decay over time
set<url> goodUrls = urls.find(failedCount < 3);
return goodUrls.random();

// use URL
if (call(url) fails) {
  urls.get(url).incrementFailedCount();
} else {
  urls.get(url).resetFailedCount();
}
```

JacekLach: Counts don't seem necessary; if you just want to change default index on failure, all you have to do is to always redirect to the 'next' url after a failure (no matter what the current default is), and update default to what you redirected to. 

```
    @Override
    public Optional<HttpUrl> redirectToNext(HttpUrl current) {
        int index = (indexFor(current) + 1) % baseUrls.size();
        currentDefault.set(index)
        return redirectTo(current, baseUrls.get(index));
    }
```

OTOH I think there's an argument for rotating default urls on successful requests too (to avoid hot-spotting single node). In which case having fail counts makes more sense - you can track which nodes are failing more and send requests to the ones that aren't failing. But I think the above snippet solves this issue too.

uschi2000:
We’ve discussed this: there’s also an argument for not rotating URLs, namely to leverage hot caches.

JacekLach:
Sounds fair. I wasn't suggesting making that change here either, just saying when having counts might be more useful than simply making `redirectToNext` ignore current default and always return `next(currentUrl)`

uschi2000:
Yep

bavardage:
@uschi2000 minimal fix (without the infinite recursion) based on @JacekLach's suggestion - think there's potentially flup work here with counts and rotating vs not, but ideally that's eaten by something like a service mesh instead

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -98,7 +98,8 @@ private static String switchWsToHttp(String url) {
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl current) {
-        int index = currentUrl.updateAndGet(operand -> (operand + 1) % baseUrls.size());
+        int index = currentUrl.updateAndGet(
+                (currentUrlIndex) -> (indexFor(current).orElse(currentUrlIndex) + 1) % baseUrls.size());
```
JacekLach: nit: running the `indexFor` inside the `compareAndSet` loop seems like it makes retries more likely, but maybe I'm prematurely optimising.

bavardage: yeah, pulled it out

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -98,7 +98,9 @@ private static String switchWsToHttp(String url) {
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl current) {
-        int index = currentUrl.updateAndGet(operand -> (operand + 1) % baseUrls.size());
+        Optional<Integer> currentIndex = indexFor(current);
+        int index = currentUrl.updateAndGet(
+                (currentUrlIndex) -> (currentIndex.orElse(currentUrlIndex) + 1) % baseUrls.size());
```
uschi2000: also needs deterministic tests

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -98,7 +98,9 @@ private static String switchWsToHttp(String url) {
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl current) {
-        int index = currentUrl.updateAndGet(operand -> (operand + 1) % baseUrls.size());
+        Optional<Integer> currentIndex = indexFor(current);
```
uschi2000: can you leave a comment indicating what the purpose of this logic is?

uschi2000 (approved):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
bavardage: pushed deterministic test for UrlSelectorImpl

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
bavardage: added a comment, not entirely sure how clear it is
(also renamed the argument to be a little less confusing - url argument named `current` vs int field named `currentUrl`)

PR #666, opened 2018-02-09
Generalise logic persisting Traces across threads to a RequestContext

cbrockington:
Allows consumers of http-remoting to leverage the ExecutorService wrapping as their own persistent state store.


cbrockington:
Leaves the current tracing public api unchanged, however deprecates certain methods which should now directly use there context equivalents.

markelliot:
what's the specific use-case you're going for? don't think we actually want this to be a broad-reaching utility, carrying context this way is pretty undesirable (we do it with tracing only because we have to)

cbrockington:
I agree that in general, you don't want to be promoting the use of global state, though (as is the case with Tracing) I think there are other useful features which more or less require it.

To be more specifically, I would like to start building out frameworks for indirectly mutating product operations to provide better tools for testing.  For example, deliberately rerouting requests to one product url to another (forcing blue products in blue/green deploys to default to talking to each other), or potentially in the future injecting communication failures to test overall product resilience.

markelliot:
I think it's preferable for us to work through the specific cases rather than proceed here.

As an example, in the blue/green setup, that's the kind of routing decision we can make via our service mesh.

cbrockington:
Taking the example of failure injection then, what would you recommend?  My intention was for other core libraries to be able to consistently simulate different interactions across a request (for instance, cache misses).  Without anything at the http-remoting layer, it becomes very difficult for any libraries other applications are using to ensure persistence across requests.

cbrockington:
I'm going to go build out the other pieces with a proposal for how the would connect, and maybe this just turns into something with a specific name, rather than generalised / exposed.

PR #675, opened 2018-02-16
Jersey maps exceptions to json using standard mechanisms

carterkozak:
This allows us to write string responses as json for
compatibility with the feign client.


uschi2000:
can you add a test that goes through Jersey?

carterkozak:
@uschi2000 Updated.

uschi2000 (approved):

uschi2000:
thanks

PR #676, opened 2018-02-18
Updated CipherSuites.FAST_CIPHER_SUITES

carterkozak:
Now FAST_CIPHER_SUITES includes
TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 which is not provided
by default by current versions of the JVM, however can be taken
advantage of by using a native JSSE implementation.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
+++ b/http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
@@ -9,6 +9,7 @@ import com.google.common.collect.ImmutableList;
 public final class CipherSuites {
 
     private static final ImmutableList<String> FAST_CIPHER_SUITES = ImmutableList.of(
+            "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
             "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384",
             "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",
             "TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384",


carterkozak:
@uschi2000

http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
```diff
@@ -9,6 +9,7 @@
 public final class CipherSuites {
 
     private static final ImmutableList<String> FAST_CIPHER_SUITES = ImmutableList.of(
+            "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
```
uschi2000, line 12: @markelliot @ellisjoe Was there any prior discussion around this one? I don't remember it.

markelliot: it's something conscrypt supports, but my understanding was that that's not something available with present versions of dependencies in this project

ellisjoe: Looks like this is already in our FAST_CIPHER_SUITES list: https://github.com/palantir/http-remoting/pull/676/files#diff-2ec1c79fe668d67c7e773ddd0853293aR26

Other info: It appears to be what [CloudFlare](https://blog.cloudflare.com/it-takes-two-to-chacha-poly/) is using and only has native support (either through Conscrypt or OpenSSL) at the moment. Agree with @markelliot that I don't believe there's currently a configuration of remoting that can use this, though the latest version of OkHttp has Conscrypt support which would be able to.

carterkozak: OkHttp should support the cipher if conscypt is in use, but not http2/alpn until the okhttp bump.

uschi2000: I presume we picked the existing order for a reason. Perf?

PR #681, opened 2018-02-26
deprecate `completeSpan`, prefer `fastCompleteSpan`

bavardage:


tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -152,14 +152,20 @@ public static void fastCompleteSpan(Map<String, String> metadata) {
     /**
      * Completes and returns the current span (if it exists) and notifies all {@link #observers subscribers} about the
      * completed span.
+     *
+     * @deprecated Prefer {@link #fastCompleteSpan()}
```
uschi2000, line 156: - should move over the docs
- explain why it's preferred?

iamdanfox:
Closing out as this code now lives in [palantir/tracing-java](https://github.com/palantir/tracing-java)

PR #683, opened 2018-02-27
Fix outdated SafeLogging note in readme#error-propagation

JacekLach:
As of https://github.com/palantir/http-remoting-api/commit/c637211db160761d0f5ad56f375ef45df402f80a the api contract is to send both safe and unsafe args over the wire, and clients cannot assume error responses are safe to log entirely.


JacekLach:
CC @iamdanfox @gracew  - I can't assign reviewers.

iamdanfox (approved):
Good catch, thanks @JacekLach

PR #686, opened 2018-02-28
Change client behavior on 503 from retry to failover

gregakinman:
@uschi2000 

Fixes https://github.com/palantir/http-remoting/issues/678

### DIFF
--- a/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
+++ b/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
@@ -118,10 +118,23 @@ public final class JaxRsClientFailoverTest extends TestBase {
 
     @Test
     public void testQosError_performsFailover() throws Exception {
-        server1.enqueue(new MockResponse().setResponseCode(503));
+        server1.enqueue(new MockResponse().setResponseCode(UNAVAILABLE_RESPONSE_CODE));
         server1.enqueue(new MockResponse().setBody("\"foo\""));
-        server2.enqueue(new MockResponse().setBody("\"foo\""));
+        server2.enqueue(new MockResponse().setBody("\"bar\""));
 
-        assertThat(proxy.string(), is("foo"));
+        assertThat(proxy.string(), is("bar"));
+    }
+
+    @Test
+    public void testQosError_performsRetryWithOneNode() throws Exception {
+        server1.enqueue(new MockResponse().setResponseCode(UNAVAILABLE_RESPONSE_CODE));
+        server1.enqueue(new MockResponse().setBody("\"foo\""));
+
+        TestService anotherProxy = JaxRsClient.create(TestService.class, AGENT, ClientConfiguration.builder()
+                .from(createTestConfig("http://localhost:" + server1.getPort()))
+                .maxNumRetries(2)
+                .build());
+
+        assertThat(anotherProxy.string(), is("foo"));
     }
 }
--- a/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
+++ b/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
@@ -118,7 +118,7 @@ public final class JaxRsClientFailoverTest extends TestBase {
 
     @Test
     public void testQosError_performsFailover() throws Exception {
-        server1.enqueue(new MockResponse().setResponseCode(UNAVAILABLE_RESPONSE_CODE));
+        server1.enqueue(new MockResponse().setResponseCode(503));
         server1.enqueue(new MockResponse().setBody("\"foo\""));
         server2.enqueue(new MockResponse().setBody("\"bar\""));
 
@@ -127,7 +127,7 @@ public final class JaxRsClientFailoverTest extends TestBase {
 
     @Test
     public void testQosError_performsRetryWithOneNode() throws Exception {
-        server1.enqueue(new MockResponse().setResponseCode(UNAVAILABLE_RESPONSE_CODE));
+        server1.enqueue(new MockResponse().setResponseCode(503));
         server1.enqueue(new MockResponse().setBody("\"foo\""));
 
         TestService anotherProxy = JaxRsClient.create(TestService.class, AGENT, ClientConfiguration.builder()
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
@@ -29,7 +29,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.function.Consumer;
 import java.util.function.Supplier;
 import okhttp3.Call;
 import okhttp3.Callback;
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
@@ -240,15 +240,24 @@ final class RemotingOkHttpCall extends ForwardingCall {
 
             @Override
             public Void visit(QosException.RetryOther exception) {
-                redirectToUrl(
-                        callback,
-                        call,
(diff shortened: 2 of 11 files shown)


uschi2000:
this looks good. i'll have to run for dinner now, but will take a more in-depth look later.

gregakinman:
Ugh looks like the build failed :( I’ll be able to take a look later tonight

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
```diff
@@ -118,10 +118,23 @@ public void testConnectionError_performsFailoverOnDnsFailure() throws Exception
 
     @Test
     public void testQosError_performsFailover() throws Exception {
-        server1.enqueue(new MockResponse().setResponseCode(503));
+        server1.enqueue(new MockResponse().setResponseCode(UNAVAILABLE_RESPONSE_CODE));
```
uschi2000: think I prefer the 503 here

uschi2000: think I prefer the 503 here

uschi2000: can you add a similar test for okhttp-clients?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -239,48 +239,44 @@ public Void visit(QosException.Throttle exception) {
...
-                    callback.onFailure(call,
-                            new IOException("Failed to complete the request due to a "
-                                    + "server-side QoS condition: 503", exception));
-                } else {
-                    log.debug("Rescheduling call after backoff",
-                            SafeArg.of("backoffMillis", backoff.get().toMillis()));
-                    scheduleExecution(callback, backoff.get());
-                }
-                return null;
+                return redirectToUrl(callback, call,
```
uschi2000: I'd like to retry another host *with backoff*

uschi2000 (changes requested):
should perform backoff on 503

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
gregakinman: sure

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
gregakinman: Are you thinking of the single-node case here? I thought about this and ended up going this way just to index on the multi-node case - if we get a 503 from one node, we have no data on the health of the other nodes, thus we can safely assume (as we always do when we are issuing the first try of a request) that any other node is healthy right now, and we can try immediately with no backoff. But in the single-node case we'd definitely want to back off. But the logic of how many URLs there are shouldn't be present at this level... and I guess you could run into the case where _all_ nodes are unavailable due to load, and the client is getting bounced around, in which case it still makes sense to back off. And backing off the first time or two probably isn't the worst penalty to pay. Okay.

uschi2000: exactly

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -212,10 +213,10 @@ public void onResponse(Call call, Response response) throws IOException {
     }
 
     @SuppressWarnings("FutureReturnValueIgnored")
-    private void scheduleExecution(Callback callback, Duration backoff) {
+    private void scheduleExecution(Callback callback, Duration backoff, Runnable execution) {
```
uschi2000: callback is now unused?

uschi2000: prefer to order `(execution, backoff)`

gregakinman: whoops, fixed

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -274,13 +263,38 @@ public Void visit(QosException.Unavailable exception) {
...
                             SafeArg.of("backoffMillis", backoff.get().toMillis()));
-                    scheduleExecution(callback, backoff.get());
+                    redirectToUrl(
+                            callback,
+                            call,
+                            // Redirect to the "next" URL, whichever that may be, after backing off.
+                            () -> urls.redirectToNext(request().url()),
+                            "Failed to determine valid redirect URL for base URLs " + urls.getBaseUrls(),
+                            request -> scheduleExecution(() -> client.newCallWithMutableState(request,
+                                    backoffStrategy, maxNumRelocations - 1).enqueue(callback), backoff.get()));
```
uschi2000: don't think we need to decrease maxNumRelocations in this case. backoff already limits the number of retries for 503s.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -274,13 +263,38 @@ public Void visit(QosException.Unavailable exception) {
...
                 }
                 return null;
             }
         };
     }
 
+    private void redirectToUrl(Callback callback, Call call, Supplier<Optional<HttpUrl>> optionalUrlSupplier,
```
uschi2000: think this construction is pretty hard to grok, maybe be more readable with a little bit of code duplication in the visit() methods. summarizing:
- RetryOther should fail if maxNumRetries <= 0 or no URL can be determined
- Unavailable should fail if backoff is empty or no URL can be determined

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -401,18 +402,17 @@ public void handlesQos_canRetryLaterAndThenRedirect() throws Exception {
...
 
     @Test
-    public void handlesQos_memorizedCurrentUrlBetweenCalls() throws Exception {
-        // First hits server,then 308 redirects to server2, then retries, waits on 503, then retries server2 again.
-        server.enqueue(new MockResponse().setResponseCode(308).addHeader(HttpHeaders.LOCATION, url2));
-        server2.enqueue(new MockResponse().setResponseCode(503));
-        server2.enqueue(new MockResponse().setResponseCode(200).setBody("foo"));
+    public void handlesQos_failsOverToAnotherUrl() throws Exception {
+        server.enqueue(new MockResponse().setResponseCode(503));
+        server.enqueue(new MockResponse().setBody("foo"));
```
uschi2000: remove this line

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -401,18 +402,17 @@ public void handlesQos_canRetryLaterAndThenRedirect() throws Exception {
     }
 
     @Test
-    public void handlesQos_memorizedCurrentUrlBetweenCalls() throws Exception {
-        // First hits server,then 308 redirects to server2, then retries, waits on 503, then retries server2 again.
-        server.enqueue(new MockResponse().setResponseCode(308).addHeader(HttpHeaders.LOCATION, url2));
-        server2.enqueue(new MockResponse().setResponseCode(503));
-        server2.enqueue(new MockResponse().setResponseCode(200).setBody("foo"));
+    public void handlesQos_failsOverToAnotherUrl() throws Exception {
```
uschi2000: handlesQos_503FailsOverToAnotherUrl

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
gregakinman: yeah realized i was possibly going overboard with the DRY there

uschi2000 (approved):

uschi2000:
thanks, @gregakinman

gregakinman:
np! would like to cut 3.19.0 with this too if you weren't planning to already

PR #690, opened 2018-03-05
Serialize dates with zone IDs

gregakinman:
Currently, the default mapper serializes `ZonedDateTime`s with an offset only. `ZonedDateTime`, however, contains timezone information, and offsets and timezones are different things. Thus, it's necessary to include timezone information in the serialized version of `ZonedDateTime`s in order to accurately encode all the information that's necessary to correctly reconstruct an equivalent object.

The test case was using a timezone that is defined as an offset by `java.time` (EST: https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/java/time/ZoneId.java#L187), so the test conveniently succeeded. However, as we all know, due to the travesty that is daylight savings time/summer time/etc., some timezones resolve to different offsets depending on the time of year, so an offset is not sufficient - the timezone is necessary, too. So, I switched the test case to use `America/Chicago` instead, since that is a timezone that resolves to different offsets at different times of the year. It would fail when using the default mapper, but when using a mapper that writes zone IDs, it succeeds.

(Relevant: https://www.youtube.com/watch?v=-5wpm-gesOY)

@uschi2000

### DIFF
--- a/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
+++ b/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
@@ -116,6 +116,7 @@ public final class ObjectMappers {
                 .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE)
                 .disable(DeserializationFeature.WRAP_EXCEPTIONS)
                 .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
-                .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES);
+                .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
+                .enable(SerializationFeature.WRITE_DATES_WITH_ZONE_ID);
     }
 }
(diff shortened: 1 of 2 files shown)


uschi2000:
I don't think we can do this. The wire format is https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators and not all languages like the idea of the Java zone identifiers.

gregakinman:
So what's the answer for those who need to serialize/deserialize `ZonedDateTime`s correctly? Just realize that it's broken unless this feature is manually enabled?

markelliot:
There’s an issue describing both the problems and the solutions to this. I think we agreed that a compound type would be preferable to a single string format.

gregakinman:
Couldn't find it stated in any issues, but did find it here, which I found when Googling earlier today and it was also linked by @bavardage in an internal PR: https://stackoverflow.com/a/42196445:

> If what you're looking for is a way to include a ZonedDateTime or similar data in an API in a standardized manner, my personal recommendation would be to pass the time zone name in a separate field. That way, each portion of data is as good as it can be. For example in JSON:

```
{
  "timestamp": "2011-12-03T10:15:30+01:00",
  "timezone": "Europe/Paris"
}
```

Closing this in favor of email thread about `ZonedDateTime` vs. `OffsetDateTime`.

PR #692, opened 2018-03-08
Test for sticky failover

gregakinman:
cc @samrogerson 

This might sort of be taken care of already by `UrlSelectorTest`, but also feels right to have something explicit in `OkHttpClientsTest`

@uschi2000


uschi2000 (approved):
thanks!

PR #693, opened 2018-03-16
Feign response streaming

carterkozak:


carterkozak:
@uschi2000 @iamdanfox fysa

PR #694, opened 2018-03-29
Retry failures after backing off first

abless:


justinuang:
@iamdanfox @uschi2000 we need this to release SMM 2.0 HA.

iamdanfox:
@abless if you want to get this pushed through after-hours in London, can you coordinate with @qinfchen who is based in NY?

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -40,7 +40,7 @@
     private static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
     private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(10);
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(10);
-    private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(500);
+    private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
```
uschi2000, line 43: why this change?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/ExponentialBackoff.java
```diff
@@ -48,6 +48,6 @@
         }
 
         int upperBound = (int) Math.pow(2, retryNumber);
-        return Optional.of(backoffSlotSize.multipliedBy(random.nextInt(upperBound)));
+        return Optional.of(Duration.ofNanos(Math.round(backoffSlotSize.toNanos() * random.nextDouble() * upperBound)));
```
uschi2000, line 51: why this change?

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
abless: This PR changes in behavior in that we now always apply a backoff duration when calls encounter IOExceptions (e.g. ConnectExceptions), whereas previous implementation would retry right away. When load-testing rolling upgrades with one of our services we noticed that this would mean that one server (the only one alive) would get hit with lots of requests all at the same time, causing ConnectExceptions.

We are halving the default backoff slot size because the previous implementation of ExponentialBackoff didn't actually adhere to its javadoc - it says it chooses a backoff from the interval of [0, backoffSlotSize * 2^c) but it really chose it from [0, backoffSlotSize * 2^(c-1)) - this has been fixed in the implementation so we're halving the default to provide consistent backoff duration.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/ExponentialBackoff.java
abless: To increase entropy. random.nextInt() has pretty low variance for small number of retries (in particular, for the first retry) and so to make sure that lots of calls don't retry all at the same time we increase the entropy/variance by going more granular.

uschi2000, line 51: the Math.round seems redundant

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/ExponentialBackoffTest.java
```diff
@@ -45,17 +42,14 @@ public void testRetriesCorrectNumberOfTimesAndFindsRandomBackoffWithInExponentia
...
         assertThat(backoff.nextBackoff()).contains(ONE_SECOND.multipliedBy(2));
-        verify(random).nextInt(4);
 
-        when(random.nextInt(anyInt())).thenReturn(3);
-        assertThat(backoff.nextBackoff()).contains(ONE_SECOND.multipliedBy(3));
-        verify(random).nextInt(8);
+        when(random.nextDouble()).thenReturn(1.0);
+        assertThat(backoff.nextBackoff()).contains(ONE_SECOND.multipliedBy(4));
+
+        when(random.nextDouble()).thenReturn(1.0);
```
uschi2000: let's vary the random number returned.

uschi2000 (commented):
ok, sounds good. two nitpicks, otherwise +1

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/ExponentialBackoff.java
abless: Duration.ofNanos() expects a long, since we're using `random.nextDouble()` we need to round.

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/ExponentialBackoffTest.java
abless: Can do, but then it's not as obvious that the backoff is exponential (2, 4, 8) - still want to vary it?

uschi2000: yep. just change the last one to 0.5 --> 4.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/ExponentialBackoff.java
uschi2000: I was thinking to just do `(long) d`, but sure, round also works.

uschi2000 (approved):

justinuang:
Can we get a release too if it's not done yet

gracew:
tagged: https://github.com/palantir/http-remoting/releases/tag/3.20.0

PR #696, opened 2018-04-17
expose HostMetricsRegistry and a global instance for additional HostMetrics

qinfchen:
Specifically for the service dependencies that do not involve http-remoting clients.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
@@ -29,13 +29,22 @@ import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-final class HostMetricsRegistry {
+public final class HostMetricsRegistry {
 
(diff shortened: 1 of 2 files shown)


uschi2000:
I don't think I want non-remoting clients to use a remoting api.

qinfchen:
@ellisjoe, should the registry and metrics live in -api stead?

markelliot:
can you ping us internally with some details on the goals of this? it's not clear why non-remoting clients should be using the _remoting_ host metrics registry; this is private exactly because we expect it to strictly contain _remoting_ client metrics.

ellisjoe:
Use case is to provide metrics on Cassandra connections which use the thrift client rather than a remoting client. My original recommendation was to just expose [HostMetrics](https://github.com/palantir/http-remoting/blob/develop/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetrics.java).

ellisjoe:
Main thing here is reporting our status based on the current state of our remote dependencies. If HostMetrics isn't the right thing to expose as the common interface here, should consider what is.

qinfchen:
Closing this as the usage of host metrics is intended for http-remoting clients only.

PR #703, opened 2018-05-01
Expose method to wrap runnable that is run with given traceId

a-k-g:
@uschi2000

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
@@ -164,7 +164,7 @@ public final class Tracers {
      * instead of any trace already set on the thread used to execute the runnable. Each execution of the runnable
      * will use a new {@link Trace tracing state} with the same given traceId.
      */
-    public static Runnable wrapWithTrace(Runnable delegate, String traceId) {
+    public static Runnable wrapWithAlternateTraceId(String traceId, Runnable delegate) {
         return () -> {
             // clear the existing trace and keep it around for restoration when we're done
             Trace originalTrace = Tracer.getAndClearTrace();
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
@@ -159,6 +159,27 @@ public final class Tracers {
     }
 
     /**
+     * Wraps the given {@link Runnable} such that it creates a fresh {@link Trace tracing state with the given traceId}
+     * for its execution. That is, the trace during its {@link Runnable#run() execution} will use the traceId provided
+     * instead of any trace already set on the thread used to execute the runnable. Each execution of the runnable
+     * will use a new {@link Trace tracing state} with the same given traceId.
+     */
+    public static Runnable wrapWithTrace(Runnable delegate, String traceId) {
+        return () -> {
+            // clear the existing trace and keep it around for restoration when we're done
+            Trace originalTrace = Tracer.getAndClearTrace();
+
+            try {
+                Tracer.initTrace(Optional.empty(), traceId);
+                delegate.run();
+            } finally {
+                // restore the trace
+                Tracer.setTrace(originalTrace);
+            }
+        };
+    }
+
+    /**
      * Wraps a given callable such that its execution operates with the {@link Trace thread-local Trace} of the thread
      * that constructs the {@link TracingAwareCallable} instance rather than the thread that executes the callable.
      * <p>
--- a/tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
+++ b/tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
@@ -245,6 +245,53 @@ public final class TracersTest {
     }
 
     @Test
+    public void testWrapRunnableWithTrace_traceStateInsideRunnableUsesGivenTraceId() {
+        String traceIdBeforeConstruction = Tracer.getTraceId();
+
+        List<String> traceIds = Lists.newArrayList();
+
+        String traceIdToUse = "someTraceId";
+        Runnable wrappedRunnable = Tracers.wrapWithTrace(() -> {
+            traceIds.add(Tracer.getTraceId());
+        }, traceIdToUse);
+
+        wrappedRunnable.run();
+        wrappedRunnable.run();
+
+        String traceIdFirstCall = traceIds.get(0);
+        String traceIdSecondCall = traceIds.get(1);
+
+        String traceIdAfterCalls = Tracer.getTraceId();
+
+        assertThat(traceIdFirstCall)
+                .isNotEqualTo(traceIdBeforeConstruction)
+                .isNotEqualTo(traceIdAfterCalls)
+                .isEqualTo(traceIdToUse);
+
+        assertThat(traceIdSecondCall)
+                .isNotEqualTo(traceIdBeforeConstruction)
+                .isNotEqualTo(traceIdAfterCalls)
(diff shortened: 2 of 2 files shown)


tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -244,6 +244,53 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
...
+    public void testWrapRunnableWithTrace_traceStateRestoredWhenThrows() {
+        String traceIdBeforeConstruction = Tracer.getTraceId();
+
+        Runnable rawRunnable = () -> {
+            throw new IllegalStateException();
+        };
+        Runnable wrappedRunnable = Tracers.wrapWithTrace(rawRunnable, "someTraceId");
+
+        assertThatThrownBy(() -> wrappedRunnable.run()).isInstanceOf(IllegalStateException.class);
+
```
uschi2000: nit: remove \n

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -244,6 +244,53 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
...
+                .isEqualTo(traceIdToUse);
+
+        assertThat(traceIdBeforeConstruction)
+                .isEqualTo(traceIdAfterCalls);
+    }
+
+    @Test
+    public void testWrapRunnableWithTrace_traceStateRestoredWhenThrows() {
+        String traceIdBeforeConstruction = Tracer.getTraceId();
+
```
uschi2000: nit: remove \n

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -244,6 +244,53 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
...
+            traceIds.add(Tracer.getTraceId());
+        }, traceIdToUse);
+
+        wrappedRunnable.run();
+        wrappedRunnable.run();
+
+        String traceIdFirstCall = traceIds.get(0);
+        String traceIdSecondCall = traceIds.get(1);
+
+        String traceIdAfterCalls = Tracer.getTraceId();
```
uschi2000: inline

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -244,6 +244,53 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
...
+
+        String traceIdToUse = "someTraceId";
+        Runnable wrappedRunnable = Tracers.wrapWithTrace(() -> {
+            traceIds.add(Tracer.getTraceId());
+        }, traceIdToUse);
+
+        wrappedRunnable.run();
+        wrappedRunnable.run();
+
+        String traceIdFirstCall = traceIds.get(0);
```
uschi2000: inline

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
```diff
@@ -244,6 +244,53 @@ public void testWrapRunnableWithNewTrace_traceStateRestoredWhenThrows() throws E
...
+        String traceIdBeforeConstruction = Tracer.getTraceId();
+
+        List<String> traceIds = Lists.newArrayList();
+
+        String traceIdToUse = "someTraceId";
+        Runnable wrappedRunnable = Tracers.wrapWithTrace(() -> {
+            traceIds.add(Tracer.getTraceId());
+        }, traceIdToUse);
+
+        wrappedRunnable.run();
```
uschi2000: what's the purpose of running it twice?

a-k-g: Yeah, no point. Removed the second run.

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
a-k-g: done

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
a-k-g: It's used twice. Also, the assertions are more readable this way. Renamed to `traceIdAfterCall` since I'm doing just one call now.

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
a-k-g: done

tracing/src/test/java/com/palantir/remoting3/tracing/TracersTest.java
a-k-g: done

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -158,6 +158,27 @@ public static Runnable wrapWithNewTrace(Runnable delegate) {
         };
     }
 
+    /**
+     * Wraps the given {@link Runnable} such that it creates a fresh {@link Trace tracing state with the given traceId}
+     * for its execution. That is, the trace during its {@link Runnable#run() execution} will use the traceId provided
+     * instead of any trace already set on the thread used to execute the runnable. Each execution of the runnable
+     * will use a new {@link Trace tracing state} with the same given traceId.
+     */
+    public static Runnable wrapWithTrace(Runnable delegate, String traceId) {
```
a-k-g: do you think the usages of this would be more readable if we had the `traceId` be the first param?

```
Runnable wrappedRunnable = Tracers.wrapWithTrace(() -> {
    ...
}, traceIdToUse);
```
vs
```
Runnable wrappedRunnable = Tracers.wrapWithTrace(traceId, () -> {
    ...
});
```

markelliot:
it's a little strange to me that we're providing this facility/the only other place we're presently overriding the trace state in this way (where we use an externally provided traceId) is to init the tracing system. Why does this avoid using something like this: https://github.com/palantir/http-remoting/blob/cb2ce26d71bdb55a798b6a3fe2c1e4c68fe3bcb1/jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/TraceEnrichingFilter.java#L54

markelliot:
in particular, in your current setup, where you're trying to plumb through a particular traceId to start, I think it'd be preferable you replace the trace stack explicitly through an initTrace call rather than by offering this Runnable utility, which is more likely to be misused.

a-k-g:
You're right that we can use the `initTrace` to init the system with a provided `traceId`. However, if there was any trace already initialized on the thread, there isn't any way to restore it. This utility method does the restore.

markelliot:
Ok, I guess `wrapWithTrace` is probably the wrong name for the method for what we're trying to do/it still encourages something pretty strange. Maybe `wrapWithAlternateTraceId`?

a-k-g:
Renamed to `wrapWithAlternateTraceId`

a-k-g:
@uschi2000 could you please have another look at this?

iamdanfox (approved):

uschi2000 (approved):

a-k-g:
Can we get a release with this?

iamdanfox:
@a-k-g Building: 
- https://circleci.com/gh/palantir/http-remoting/2032
- https://github.com/palantir/http-remoting/releases/tag/3.21.0

PR #705, opened 2018-05-02
Customizable okhttp builder

dansanduleac:
I'd like to be able to create remoting clients where I change some details about the okhttp client, such as
* dispatcher
* connection pool
* cache

Use case is I will create a lot of clients, most of which will share the same underlying set of hosts but have different URIs.
I'd like all these clients to share the above objects.
I'd also like to configure the cache such that I can make it an on-disk cache, and it needs to be shared in order to enforce a maximum cache size across all clients.

Currently, each new okhttp client (either created directly or through the jaxrs / retrofit2 factories) creates its own new dispatcher and connection pool. This means we can't enforce global limits on the 
* total number of connections
* number of connections per host

via the dispatcher, nor can we reuse existing connections to the same hosts from a different client.
They also don't allow configuring the cache, leaving it as the default, so it's impossible to configure a bounded on-disk cache.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
@@ -36,11 +36,13 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.function.UnaryOperator;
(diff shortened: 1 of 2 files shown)


uschi2000:
Dan, instead of this change, should we use the same connection pool and dispatcher in all clients?

dansanduleac:
Yep - I'm super happy to do that.
But I still need a configurable cache, which I think is a generally good idea. Any remoting client that hits cacheable endpoints could benefit from that.
It needs to be configurable so I can give it a path on disk.
I made a PR just for that in #707.

dansanduleac:
Closing in favour of #706 and #707

PR #706, opened 2018-05-03
Shared dispatcher and connection pool

dansanduleac:
Decided to share dispatcher and connection pool for all clients produced by the `OkHttpClients` class. This way we can enforce limits on connections globally, and allow reusing connections to the same host even if they come from different clients.

@uschi2000


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -54,6 +54,16 @@
      */
     private static final ExecutorService executionExecutor = Tracers.wrap(new Dispatcher().executorService());
 
+    /**
+     * Shared dispatcher with static executor service.
```
uschi2000: super nit: can use `/** One-line javadoc. */`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -54,6 +54,16 @@
...
+     */
+    private static final Dispatcher dispatcher = createDispatcher();
```
uschi2000: maybe move the initialization to a `static {}` block so it's clear that the createDispatcher method shouldn't be used?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -212,6 +221,7 @@ private static UserAgent augmentUserAgent(UserAgent agent, Class<?> serviceClass
                 ConnectionSpec.CLEARTEXT);
     }
 
+    @SuppressWarnings("checkstyle:HiddenField")
```
uschi2000: that seems avoidable

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
dansanduleac: sure. I just followed the style of the above field's javadoc, which admittedly just barely doesn't fit on 1 line.

uschi2000 (approved):

PR #707, opened 2018-05-03
Customizable on disk cache

dansanduleac:
@uschi2000


uschi2000:
Before I look at the code: I'd prefer application-level caching over HTTP-level caching. The rationale is that we want our RPC clients to be as simple as possible. Further, the application logic is easier to understand if caching logic is made explicit. Even further, we have relatively poor (read: mostly non-existing) static knowledge about which endpoints are cacheable at all, so I think it's a hard sell to enable global caching with uncertain behavior.

uschi2000:
Discussed offline, punting on this for now.

PR #708, opened 2018-05-07
add service port to HostMetrics

tabboud:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/DefaultHostMetrics.java
```diff
@@ -43,10 +44,11 @@
 
     private volatile long lastUpdateEpochMillis;
 
-    /** Creates a metrics registry for calls from the given service to the given host. */
-    DefaultHostMetrics(String serviceName, String hostname, Clock clock) {
+    /** Creates a metrics registry for calls from the given service to the given host and url. */
+    DefaultHostMetrics(String serviceName, String hostname, String url, Clock clock) {
```
uschi2000: not backwards-compatible

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetrics.java
```diff
@@ -36,6 +36,11 @@
      */
     String hostname();
 
+    /**
+     * The url used to gather the metrics.
+     */
+    String url();
```
uschi2000: where is this used?

tabboud: In the health checks for remote services in witchcraft. We currently use the hostname rather than the url.

uschi2000: the URL contains potentially unsafe information (e.g., path and query params). how are you dealing with those?

tabboud: Downstream consumers would be parsing this and could handle the unsafe params. Although to avoid this altogether, we could reconstruct the uri via the `scheme`, `hostname` and `port`, which would be enough to uniquely identify it without the full url.

So rather than adding the url, just change the hostname field to uri of the above.

uschi2000: we can add the port if it's critical for disambiguation. not a fan of the "parse later" proposal, it's very brittle.

ellisjoe: yeah, curious what this information will be used for?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
```diff
@@ -33,33 +33,38 @@
...
+    void record(String serviceName, String hostname, int port, int statusCode, long micros) {
         try {
-            hostMetrics.getUnchecked(ImmutableServiceAndHost.of(serviceName, hostname)).record(statusCode, micros);
+            hostMetrics.getUnchecked(
+                    ImmutableServiceHostAndPort.of(serviceName, hostname, port)).record(statusCode, micros);
         } catch (Exception e) {
-            log.warn("Unable to record metrics for host", UnsafeArg.of("hostname", hostname));
+            log.warn("Unable to record metrics for host and url",
+                    UnsafeArg.of("hostname", hostname),
+                    UnsafeArg.of("url", port));
```
ellisjoe: first arg should be `port`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
```diff
@@ -33,33 +33,38 @@
...
 
-    void record(String serviceName, String hostname, int statusCode, long micros) {
+    void record(String serviceName, String hostname, int port, int statusCode, long micros) {
         try {
-            hostMetrics.getUnchecked(ImmutableServiceAndHost.of(serviceName, hostname)).record(statusCode, micros);
+            hostMetrics.getUnchecked(
+                    ImmutableServiceHostAndPort.of(serviceName, hostname, port)).record(statusCode, micros);
         } catch (Exception e) {
-            log.warn("Unable to record metrics for host", UnsafeArg.of("hostname", hostname));
+            log.warn("Unable to record metrics for host and url",
```
ellisjoe: `port`

ellisjoe (approved):
fine by me

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
iamdanfox: port is always safe btw, it's a plain number.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
```diff
@@ -33,33 +33,38 @@
...
-    void recordIoException(String serviceName, String hostname) {
+    void recordIoException(String serviceName, String hostname, int port) {
         try {
-            hostMetrics.getUnchecked(ImmutableServiceAndHost.of(serviceName, hostname)).recordIoException();
+            hostMetrics.getUnchecked(ImmutableServiceHostAndPort.of(serviceName, hostname, port)).recordIoException();
         } catch (Exception e) {
-            log.warn("Unable to record IO exception for host", UnsafeArg.of("hostname", hostname));
+            log.warn("Unable to record IO exception for host and url",
+                    UnsafeArg.of("hostname", hostname),
+                    UnsafeArg.of("url", port));
```
iamdanfox: `s/urls/port/`

iamdanfox (changes requested):
Just some tweaks to the args!

uschi2000:
looks good modulo joe's + dan's comments

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
tabboud: whoops, fail on my part to rename.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
tabboud: done and converted to `SafeArg`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/HostMetricsRegistry.java
tabboud: done

iamdanfox (approved):
:+1: LGTM

PR #709, opened 2018-05-07
use SafeIoException to log body-not-serializable message and response code

yifeih:
Addresses: https://github.com/palantir/http-remoting/issues/702

So, currently, our error stack traces look something like this: 

```
feign.RetryableException: {throwable0_message}
	.... <stacktrace>
Caused by: java.io.IOException: {throwable1_message}
	at ... <stacktrace>
```

I believe that, since the outer message is a feign exception, the SafeArgs from the SafeIoException won't be propagate up? so it's better to put the response code in the SafeIoException message instead of the args? Please correct me if I'm understanding this though. Thanks!


iamdanfox:
@yifeih I think this is a good start for now - there's open work on our internal logging library to capture the safeargs from throwable causes, which should solve the problem you described above!

Just unit tests that need to be fixed up!

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
```diff
@@ -38,8 +40,9 @@
             String body = response.body() != null && response.body().byteStream() != null
                     ? toString(response.body().byteStream())
                     : "<empty>";
-            return Optional.of(new IOException(String.format("Error %s. "
-                    + "(Failed to parse response body as SerializableError.) Body:%n%s", response.code(), body)));
+            return Optional.of(new SafeIoException(
+                    String.format("Error %s. (Failed to parse response body as SerializableError.) ", response.code()),
```
justinuang: Lets not use string format anymore, instead just having "Failed to parse response body as SerializableError".

justinuang (changes requested):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/IoExceptionResponseHandler.java
yifeih: Wouldn't that lose the error code? The SafeArgs currently don't yet get propagated up to the feign exception (see comment above)

iamdanfox: @justinuang I think your suggestion makes sense once we've changed our internal logging implementation to pull out the args, but until that's implemented I think it's OK to run with the String.format!

iamdanfox (approved):

yifeih:
hey @iamdanfox, I don't think this is working. we've added this to our service, but we're still seeing:
```
feign.RetryableException: {throwable0_message}
...
Caused by: java.io.IOException: {throwable1_message}
```
it seems like we might need the safe-arg/message propagation in order for this to work?

helenyuyu:
It looks like the body is completely dropped from the logs altogether, which makes debugging completely impossible now.

yifeih:
i wasn't sure whether the body always contained safe-loggable content, so i made it an unsafe arg

helenyuyu:
I think both safe and unsafe args are dropped from logs.

yifeih:
@helenyuyu I added you to this thread, but you might not have seen it: [redacted internal url]. It's probably because args are not being propagated from the safe exception to the one wrapping it

PR #710, opened 2018-05-10
Add CloseableTracer utility class

raiju:
Adds a convenience class to make adding tracers cleaner.

@j-baker @uschi2000

### DIFF
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
@@ -42,23 +42,6 @@ public final class CloseableTracer implements AutoCloseable {
         return INSTANCE;
     }
 
-    /**
-     * Like {@link #startSpan(String)}, but opens a span of the explicitly given {@link SpanType span type}.
-     */
-    public static CloseableTracer startSpan(String operation, SpanType spanType) {
-        Tracer.startSpan(operation, spanType);
-        return INSTANCE;
-    }
-
-    /**
-     * Opens a new span for this thread's call trace, labeled with the provided operation and parent span. Only allowed
-     * when the current trace is empty.
-     */
-    public static CloseableTracer startSpan(String operation, String parentSpanId, SpanType spanType) {
-        Tracer.startSpan(operation, parentSpanId, spanType);
-        return INSTANCE;
-    }
-
     @Override
     public void close() {
         Tracer.fastCompleteSpan();
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
@@ -25,6 +25,9 @@ import com.palantir.remoting.api.tracing.SpanType;
  *   try (CloseableTracer trace = CloseableTracer.start("traceName")) {
  *       [...]
  *   }
+ * 
+ * Note that network calls using remoting will be automatically wrapped with appropriate traces. Only use this
+ * to add traces to internal code.
  */
 public final class CloseableTracer implements AutoCloseable {
     private static final CloseableTracer INSTANCE = new CloseableTracer();
--- a/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
+++ b/tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
@@ -25,7 +25,7 @@ import com.palantir.remoting.api.tracing.SpanType;
  *   try (CloseableTracer trace = CloseableTracer.start("traceName")) {
  *       [...]
  *   }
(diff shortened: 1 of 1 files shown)


raiju:
The test failure looks unrelated to this PR

tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Wraps the {@link Tracer} methods in a closeable resource to enable the usage of the try-with-resources pattern.
+ *
+ * Usage:
+ *   try (CloseableTracer trace = CloseableTracer.start("traceName")) {
+ *       [...]
+ *   }
+ */
+public final class CloseableTracer implements AutoCloseable {
```
j-baker: maybe just have this be an enum?

raiju: What is the advantage?

raiju: Thinking about it some more, I'm not 100% sure on what you mean. Do you mean something like:
```
enum CloseableTracer implements AutoCloseable {
  INSTANCE;
  static CloseableTracer startSpan(...) { ... }
}
```
?

In that case what would stop the user from accessing `CloseableTracer.INSTANCE` directly? It's also not obvious to me what the advantage would be. Did I misunderstand what you meant?

j-baker: oh, sorry. So I was just suggesting that given that you only want there to be one of these, one thing which people frequently end up doing is making it an enum. But I don't really care and don't want to block this.

j-baker (approved):

tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.palantir.remoting.api.tracing.SpanType;
+
+/**
+ * Wraps the {@link Tracer} methods in a closeable resource to enable the usage of the try-with-resources pattern.
+ *
+ * Usage:
+ *   try (CloseableTracer trace = CloseableTracer.start("traceName")) {
+ *       [...]
+ *   }
```
iamdanfox, line 27: Hey @raiju any chance we could also add one sentence of guidance so people know roughly when they should use this? The main thing I want to convey is that http-remoting calls are already traced, so people don't need to wrap every single call they ever do in this!

iamdanfox (approved):
Approve pending a tiny bit of Javadoc

tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public static CloseableTracer startSpan(String operation) {
+        Tracer.startSpan(operation);
+        return INSTANCE;
+    }
+
+    /**
+     * Like {@link #startSpan(String)}, but opens a span of the explicitly given {@link SpanType span type}.
+     */
+    public static CloseableTracer startSpan(String operation, SpanType spanType) {
```
iamdanfox: `SpanType` allows SERVER_INCOMING, CLIENT_OUTGOING or LOCAL... but we're only ever expecting this to be called with CLIENT_OUTGOING or LOCAL right?

j-baker: probably only LOCAL, here. CLIENT_OUTGOING one would anticipate should be implemented inside the client.

iamdanfox: Hmm, so maybe the single-arg variant would be sufficient then?  `public static CloseableTracer startSpan(String operation)` just does a LOCAL span

raiju: I'm not sure why we'd limit it; the current APIs (`Tracer.[...]`) allow all these calls.

tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
raiju: Is this class that different from https://github.com/palantir/http-remoting/blob/29b83d50358d17e8f57eca62b2790ec6e08ecbf6/tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java#L42 to warrant this?

uschi2000:
Please add test coverage for these kind of changes, even if it's trivial. https://github.com/palantir/http-remoting/pull/718

raiju:
@uschi2000 My bad, thanks for adding!

PR #713, opened 2018-05-18
Upgrade jackson to 2.9.5

robert3005:


jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
```diff
@@ -132,9 +132,9 @@ public void testComplexType() {
                         new GuavaOptionalComplexType(
                                 Optional.absent(),
                                 Optional.absent(),
-                                Paths.get("bar"))),
+                                Paths.get("/bar"))),
                 Optional.of("baz"),
-                Paths.get("foo"));
+                Paths.get("/foo"));
```
uschi2000: why this change?

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8TestServer.java
```diff
@@ -46,6 +47,7 @@
     public final void run(Configuration config, final Environment env) throws Exception {
         env.jersey().register(HttpRemotingJerseyFeature.INSTANCE);
         env.jersey().register(new TestResource());
+        env.jersey().register(new EmptyOptionalTo204ExceptionMapper());
```
uschi2000: why do we need this?

jersey-servers/src/test/resources/test-server.yml
```diff
@@ -6,6 +6,8 @@ server:
   adminConnectors:
     - type: http
       port: 0
+  gzip:
```
uschi2000, line 9: why this change?

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8TestServer.java
robert3005: Dropwizard 1.x converts Optional.empty to 404 which breaks our tests. While 0.x version did nothing special about Optionals we need this now to preserve the expected behaviour of returning 204

jersey-servers/src/test/resources/test-server.yml
robert3005: Dropwziards gzip filter gets in the way of remoting server compression filter. I think they should do the same thing but dropwizard returns different vary headers

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
robert3005: This is explained in https://github.com/FasterXML/jackson-databind/issues/1264 . Unfortunately previously java.nio.file.Path didn't round trip correctly. In order to fix it they changed the serialization format of Path to always write an absolute path. If you pass a relative path and serialize it will make it absolute which is hard to handle in tests

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8TestServer.java
robert3005: https://github.com/dropwizard/dropwizard/pull/2350 added this class in dropwizard

keystores/src/test/java/com/palantir/remoting3/config/ssl/DropwizardSslClientAuthTests.java
```diff
@@ -52,6 +52,8 @@
                     TestConstants.SERVER_KEY_STORE_JKS_PASSWORD),
             ConfigOverride.config("server.applicationConnectors[0].trustStorePath",
                     TestConstants.CA_TRUST_STORE_PATH.toString()),
+            ConfigOverride.config("server.applicationConnectors[0].trustStorePassword",
```
robert3005, line 55: As of https://github.com/eclipse/jetty.project/commit/a8d08df84fb7b6944e3bd11adb724fd05c29e357 truststore password will default to keystore password which will fail upon reading the jks keystore. Internally we require client cert auth truststore to come with passwords

jersey-servers/build.gradle
```diff
@@ -15,4 +15,5 @@ dependencies {
     testCompile "io.dropwizard:dropwizard-testing"
     testCompile "junit:junit"
     testCompile "org.hamcrest:hamcrest-all"
+    testCompile "org.mockito:mockito-core"
```
robert3005, line 18: You have been relying on transitive mockito-core import from dropwizard that's no longer there

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalAwareDecoderTest.java
uschi2000: ok

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8TestServer.java
uschi2000: we'll need to minimally understand (and maybe deduplicate) this with Java8OptionalMessageBodyWriter and NoContentExceptionMapper

uschi2000: (because it's not obvious how to use a HttpRemotingJerseyFeature.INSTANCE with dropwizard to get the desired behavior)

jersey-servers/build.gradle
uschi2000: cheers

jersey-servers/src/test/resources/test-server.yml
uschi2000: this is only for CompressionFilterTest, ja?

keystores/src/test/java/com/palantir/remoting3/config/ssl/SslConfigurationTest.java
```diff
@@ -28,18 +28,18 @@
...
                 .trustStoreType(SslConfiguration.StoreType.JKS)
-                .keyStorePath(Paths.get("keystore.jks"))
+                .keyStorePath(Paths.get("/keystore.jks"))
                 .keyStoreType(SslConfiguration.StoreType.JKS)
                 .keyStoreKeyAlias("alias")
                 .keyStorePassword("password")
                 .build();
-        String deserializedCamelCase = "{\"trustStorePath\":\"truststore.jks\",\"trustStoreType\":\"JKS\","
-                + "\"keyStorePath\":\"keystore.jks\",\"keyStorePassword\":\"password\","
+        String deserializedCamelCase = "{\"trustStorePath\":\"file:///truststore.jks\",\"trustStoreType\":\"JKS\","
```
uschi2000: can old jackson versions deserialize this? (asking because some services (SMM?) may use the SslConfiguration class over the wire, I believe)

keystores/src/test/java/com/palantir/remoting3/config/ssl/DropwizardSslClientAuthTests.java
uschi2000: ok

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8TestServer.java
uschi2000: but then I guess it's called JerseyFeature, not DropwizardFeature. not your fault, let's move on. https://github.com/palantir/http-remoting/issues/714

uschi2000:
looks good modulo the path deserialization wire break question.

jersey-servers/src/test/resources/test-server.yml
robert3005: yes, just for those tests

keystores/src/test/java/com/palantir/remoting3/config/ssl/SslConfigurationTest.java
robert3005: I think I went overboard here with making behaviour the same in both cases. Let me revert that

robert3005: Added extra cases to show the compatibility guarantees

uschi2000 (approved):

uschi2000:
@iamdanfox do you want to take a look, too, or are you good?

extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
```diff
@@ -108,7 +108,6 @@ public static ObjectMapper newCborServerObjectMapper() {
     public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
         return mapper
                 .registerModule(new GuavaModule())
-                .registerModule(new ShimJdk7Module())
```
iamdanfox: If this is the only piece of non-test code that was changed then in theory if people hit sad times with the new jackson, they could conceivably force back to a lower version (and re-add this) right?

j-baker: So for this code, there are two arguments:

1. This code makes remoting Jackson version independent. It's why some internal projects run 2.9. So leaving it in is nice because it gives you version independence.
2. OTOH almost everything uses Jackson 2.7+ now, and we might want to use remoting to leverage that.

error-handling/versions.lock
```diff
@@ -1,13 +1,14 @@
 {
     "compileClasspath": {
         "com.fasterxml.jackson.core:jackson-annotations": {
-            "locked": "2.6.7",
+            "locked": "2.9.5",
             "transitive": [
-                "com.fasterxml.jackson.core:jackson-databind"
+                "com.fasterxml.jackson.core:jackson-databind",
+                "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
```
uschi2000, line 7: in 2.8 or 2.9 a lot of this stuff got renamed, no? specifically, why don't we have to update `extras/jackson-support/build.gradle` to reflect the new maven coordinates of the Guava and Java8 data formats?

robert3005:
@iamdanfox @uschi2000 turns out my debugging wasn't quite right - you can override path deserializer and the code we have just works. I mistakenly thought that what was actually a dropwizard misconfiguration was a jackson issue. I have reverted all the path related changes and instead adjusted DW used in test to use `ObjectMappers#newServerObjectMapper` as jackson message body provider

As far as I can tell the source repos moved around but artifacts didn't so we're all good (I just verified all java 8 related packages manually).

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/GuavaOptionalComplexType.java
```diff
@@ -49,6 +50,15 @@ public Path getPath() {
         return path;
     }
 
+    @Override
+    public String toString() {
```
robert3005, line 54: Added for ease of debugging - you get useful assertions errors with this

uschi2000 (approved):

uschi2000:
thanks, @robert3005

PR #715, opened 2018-05-21
Deprecate unused CompressionFilter

uschi2000:


uschi2000:
@iamdanfox

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/CompressionFilter.java
```diff
@@ -40,7 +40,10 @@
  * This is an alternative compression filter implementation that uses jzlib for compression rather than the java
  * standard library due to performance issues observed on high core machines:
  * https://bugs.openjdk.java.net/browse/JDK-8179001
+ *
+ * @deprecated This class will be removed in the next major release.
```
iamdanfox: It's a bit confusing to see a class will be removed with no explanation of why... could we use the `@deprecated` comment to give some more context/recommend what people should do instead?  Internally, it seems we use `org.eclipse.jetty.server.handler.gzip.GzipHandler` with a min size of 1MB instead of this filter as a workaround for that openjdk bug which remains unfixed.

iamdanfox (approved):

PR #719, opened 2018-05-31
Expose a way to enable a round robin RPC mode in OkHttpClients

justinbchan:


palantirtech:
Thanks for your interest in [palantir/http-remoting](https://github.com/palantir/http-remoting), @justinbchan! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

justinbchan:
Tests are forthcoming -- wanted to get this up for review.

j-baker:
Please can we wire this through via the ClientConfiguration object (which will necessitate changes to the remoting api package as well)? That's our config point right now (e.g. 'allow using gcm ciphers'). This idea seems fine to me otherwise.

uschi2000:
@j-baker the counter point to that proposal is: we don't want different environments to run substantially different config. thus, we want compile-time configurability, not run-time configurability.

justinbchan:
That was originally done in https://github.com/palantir/http-remoting-api/pull/81, but there was an expressed desire from @uschi2000 and @markelliot to have consistent behaviour in services

j-baker:
Well a counterpoint to that is our service configuration block

j-baker:
Like, I'm very in sync that minimizing our quantity of configuration is good - i'm just cognizant that the way we've built our infra assumes a configuration stored in YAML, and forcing people to construct clients this way sucks because it means that without fun changes internally, you can't easily get e.g. a live reloadable client.

My ideal here would be something like:

1. You have some amount of stuff in config.
2. You have some amount of stuff in code.
3. You have code which takes both and creates a ClientConfiguration.

This is so similar to what we have right now - it's just the idea that you get some extra per-service configuration.

ClientConfiguration is the right place to put this flag, even if it doesn't end up directly mapping to on-disk config.

http-clients/src/main/java/com/palantir/remoting3/clients/RpcMode.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.palantir.remoting3.clients;
+
+public enum RpcMode {
+    /**
+     * Once a node is found that returns responses successfully, continue to use that node until a failure is
+     * received.
+     */
+    FAILOVER_ON_ERROR,
```
uschi2000: nit: new line before

uschi2000: to make the contrast to RR more explicit, how about PIN_UNTIL_ERROR?

http-clients/src/main/java/com/palantir/remoting3/clients/RpcMode.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.remoting3.clients;
+
+public enum RpcMode {
```
uschi2000: I think this is a `NodeSelectionStrategy`

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -50,6 +51,14 @@ private ClientConfigurations() {}
      * empty/absent configuration with the defaults specified as constants in this class.
      */
     public static ClientConfiguration of(ServiceConfiguration config) {
+        return of(config, DEFAULT_RPC_MODE);
+    }
+
+    /**
+     * Creates a new {@link ClientConfiguration} instance from the given {@link ServiceConfiguration} and
```
uschi2000: better: `Like {@link of(ServiceConfiguration)}, but using the explicitly given NodeSelectionStrategy.`

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfiguration.java
```diff
@@ -78,6 +78,12 @@
     /** The maximum number of times a failed request is retried. */
     int maxNumRetries();
 
+    /**
+     * The mode to use while processing requests. By default, requests are pinned to a URL until conditions are met
```
uschi2000: `Indicates how the target node is selected for a given request.` (no need to say something about the default here, since it's always present)

uschi2000 (approved):

uschi2000:
@iamdanfox are you happy with this?

iamdanfox:
Just to make sure I understand the expected usage here- we’re intending the majority of consumers to not touch this and keep using the nice ClientFactory from WC, but if a service knows its requests are particularly expensive, it can opt-in to ROUND_ROBIN to try to be a good citizen and balance load on upstream servers?

What’s the long term goal here? Are we intending to gradually promote round robin as a forcing function for server authors to not rely on sticky sessions?

justinbchan:
Your understanding is correct; without any code changes, the current pin-until-failure strategy will continue to be employed. This allows for services that have identified problems tied to that strategy to perform round robin where appropriate (e.g. where caching is not a concern, where performance degrades due to unbalanced traffic/load).

iamdanfox (approved):
LGTM

PR #720, opened 2018-05-31
Allow host-metrics to be non-static

gatesn:
Kinda ugly, let me know if it would be cleaner to create an OkHttpClientFactory class for which the OkHtpClients class keeps an internal static instance?


ferozco:
I think this looks fine to me, but I defer to @uschi2000

j-baker:
You need to update the FeignClientBuilder and RetrofitClientBuilder to take one of these

PR #721, opened 2018-06-01
Expose host metrics registry

gatesn:
Alternative to #720

### DIFF
--- a/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -26,6 +26,7 @@ import com.palantir.remoting3.jaxrs.feignimpl.Java8OptionalAwareContract;
 import com.palantir.remoting3.jaxrs.feignimpl.PathTemplateHeaderEnrichmentContract;
 import com.palantir.remoting3.jaxrs.feignimpl.PathTemplateHeaderRewriter;
 import com.palantir.remoting3.jaxrs.feignimpl.SlashEncodingContract;
+import com.palantir.remoting3.okhttp.HostMetricsRegistry;
 import com.palantir.remoting3.okhttp.OkHttpClients;
 import feign.CborDelegateDecoder;
 import feign.CborDelegateEncoder;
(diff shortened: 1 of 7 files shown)


jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -80,6 +93,9 @@
     public final <T> T build(Class<T> serviceClass, UserAgent userAgent) {
         ObjectMapper objectMapper = getObjectMapper();
         ObjectMapper cborObjectMapper = getCborObjectMapper();
+        okhttp3.OkHttpClient okHttpClient = Optional.ofNullable(hostMetricsRegistry)
+                .map(hostMetrics -> OkHttpClients.create(config, userAgent, hostMetrics, serviceClass))
+                .orElse(OkHttpClients.create(config, userAgent, serviceClass));
```
iamdanfox: Could we do `orElseGet` here just to avoid any unintended side-effects?

iamdanfox:
I think this looks sensible - it always pains me a bit when we have to add more methods to the top level build & static factories, but I don't really see an easy way around this.

I've retriggered that last build, but it still looks red?

gatesn:
@iamdanfox I fixed the tests, and also deprecated the `OkHttpClients` methods that didn't accept a custom host metrics registry.

retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/Retrofit2Client.java
```diff
@@ -56,7 +71,19 @@ private Retrofit2Client() {}
     public static <T> T create(
             Class<T> serviceClass,
             UserAgent userAgent,
-            Refreshable<ClientConfiguration> config) {
+            Refreshable<ClientConfiguration> config,
```
iamdanfox: Just one more request: could we keep the `Refreshable<ClientConfiguration>` as the last argument?  (It's just to match the pattern from functional pattern where currying is a thing.)

iamdanfox (approved):
Approved pending argument ordering question

gatesn:
Agree, I've re-ordered the arguments

PR #723, opened 2018-06-04
overload withNewTrace to ExecutorService

rubenfiszel:


iamdanfox (approved):

iamdanfox:
@rubenfiszel thanks for the PR, this looks like a good addition (especially as we already do this for ScheduledExecutorService).  Could you add a tiny unit test just to make sure this behaviour doesn't get accidentally broken in a future refactor?  (Also looks like checkstyle needs fixing up)

rubenfiszel:
@iamdanfox fixed checkStyle, added test :)

PR #724, opened 2018-06-04
Add temporal cache to avoid failed URLs

justinbchan:
On request failure, mark URLs as failed, which removes them from the pool of URLs for a given duration. This will be important to round-robin efficiently.


jaxrs-clients/versions.lock
```diff
@@ -69,6 +69,12 @@
                 "com.palantir.remoting3:tracing"
             ]
         },
+        "com.github.ben-manes.caffeine:caffeine": {
```
uschi2000: Think I'd prefer to keep the dependencies as minimal as possible. Guava's cache is probably good enough for this use case.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -144,6 +144,8 @@ public void enqueue(Callback callback) {
         super.enqueue(new Callback() {
             @Override
             public void onFailure(Call call, IOException exception) {
+                urls.markFailed(request().url());
```
uschi2000: what about 503s, do they count as "failed" in this sense?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelector.java
```diff
@@ -51,4 +51,9 @@
      * okhttp3.Request#url} does.
      */
     List<HttpUrl> getBaseUrls();
+
+    /**
+     * Marks the specified URL as failed, which removes it from the pool of available URLs.
```
uschi2000: "Indicates that a call against the given URL has failed. Implementations can use failure statistics to determine which hosts may be unavailable and should be avoided for future calls."

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -32,15 +35,22 @@
...
+                .maximumSize(baseUrls.size())
+                .expireAfterWrite(failedUrlExpiration)
+                .build();
+        this.useFailedUrlCache = failedUrlExpiration.compareTo(Duration.ZERO) > 0;
 
         Preconditions.checkArgument(!baseUrls.isEmpty(), "Must specify at least one URL");
     }
 
-    static UrlSelectorImpl create(Collection<String> baseUrls, boolean randomizeOrder) {
+    static UrlSelectorImpl create(Collection<String> baseUrls, boolean randomizeOrder, Duration failedUrlExpiration) {
```
uschi2000: maybe now just complex enough to warrant a little bit of javadoc?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -103,17 +117,35 @@ private static String switchWsToHttp(String url) {
...
-        int index = currentUrl.updateAndGet(
-                (currentUrlIndex) -> (existingUrlIndex.orElse(currentUrlIndex) + 1) % baseUrls.size());
-        return redirectTo(existingUrl, baseUrls.get(index));
+        int numAttempts = 0;
+        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
+
+        // Skip over any URLs that are marked as failed
+        while (numAttempts < baseUrls.size()) {
+            potentialNextIndex = (potentialNextIndex + 1) % baseUrls.size();
+            Boolean maybeInCache = this.failedUrls.getIfPresent(baseUrls.get(potentialNextIndex));
```
uschi2000: remove `this.`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -103,17 +117,35 @@ private static String switchWsToHttp(String url) {
...
         Optional<Integer> existingUrlIndex = indexFor(existingUrl);
 
-        // ...otherwise we fall back to the stateful current url index
-        int index = currentUrl.updateAndGet(
-                (currentUrlIndex) -> (existingUrlIndex.orElse(currentUrlIndex) + 1) % baseUrls.size());
-        return redirectTo(existingUrl, baseUrls.get(index));
+        int numAttempts = 0;
+        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
+
+        // Skip over any URLs that are marked as failed
```
uschi2000: "Find the next URL that is not marked as failed"

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
uschi2000: maybe rename to `isMarkedAsFailed` ?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -103,17 +117,35 @@ private static String switchWsToHttp(String url) {
...
     }
 
     @Override
     public Optional<HttpUrl> redirectToCurrent(HttpUrl current) {
         return redirectTo(current, baseUrls.get(currentUrl.get()));
     }
 
+    @Override
+    public void markFailed(HttpUrl failedUrl) {
+        if (this.useFailedUrlCache) {
```
uschi2000: 2x remove `this.`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -103,17 +117,35 @@ private static String switchWsToHttp(String url) {
...
+            this.failedUrls.put(failedUrl, Boolean.TRUE);
```
uschi2000: should be able to put `true`, no?

uschi2000 (changes requested):
implementation looks cool, but needs tests.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
justinbchan: As is, yes. Happy to discuss if that's too broad of a failure condition

uschi2000: I think the behavior you chose is actually the right one: markFailed on IOExceptions (indicating that the host is down), but not on 503 (which often indicates a temporary problem). Worth explaining this rationale in a comment though.

j-baker:
Just want to flag that most open source implementations of stuff like this end up with circuit breaking to avoid cascading failures.

It feels a bit odd to be doing connection pool stuff at this level, do we have any way to reuse the code that OkHttp already has with regards to doing this stuff?

j-baker:
like e.g., is there anything we could do with this? https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/connection/ConnectInterceptor.java#L41, which comes from https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RealCall.java#L185

justinbchan:
Something's a little fishy with JaxRsClientFailoverTest::testQosError_performsFailover; seems to non-deterministically fail. Looking into this.

justinbchan:
Probably related to static vars and Theory/Datapoint. Will refactor

http-clients/src/main/java/com/palantir/remoting3/clients/ClientConfigurations.java
```diff
@@ -51,13 +52,22 @@ private ClientConfigurations() {}
...
      */
     public static ClientConfiguration of(ServiceConfiguration config, NodeSelectionStrategy nodeSelectionStrategy) {
+        return of(config, nodeSelectionStrategy, DEFAULT_FAILED_URL_COOLDOWN);
+    }
+
+    /**
+     * Like {@link #of(ServiceConfiguration, NodeSelectionStrategy)}, with a specified {@link Duration} for the
+     * amount of time URLs marked as failed should be avoided in subsequent calls.
+     */
+    public static ClientConfiguration of(ServiceConfiguration config, NodeSelectionStrategy nodeSelectionStrategy,
```
iamdanfox: Rather than adding lots of static factories with more and more overridden parameters, can we just recommend people use:

```groovy
        ClientConfiguration.builder()
                .from(ClientConfigurations.of(config))
                .failedUrlCooldown(failedUrlCooldown)
                .build()
```

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -103,17 +125,36 @@ private static String switchWsToHttp(String url) {
...
-        int index = currentUrl.updateAndGet(
-                (currentUrlIndex) -> (existingUrlIndex.orElse(currentUrlIndex) + 1) % baseUrls.size());
-        return redirectTo(existingUrl, baseUrls.get(index));
+        int numAttempts = 0;
+        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
+
+        // Find the next URL that is not marked as failed
+        while (numAttempts < baseUrls.size()) {
+            potentialNextIndex = (potentialNextIndex + 1) % baseUrls.size();
+            Boolean maybeInCache = failedUrls.getIfPresent(baseUrls.get(potentialNextIndex));
```
iamdanfox: Could you store a single variant enum `FAILED` instead of a boolean? It's kinda confusing that we're never reading the value of that boolean and just checking it for null!

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -18,29 +18,47 @@
...
     }
 
-    static UrlSelectorImpl create(Collection<String> baseUrls, boolean randomizeOrder) {
+    /**
+     * Creates a new {@link UrlSelector} with the supplied URLs. The order of the URLs may be randomized by setting
+     * {@code randomizeOrder} to true. If a non-zero {@code duration} and {@code timeUnit} are specified, URLs that are
+     * marked as failed using {@link #markAsFailed(HttpUrl)} will be removed from the pool of prioritized, healthy URLs
+     * for that period of time.
+     */
+    static UrlSelectorImpl create(Collection<String> baseUrls, boolean randomizeOrder, long duration,
```
iamdanfox: We now have two methods called `create` which do different things.  Can we go for descriptive names?  e.g. 

```
create()
createWithFailedUrlCooloff()
```

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/UrlSelectorImpl.java
```diff
@@ -18,29 +18,47 @@
...
 
 final class UrlSelectorImpl implements UrlSelector {
 
     private final ImmutableList<HttpUrl> baseUrls;
     private final AtomicInteger currentUrl;
+    private final Cache<HttpUrl, Boolean> failedUrls;
+    private final boolean useFailedUrlCache;
 
-    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls) {
+    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, long duration, TimeUnit unit) {
```
iamdanfox: Rather than using duration == 0 as a magic number, could we just pass in Optional<Duration> here?

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientFailoverTest.java
```diff
@@ -24,54 +24,51 @@
...
-import org.junit.Before;
-import org.junit.Rule;
 import org.junit.Test;
+import org.junit.experimental.theories.DataPoint;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
 
-
+@RunWith(Theories.class)
```
iamdanfox, line 42: Woah I'd never seen this before! Does it have any particular advantages over `@Parameterized`?

justinbchan: I haven't used `@RunWith(Parameterized.class)` but it seems like `Theories.class` allows for a mix of DataPoint/Theory-style tests and normal `@Test`s (which is beneficial here).

uschi2000 (approved):

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -124,7 +124,8 @@ private static RemotingOkHttpClient createInternal(
         TaggedMetricRegistry registry = DefaultTaggedMetricRegistry.getDefault();
 
         // Routing
-        UrlSelectorImpl urlSelector = UrlSelectorImpl.create(config.uris(), randomizeUrlOrder);
+        UrlSelectorImpl urlSelector = UrlSelectorImpl.createWithFailedUrlCooldown(config.uris(), randomizeUrlOrder,
+                config.failedUrlCooldown());
```
iamdanfox, line 128: I tried breaking this by replacing `config.failedUrlCooldown()` with `Duration.ZERO` and tests still seem to pass - can we try to add a test that would catch a bad refactor like this?

justinbchan: `UrlSelectorImpl.create` calls `UrlSelectorImpl.createWithFailedUrlCooldown` with `Duration.ZERO`, which is equivalent to not caching.

justinbchan: I'm also hesitant to have `createWithFailedUrlCooldown` take an Optional Duration since it would be odd to invoke a function that explicitly says it should be provided

iamdanfox (approved):
Looks good to me! Seems safe because this shouldn't affect the majority of users (unless they opt-in to ROUND_ROBIN)

PR #726, opened 2018-06-07
switch to java-library-oss circle2 template

qinfchen:
similar to the setup in http-remoting-api


gradle/publish-jar.gradle
```diff
@@ -0,0 +1,24 @@
+apply plugin: 'com.jfrog.bintray'
```
qinfchen, line 1: @iamdanfox, maybe worth changing this `publish-jar.gradle` to `publish-jar-base.gradle` if people want to customize it?

iamdanfox: Nah I'd prefer to try to find a common thing that's good enough for all our OSS repos so we don't have some of them customizing and some of them not customizing!

qinfchen: in that case, what do we do about these? https://github.com/palantir/http-remoting/blob/31060a2a3a18d82b093fe804fb394cad95cc5e6d/gradle/publish.gradle

.circleci/config.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  build:
+    docker:
+      - image: circleci/openjdk:8u171-jdk-node-browsers
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.console=plain -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false
+
+    steps:
+      - checkout
+      - restore_cache:
```
uschi2000, line 12: is there way not copy&paste these 20ish lines for restore&save cache everywhere?

qinfchen: Not that i know of from the [doc](https://circleci.com/docs/2.0/caching/),  @jmcampanini?

jmcampanini: you would need to use yaml anchors, etc. it's pretty useful since you only name the caches in 1 place, but using anchors makes the yaml harder to read in general. tradeoffs until circle implements "orbs" (re-usable steps that are defined by us)

qinfchen: anchors are only useful if they are being used multiple times, the save cache or restore cache is used once here.

qinfchen:
@iamdanfox, @uschi2000 ping?

gradle/publish.gradle
```diff
@@ -1,9 +1,10 @@
...
-    publications {
-        nebula(MavenPublication) {
-            from components.java
-        }
+jar {
+    manifest {
+        attributes(
+                "Implementation-Title"   : project.name,
+                "Implementation-Version" : project.version,
+                "Implementation-Vendor"  : "Palantir Technologies Inc.")
```
iamdanfox: I don't think Implementation-Title or Vendor are used for anything.  Can we delete and stick to just the publish-jar.gradle file?

circle.yml
```diff
@@ -1,38 +0,0 @@
...
-          1)
-            ./gradlew testBouncyCastle testSun --parallel
-            ;;
-          esac
-      :
-        parallel: true
-        timeout: 1200
-  post:
-    - mkdir -p $CIRCLE_TEST_REPORTS/junit/
-    - find . -type f -regex ".*/build/test-results/.*xml" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
```
iamdanfox, line 32: I think there's a chunk of build.gradle that mentions `System.env.CIRCLE_TEST_REPORTS` which can be deleted (as this variable no longer exists in Circle2)

iamdanfox (approved):
Minor comments, but looks like a big improvement overall :)

iamdanfox (changes requested):
Hang on it's reporting 905 tests passed before and 430 tests after

https://circleci.com/gh/palantir/http-remoting/2106
https://circleci.com/gh/palantir/http-remoting/2112

Is the 7m59 -> 3m24 speedup too good to be true?

qinfchen:
The reason that the tests are less than before is that the test files get overwritten, which was not a problem with  `CIRCLE_TEST_REPORTS`
```
./keystores/build/test-results/test-sun/TEST-com.palantir.remoting3.config.ssl.SslConfigurationTest.xml
./keystores/build/test-results/test/TEST-com.palantir.remoting3.config.ssl.SslConfigurationTest.xml
./keystores/build/test-results/test-bouncy-castle/TEST-com.palantir.remoting3.config.ssl.SslConfigurationTest.xml
```
I am still trying to figure out why the number is different. https://circleci.com/gh/palantir/http-remoting/2125

qinfchen:
@iamdanfox, turns out circle1 also collects gradle xmls which accounts for about 360 tests
>pkcs1-reader-sun, pkcs1-reader-bouncy-castle, http2-agent, refresh-utils, keystores-bouncy-castle, tracing-jaxrs, jaxrs-clients, keystores-sun, tracing-okhttp3, **gradle**, tracing, jersey-servers, retrofit2-clients, error-handling, keystores, jackson-support, http-clients, okhttp-clients

iamdanfox (approved):

PR #727, opened 2018-06-08
AssertionErrors are converted into service exceptions with type internal

j-baker:
At the moment, AssertionErrors are not handled by remoting, even though
they're used fairly widely within services and libraries we have
internally. This means that the body ends up being empty if they
ever get hit, which leads to a fairly cryptic exception being
thrown on the client.

Bugs are still debuggable, it's just harder. This PR makes the claim
that AssertionErrors should be handled specially.

Could also argue that all errors should be handled the same as e.g.
a random RuntimeException (specifically thinking NoSuchMethodError).


jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/AssertionErrorExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.remoting3.servers.jersey;
+
+import com.palantir.remoting.api.errors.ErrorType;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+final class AssertionErrorExceptionMapper extends JsonExceptionMapper<AssertionError> {
```
iamdanfox: It does seem pretty obscure that people are throwing `AssertionErrors` - I've generally only seen them in private constructors (`throw new AssertionError("Not instantiable");`).  However, I think it does makes sense that in the strange case that this happens, we might as well give clients something helpful.

Can we expand it to all Errors? This would get us `NoSuchFieldError`, `NoSuchMethodError` and `NoClassDefFoundError`?

schlosna: Expanding to `Error` or `Throwable` would also cover the occasional `OutOfMemoryError`

j-baker: yeah that was my reticence at the start. I guess there's the thing where we hope that OOMs get handled by the JVM itself?

j-baker:
I have rewritten to be in terms of the above - aka all Throwables - since sls-packaging handles throwing on out-of-memory.

iamdanfox (approved):

iamdanfox:
This should be fine because we have `-XX:+CrashOnOutOfMemoryError` in sls-packaging. 

https://github.com/palantir/sls-packaging/blob/2.18.0/gradle-sls-packaging/src/main/groovy/com/palantir/gradle/dist/service/gc/GcProfile.java#L28

PR #729, opened 2018-06-11
Mark URL as failed on QoS unavailable

justinbchan:
If a node responds with a QoS unavailable exception, mark it as failed.
* In the no-cache case (the default), this is a no-op
* In the cache + pin until failover case, this is also a no-op. Another node will be sought (since `urls.redirectToNext` is invoked), then pinned to that (if a successful response is received)
* In the soon-to-be-implemented round robin case, this will help avoid unavailable nodes allowing for recovery


j-baker:
so I don't think this meshes well with @uschi2000's load shedding work

j-baker:
getting a 503 shouldn't be an indication of downtime

uschi2000:
yeah, I agree with that assessment.

justinbchan:
I think that this is more a function of nomenclature than actual behaviour; marking as "failed" is more like "avoid this URL" than "don't hit this URL". In the case of a 503 Unavailable, I think that it's reasonable for subsequent calls to avoid that URL.

uschi2000:
The thing I don't like about this is that there are now two different backoff/timeout mechanisms.

uschi2000:
And absent data proving otherwise, I'd prefer the simpler system.

justinbchan:
The tangible side-effect is that one of the tests which I think we'd expect to pass (testQosError_performsFailover) will fail non-deterministically in the round robin case:
1. Suppose the initial request gets routed to server 1. It will receive the 503 and be routed to the next server.
2. Without a cooldown period, a round robin interceptor may route it back to server 1, which responds with `foo`. If there was a cooldown period, it will correctly route to server 2 since the retry will avoid server 1.

If we're happy to accept some funkiness with QoS and round robin, I can close this PR and adjust the upcoming round robin interceptor one accordingly.

uschi2000:
yep, let's do that.

justinbchan:
Closing this. RoundRobinInterceptor tests will not include QoS failure cases per https://github.com/palantir/http-remoting/pull/737

PR #734, opened 2018-06-13
bug catching logger

j-baker:


okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java
```diff
@@ -43,16 +46,27 @@
...
+                            uncaughtException))
+            .setNameFormat("remoting-okhttp-dispatcher-%d")
+            .build();
     /**
      * The {@link ExecutorService} used for the {@link Dispatcher}s of all OkHttp clients created through this class.
+     * Same as OkHttp's default, but with a logging uncaught exception handler.
      */
-    private static final ExecutorService executionExecutor = Tracers.wrap(new Dispatcher().executorService());
+    private static final ExecutorService executionExecutor =
+            Tracers.wrap(Executors.newCachedThreadPool(executionThreads));
```
uschi2000, line 69: should we use the same type of executor as the okhttp Dispatcher?

j-baker: it is!

j-baker: noted in the Javadoc

uschi2000: ah, just had to chase the `Executors` call. thanks.

uschi2000 (approved):

PR #738, opened 2018-06-20
[do not merge] Use unknown remote exception

### LINKED ISSUE #628: RemotingOkHttpClient should differentiate different IOExceptions

gatesn:
Fixes #628. There is an http-remoting-api change to go along with this but wanted to keep discussion in one place. 

Previously, for non-remoting servers http-remoting would return a `SafeIoException` with a `SafeArg{code=<http status code}`, and an `UnsafeArg{body=<error body>}`.

This change introduced an `UnknownRemoteException` so clients can retrieve the status code and response body in a typed way.

This _could_ be considered an API break given a remote 404 error used to throw a checked IOException, but would now throw an unchecked `UnknownRemoteException`. Only _actual_ IOExceptions (e.g. connection timeout, connection refused) will still be propagated as IOExceptions. What do people think?

Side note, how do I generate a versions.lock?


uschi2000:
`gd generateLock saveLock`

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -124,6 +126,14 @@ public void onResponse(Call call, Response response) throws IOException {
...
                 correctStackTrace.initCause(e);
                 throw correctStackTrace;
+            } else if (e.getCause() instanceof IoUnknownRemoteException) {
+                UnknownRemoteException wrappedException = ((IoUnknownRemoteException) e.getCause())
+                        .getWrappedException();
+                UnknownRemoteException correctStackTrace = new UnknownRemoteException(
+                        wrappedException.getStatus(),
+                        wrappedException.getBody());
+                correctStackTrace.initCause(e);
+                throw correctStackTrace;
```
uschi2000, line 130: this throws a non-IOException. I'm not sure that's kosher given the OkHttp/Feign contract?

gatesn: But a RemoteException isn't an IOException either?

uschi2000: oh, I misread, thanks

uschi2000:
i'm fine with this

okhttp-clients/src/test/java/com/palantir/remoting3/okhttp/OkHttpClientsTest.java
```diff
@@ -216,12 +216,12 @@ public void throwsIoExceptionWithCorrectBodyAfterFailingToDeserializeSerializabl
 
         OkHttpClient client = createRetryingClient(1);
         Call call = client.newCall(new Request.Builder().url(url).build());
-        assertThatThrownBy(call::execute).isInstanceOf(SafeIoException.class)
+        assertThatThrownBy(call::execute).isInstanceOf(UnknownRemoteException.class)
```
iamdanfox: Could we make one of these assertions also check the `getStatus()` method?

versions.props
```diff
@@ -6,7 +6,7 @@ com.google.errorprone:error_prone_core = 2.0.18
 com.google.guava:guava = 20.0
 com.jcraft:jzlib = 1.1.3
 com.netflix.feign:feign-*= 8.17.0
-com.palantir.remoting-api:* = 1.7.0
+com.palantir.remoting-api:* = 1.8.0-15-gb1c30f7
```
iamdanfox: I've just chopped a release: https://github.com/palantir/http-remoting-api/releases/tag/1.9.0

Wanna do the updateLocks saveLocks thing again to try to bring back that `okhttp-clients/versions.lock` file?

gatesn:
@iamdanfox addresses all comments and regenerated lock files.

gatesn:
Actually, can we hold off on this until we verify whether it's a break?

PR #739, opened 2018-06-20
Don't use a sync timeout

### LINKED ISSUE #740: Retrying on timeout does not work properly when connect, read and write timeout are set
## What happened?

Currently timeouts do not generally get retried

Take the following case: connect timeout 10s, read timeout 1min, write timeout 1min

If a request hits the 1min read timeout, it won't have an opportunity to retry because it will hit
https://github.com/palantir/http-remoting/blob/develop/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java#L112

Which is set as the largest of connect, read and write timeout
https://github.com/palantir/http-remoting/blob/develop/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/OkHttpClients.java#L207

Requests should be able to retry on timeouts

dansanduleac:
In light of using a backoff strategy, and the fact that requests can be queued in the okhttp dispatcher, it doesn't make sense to enforce a sync timeout of `max(read,write,connect timeouts)` on the future we wait on in `RemotingOkHttpCall.execute()`.
Attempts to make this timeout duration large enough to accommodate for the backoff strategy would not be completely correct either because we'd have to account for queueing too.
Therefore, it's better just to drop this blanket timeout, and trust okhttp's timeout logic.

I'm planning to follow this up by
- [x] adding metrics on the dispatcher's `queuedCallsCount()` to give more visibility into how often requests end up getting queued #742
- [ ] (if we think it necessary based on the metrics) enforcing a limit on the queued calls

Fixes #740

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
@@ -101,6 +101,9 @@ final class RemotingOkHttpCall extends ForwardingCall {
         });
 
         try {
+            // We don't enforce a timeout here because it's not possible to know how long this operation might take.
+            // First, it might get queued indefinitely in the Dispatcher, and then it might get retried a (potentially)
+            // unknown amount of times by the BackoffStrategy.
             return future.get();
         } catch (InterruptedException e) {
             getDelegate().cancel();
--- a/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
@@ -29,7 +29,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


samsinensky:
https://github.com/palantir/http-remoting/issues/740

uschi2000 (approved):
LGTM. Can you leave a comment on the `.get()`, indicating why this is the correct thing to do?

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/RemotingOkHttpCall.java
```diff
@@ -101,6 +101,9 @@ public void onResponse(Call call, Response response) throws IOException {
         });
 
         try {
+            // We don't enforce a timeout here because it's not possible to know how long this operation might take.
+            // First, it might get queued indefinitely in the Dispatcher, and then it might get retried a (potentially)
+            // unknown amount of times by the BackoffStrategy.
```
uschi2000: plus: This call times out when the underlying OkHttp call times out, possibly after a number of retries.

right?

dansanduleac: Would you rather I said only that? There is some overlap with my comment. Feel free to modify it as you wish & merge.

uschi2000 (approved):

uschi2000:
cheers, @dansanduleac

PR #747, opened 2018-06-27
Upgrade okhttp to 3.10.0

iamdanfox:
https://github.com/square/okhttp/blob/master/CHANGELOG.md

cc @dansanduleac


uschi2000:
What does this mean for us?
```
- New: Honor the Retry-After response header. HTTP 503 (Unavailable) responses are retried automatically if this header is present and its delay is 0 seconds. HTTP 408 (Client Timeout) responses are retried automatically if the header is absent or its delay is 0 seconds.
```

iamdanfox:
Not sure yet - @dansanduleac is most interested in upgrading to okhttp 3.10.0 because it supports, conscrypt so just wanted to see whether http-remoting was going to work!!

EDIT apparently the internal CRP product is using 3.10.0 and it seems to be working for them (??)

carterkozak:
Note that 3.10.0 removed support for hostname verification via CN, and only validates SAN values.
This is ideal and matches modern versions of chrome, though if we're worried about breaking things we can supply our own HostnameVerifier instance to okhttp.

iamdanfox:
Closing in favour of https://github.com/palantir/http-remoting/pull/794

PR #750, opened 2018-06-29
Prefer chacha ciphers when available

carterkozak:
More secure and performant than the other ciphers listed.

### DIFF
--- a/http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
+++ b/http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
@@ -9,6 +9,8 @@ import com.google.common.collect.ImmutableList;
 public final class CipherSuites {
 
     private static final ImmutableList<String> FAST_CIPHER_SUITES = ImmutableList.of(
+            "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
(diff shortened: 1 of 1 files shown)


http-clients/src/main/java/com/palantir/remoting3/clients/CipherSuites.java
```diff
@@ -9,6 +9,8 @@
 public final class CipherSuites {
 
     private static final ImmutableList<String> FAST_CIPHER_SUITES = ImmutableList.of(
+            "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
```
j-baker, line 12: just want to flag that this list is currently broken, since Jetty refuses to use anything other than GCM with http/2, so will just run with HTTP/1. If you're using undertow this may be the right choice

carterkozak: This does not relate in any way to http/2, just tls ciphers.

Is this actually broken, or are you referring to this?
https://http2.github.io/http2-spec/#rfc.section.9.2.2

j-baker: It relates to how we configure our web servers internally. So the presence of this list (without GCM) breaks us internally because it means that Jetty will refuse to use HTTP/2.

PR #751, opened 2018-06-29
Enable GCM ciphers by default

carterkozak:
They are more secure than the others aside from
CHACHA20_POLY1305, and provided by the java 8 jvm.

Allowing these ciphers by default provides betters security and
compatibility with other services. Servers may choose to prefer
other ciphers if available, and clients using native tls
implementations (e.g. Conscrypt) will perform much better
using gcm ciphers than cbc.


j-baker:
the reason these were disabled were because they were _considerably_ less performant than the CBC ciphers unless you are using something like Conscrypt. I'd therefore not change this flag without considerable internal thought (like enabling conscrypt everywhere).

j-baker:
but e.g. we have boxes with 4 cores internally which are frequently writing out at 180MB/sec, and our testing implies that if we turn on GCM this simply won't work.

carterkozak:
I'm not suggesting that all clients must use gcm, gcm ciphers are late in the list, it will only be used if there are no other options available, or if a server prefers gcm over other ciphers, which we configure.

j-baker:
If we add these two lines, the vast majority of our internal servers will pick GCM and no others.

markelliot:
the ciphers are already allowed/one can flip the default in configuration.

PR #757, opened 2018-07-11
disable ALLOW_COERCION_OF_SCALARS

qinfchen:
this is to prevent object mappers from mapping integer values to booleans or string to integer values and vise versa..

### DIFF
--- a/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
+++ b/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
@@ -17,6 +17,7 @@
 package com.palantir.remoting3.ext.jackson;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
@@ -116,6 +117,7 @@ public final class ObjectMappers {
                 .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE)
                 .disable(DeserializationFeature.WRAP_EXCEPTIONS)
                 .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
+                .disable(MapperFeature.ALLOW_COERCION_OF_SCALARS)
                 .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES);
     }
 }
--- a/extras/jackson-support/src/test/java/com/palantir/remoting3/ext/jackson/ObjectMappersTest.java
+++ b/extras/jackson-support/src/test/java/com/palantir/remoting3/ext/jackson/ObjectMappersTest.java
@@ -17,9 +17,11 @@
 package com.palantir.remoting3.ext.jackson;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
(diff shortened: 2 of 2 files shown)


markelliot:
I think we want this liberalness?

qinfchen:
So deserializing conjure Boolean Object, jackson is casting `1` or any value as `true`, and `0` to `false` which fails conjure de/ser tests.

markelliot:
I think that implies a problem with the tests; the leniency here is positive, IMO

iamdanfox:
@markelliot this was exposed by our exhaustive Conjure wire format tests.  I don't think we should normalize the practise of relying on coercion like this, as it seems prone to break subtly between jackson versions...  Given that we've got a well defined wire.md and conjure can generate compliant servers and clients, what remaining use cases are there for this leniency?

markelliot:
I think there's a pretty big difference between _relying on_ and _accepting_ -- I'm voting that we continue to make our servers lenient in what they accept, even if the overwhelming majority of our clients will produce pedantically correct input.

iamdanfox:
So are we OK with clients doing some coercing when they receive imprecise server responses?  This seems pretty hard to specify/implement in a cross-language way and would make backend implementations less fungible

uschi2000:
What's your reasoning here, Mark? Do you have concerns with existing misbehaved clients?

uschi2000:
In a vacuum, I'd prefer the stricter behavior, both in clients and in servers.

markelliot:
I have only weak justification, mostly I'm concerned we don't know where this may be misused now (these ObjectMappers occasionally make their way into non-RPC usage) and that the lenience of this behavior is unlikely to be an error.

It seems y'all feel more strongly than I do at this point, so we can proceed.

iamdanfox:
OK, I think we should proceed and make http-remoting strictly compliant - if it turns out that some product is using `ObjectMappers.newServerObjectMapper()` for some non-RPC thing, they can always just make a code-change to `.enable(MapperFeature.ALLOW_COERCION_OF_SCALARS)`.

extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
```diff
@@ -116,6 +117,7 @@ public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
                 .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE)
                 .disable(DeserializationFeature.WRAP_EXCEPTIONS)
                 .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS)
+                .disable(MapperFeature.ALLOW_COERCION_OF_SCALARS)
```
iamdanfox, line 120: One thing to note is that they've moved this from `MapperFeature` -> `DeserializationFeature` to be released in 3.X, so this will be a guaranteed break when we eventually adopt jackson 3.X.

https://github.com/FasterXML/jackson-databind/blob/481592ca8386bb4e9ce79bfc0596bb7cf85ee780/src/main/java/com/fasterxml/jackson/databind/DeserializationFeature.java#L334

iamdanfox:
@qinfchen could we add one little unit test to ensure this doesn't accidentally regress?

extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/ObjectMappers.java
qinfchen: How so? This would only be a dev break in http when picking up 3.X, no? Since it is a major version for jackson, there will be other breaking changes regardless.

j-baker: given that DeserializationFeature does exist, might as well use? Minimizing our breaks always good

j-baker: ideally you have something that works with both, because right now it's the only breaking change I can see that would affect remoting

j-baker: https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0

qinfchen: DeserializationFeature doesn't have this feature until 3.0

j-baker: ah, ok

j-baker:
note: remoting object mappers are frequently used for DB persistence

iamdanfox:
The latest plan is to not make this change to the existing http-remoting ObjectMapper (to ensure we don't unintentionally break anyone's DB persistance) and instead incorporate this new `.disable(MapperFeature.ALLOW_COERCION_OF_SCALARS)` change into the proposed renaming PR (http-remoting -> conjure-java-client).

This should ensure that users can continue to use `com.palantir.remoting3.ext.jackson.ObjectMappers` if they want to, while new users will be recommended to take `com.palantir.conjure.java.serialization.ObjectMappers` (with the improved behaviour).

dansanduleac:
Subsumed into #764

PR #759, opened 2018-07-16
Log exceptions when retrying

dvogelbacher:
We have a debug log that indicates when we do retries but we don't log the reason why.
We should also log the exception that was obtained.


palantirtech:
Thanks for your interest in [palantir/http-remoting](https://github.com/palantir/http-remoting), @dvogelbacher! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

iamdanfox (approved):
This seems like a nice improvement to usability :)

PR #761, opened 2018-07-16
Bans NaN, Infinity, +Infinity for double type serde

qinfchen:
## Before this PR
remoting objectmapper allows `NaN` and `Infinity` as values for double type during serde.

## After this PR
`NaN` and `Infinity` are now banned and object  mapper will throw an exception when serialzing/deserialing doubles that have those values. This is to be compliant with the [conjure spec(to be updated)](https://github.com/palantir/conjure/blob/develop/docs/specification.md)

### DIFF
--- a/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/StrictDoubleModule.java
+++ b/extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/StrictDoubleModule.java
@@ -44,9 +44,9 @@ import java.lang.reflect.Type;
  */
 public class StrictDoubleModule extends SimpleModule {
 
-   public StrictDoubleModule() {
+    public StrictDoubleModule() {
         super(StrictDoubleModule.class.getCanonicalName());
-        
+
         addDeserializer(Double.class, new StrictDoubleDeserializer(Double.class, null));
         addDeserializer(double.class, new StrictDoubleDeserializer(double.class, 0.d));
         addSerializer(Double.class, new StrictDoubleSerializer(Double.class));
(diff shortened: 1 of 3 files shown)


extras/jackson-support/src/main/java/com/palantir/remoting3/ext/jackson/StrictDoubleModule.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Used to enforce strict double values by banning "NaN", "+Infinity", and "-Infinity" for both serialization
+ * and deserialization.
+ * <p>
+ * Jackson does not support serialization/deserialization of strict numbers as of 2.9.6.
+ *
+ * @see <a href="https://github.com/FasterXML/jackson-databind/issues/911">https://github.com/FasterXML/jackson-databind/issues/911</a> for more details.
+ */
+public class StrictDoubleModule extends SimpleModule {
```
iamdanfox: If jackson does end up implementing something natively to ban NaN, +/-Infinity, then obviously it would be great to replace our hand-rolled `StrictDoubleModule` with their native solution: https://github.com/FasterXML/jackson-databind/issues/911

shuyangli:
What's the issue that this change is attempting to address? If the goal is to comply with IEEE 754, these values are defined in the spec.

In a lot of situations (especially in the timeseries world), `NaN` and `Infinity` are valid values that have semantic meanings; what values would you propose that we use instead of `NaN` or `Infinity`?

iamdanfox (approved):
This looks super nice! Very much looking forward to not ever having to worry about NaNs or Infinity again :)

j-baker:
We should not be merging this, since it is highly likely that this will break a large number of internal clients.

This should only happen in a major version bump of remoting.

iamdanfox (changes requested):
Good point - gonna table this for now to avoid breaking non-Conjure users who are relying on serializing NaNs and Infinitys

PR #766, opened 2018-07-24
Reject null body in feign clients

iamdanfox:
relevant to https://github.com/palantir/http-remoting/issues/758

### DIFF
--- a/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -37,6 +37,7 @@ import feign.InputStreamDelegateDecoder;
 import feign.InputStreamDelegateEncoder;
 import feign.Java8OptionalAwareDecoder;
 import feign.Logger;
+import feign.RejectNullDecoder;
 import feign.Request;
 import feign.Retryer;
 import feign.TextDelegateDecoder;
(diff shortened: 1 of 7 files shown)


jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -135,6 +136,7 @@ private static Decoder createDecoder(ObjectMapper objectMapper, ObjectMapper cbo
                                 new TextDelegateDecoder(
                                         new CborDelegateDecoder(
                                                 cborObjectMapper,
-                                                new JacksonDecoder(objectMapper))))));
+                                                new RejectNullDecoder(
+                                                        new JacksonDecoder(objectMapper)))))));
```
dansanduleac, line 140: moar nesting!!

dansanduleac (approved):

qinfchen:
i think this violates the wire spec
> Empty or null optional, map, set and list deserialize without error.

markelliot:
what if the main request body is `optional<string>`, for instance?

iamdanfox:
@markelliot the 204 -> optional conversion behaviour all still works, proved by this test: https://github.com/palantir/http-remoting/blob/fc89c8b85182de73ee441faee907052762e1f034/jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/feignimpl/Java8OptionalAwareDecoderTest.java#L56

If we also want a HTTP 200 containing the body `null` to deserialize to Optional.empty(), then I need to add that assertion to our master [test-cases.yml](https://github.com/palantir/conjure-verification/blob/develop/test-cases.yml)

PR #772, opened 2018-07-26
Feign coerces null / absent response to empty collection

dansanduleac:
Fixes the break introduced in #766 - namely, that servers returning `null` where a collection was expected suddenly started throwing.

However, whereas before a `null` response would convert into a java null, with this change we
* still blow up on nulls if not expecting a collection
* create an empty collection otherwise

We support deserializing an empty collection out of

| Response code | Body      |
|---------------|-----------|
| 200           | \<empty\> |
| 200           | `null`    |
| 204           | \<empty\> |

### DIFF
--- a/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -30,6 +30,7 @@ import com.palantir.remoting3.okhttp.HostMetricsRegistry;
 import com.palantir.remoting3.okhttp.OkHttpClients;
 import feign.CborDelegateDecoder;
 import feign.CborDelegateEncoder;
+import feign.CoerceNullCollectionsDecoder;
 import feign.Contract;
 import feign.Feign;
 import feign.GuavaOptionalAwareDecoder;
@@ -131,12 +132,13 @@ abstract class AbstractFeignJaxRsClientBuilder {
 
     private static Decoder createDecoder(ObjectMapper objectMapper, ObjectMapper cborObjectMapper) {
         return new RejectNullDecoder(
-                new Java8OptionalAwareDecoder(
-                    new GuavaOptionalAwareDecoder(
-                            new InputStreamDelegateDecoder(
-                                    new TextDelegateDecoder(
-                                            new CborDelegateDecoder(
-                                                    cborObjectMapper,
-                                                        new JacksonDecoder(objectMapper)))))));
-    }
+                new CoerceNullCollectionsDecoder(
+                    new Java8OptionalAwareDecoder(
+                        new GuavaOptionalAwareDecoder(
+                                new InputStreamDelegateDecoder(
+                                        new TextDelegateDecoder(
+                                                new CborDelegateDecoder(
+                                                        cborObjectMapper,
+                                                            new JacksonDecoder(objectMapper))))))));
+        }
 }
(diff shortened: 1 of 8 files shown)


jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientCollectionHandlingTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public final MockWebServer server = new MockWebServer();
+
+    private Service proxy;
+
+    @Parameters(name = "{index}: code {0} body: \"{1}\"")
+    public static Collection<Object[]> responses() {
+        return Arrays.asList(new Object[][]{
+                { 200, "null" },
+                { 200, "" },
+                { 204, "" }
```
iamdanfox, line 51: Could we add the `{ 204, "null" }` case just to see what happens?

jaxrs-clients/src/main/java/feign/CoerceNullCollectionsDecoder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import feign.codec.Decoder;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public final class CoerceNullCollectionsDecoder implements Decoder {
```
iamdanfox: I think we should javadoc to explain why this exists... ideally we'd link to the spec but it's still not merged yet :/

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -130,13 +131,14 @@ private Contract createContract() {
...
     private static Decoder createDecoder(ObjectMapper objectMapper, ObjectMapper cborObjectMapper) {
-        return new Java8OptionalAwareDecoder(
-                new GuavaOptionalAwareDecoder(
-                        new InputStreamDelegateDecoder(
-                                new TextDelegateDecoder(
-                                        new CborDelegateDecoder(
-                                                cborObjectMapper,
-                                                new RejectNullDecoder(
-                                                        new JacksonDecoder(objectMapper)))))));
+        return new RejectNullDecoder(
```
iamdanfox: Thoughts on renaming this to "NeverReturnNullDecoder" ?

iamdanfox:
We're gonna follow-up with a retrofit one too right?

jaxrs-clients/src/test/java/com/palantir/remoting3/jaxrs/JaxRsClientCollectionHandlingTest.java
dansanduleac:
```
13:23:18.900 [OkHttp http://localhost:54629/...] DEBUG com.palantir.remoting3.okhttp.RemotingOkHttpCall - Rescheduling call after backoff
java.net.ProtocolException: HTTP 204 had non-zero Content-Length: 4
```
and then blocks indefinitely

jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/AbstractFeignJaxRsClientBuilder.java
dansanduleac: sure, don't feel strongly either way

iamdanfox (approved):

PR #774, opened 2018-07-26
Retrofit2 coerce null / absent response to empty collection

dansanduleac:
Fixes the break introduced in #767 - namely, that servers returning `null` where a collection was expected suddenly started throwing.

However, whereas before a `null` response would convert into a java null, with this change we
* still blow up on nulls if not expecting a collection
* create an empty collection otherwise

We support deserializing an empty collection out of

| Response code | Body      |
|---------------|-----------|
| 200           | \<empty\> |
| 200           | `null`    |
| 204           | \<empty\> |

Sadly this is a bit more gross than #772 because retrofit2 doesn't run Converters if the response code was 204.
Instead, for that case, we had to add a `CallAdapter.Factory` that creates special Call instances which sneak in the default value if 
* the expected type was `SomeParametrizedType<List | Set | Map>` where currently the parametrized type can be only `Call<T>` or `CompletableFuture<T>`
* the delegate call returned `null`

### DIFF
--- a/retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/CoerceNullCollectionsCallAdapterFactory.java
+++ b/retrofit2-clients/src/main/java/com/palantir/remoting3/retrofit2/CoerceNullCollectionsCallAdapterFactory.java
@@ -4,6 +4,9 @@
 
 package com.palantir.remoting3.retrofit2;
 
+import com.palantir.logsafe.Preconditions;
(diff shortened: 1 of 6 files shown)


iamdanfox (approved):
Tests look great.  It's a shame retrofit makes it so tricky to change this behaviour, but everything is private so we'll be free to refactor without fear of breaking people.

Overall, I think the outcome of users never worrying about nulls coming from a remoting client is worth the internal complexity!

PR #779, opened 2018-07-30
Builders take a HostEventsSink (that HostMetricsRegistry impls)

dansanduleac:
### Before this PR

Client builders take a [`HostMetricsRegistry`](https://github.com/palantir/http-remoting/blob/6693b375aab63fe34a7ac1f824c4e25c28636292/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java), which accepts marking host response / exception events as well as getting per-host metrics computed from those events.

### After this PR

Client builders take a `HostEventsSink`, which is just an interface that allows one to mark events. This interface doesn't need to concern itself with metrics at all, it just gets told when certain events have occurred.

This would allow us to pass a remoting3 HostMetricsRegistry into conjure client builders, by allowing us to manually implement `HostEventsSink` for the remoting3 HostMetricsRegistry, so that we can share a metrics registry between conjure and remoting3 clients.

However, it requires that remoting3 HostMetricsRegistry also makes its `record`/`recordIoException` methods public, so that we can delegate to them (https://github.com/palantir/http-remoting/pull/780)

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
@@ -5,9 +5,9 @@
 package com.palantir.conjure.java.okhttp;
 
 /**
- * A destination for recording host metrics.
+ * A destination for recording host events.
  */
-public interface HostMetricsSink {
+public interface HostEventsSink {
     void record(String serviceName, String hostname, int port, int statusCode, long micros);
 
     void recordIoException(String serviceName, String hostname, int port);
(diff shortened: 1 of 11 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ */
+
+package com.palantir.conjure.java.okhttp;
+
+/**
+ * A destination for recording host events.
```
iamdanfox: Could we add an explanation for why this exists? I know we didn't have anything previously either, but I think this is an opportunity to be more readable than before!

iamdanfox (approved):
LGTM.  cc @gatesn who implemented this originally

qinfchen (approved):

PR #782, opened 2018-07-31
Re-add WebApplicationExceptionMapper

dansanduleac:
Without this, we won't special-handle exceptions coming from jersey (e.g. routing failure, where jersey throws `NotFoundException`), so they instead get mapped by the less specific `RuntimeExceptionMapper` and incorrectly turned into internal exceptions.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
@@ -35,6 +35,7 @@ public enum ConjureJerseyFeature implements Feature {
         context.register(new IllegalArgumentExceptionMapper());
         context.register(new NoContentExceptionMapper());
         context.register(new RuntimeExceptionMapper());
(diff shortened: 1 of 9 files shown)


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -35,6 +35,7 @@ public boolean configure(FeatureContext context) {
         context.register(new IllegalArgumentExceptionMapper());
         context.register(new NoContentExceptionMapper());
         context.register(new RuntimeExceptionMapper());
+        context.register(new WebApplicationExceptionMapper());
```
iamdanfox, line 38: Can we also add the javadoc explanation for why we need this to the top of the WebApplicationExceptionMapper class?

dansanduleac: Sure. I'd also like to remove the log.error from this mapper, since people are gonna be really confused about it

iamdanfox (approved):

PR #788, opened 2018-08-03
[WIP] jersey-server BOM inherits jersey-parent recommendations

dansanduleac:


build.gradle
```diff
@@ -7,10 +7,13 @@ buildscript {
     dependencies {
         classpath 'gradle.plugin.com.palantir:gradle-circle-style:1.1.2'
         classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.4'
-        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.3'
-        classpath 'com.netflix.nebula:gradle-dependency-lock-plugin:4.9.5'
-        classpath 'com.netflix.nebula:nebula-dependency-recommender:6.0.0'
-        classpath 'com.netflix.nebula:nebula-publishing-plugin:5.1.5'
+        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.3', {
+            // This brings in maven 2, conflicts with our maven dependencies from gradle/nebula-dependency-recommender
+            exclude module: 'maven-ant-tasks'
```
dansanduleac, line 12: This is sad that we have to do this, but a dep of this guy, namely `org.codehaus.groovy.modules.http-builder:http-builder`, brings in this old library which no longer exists in maven 3.

Since both nebula dependency recommender and gradle use maven 3, this dependency's presence was causing runtime JVM errors when calling `netflix.nebula.dependency.recommender.provider.MavenBomRecommendationProvider#getMavenRecommendations`.

It seems I can run bintrayUpload (at least locally) without this and it doesn't crash in an unexpected way, so maybe it's fine.

```
+--- com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.3
|    +--- org.codehaus.groovy.modules.http-builder:http-builder:0.7.2
|    |    +--- org.apache.httpcomponents:httpclient:4.2.1 -> 4.3.6
|    |    |    ...
|    |    +--- net.sf.json-lib:json-lib:2.3
|    |    |    ...
|    |    +--- net.sourceforge.nekohtml:nekohtml:1.9.16
|    |    |    ...
|    |    \--- xml-resolver:xml-resolver:1.2
|    \--- org.apache.maven:maven-ant-tasks:2.1.3
```

uschi2000:
What's the motivation for this change?

gradle/publish-bom.gradle
```diff
@@ -1,10 +1,41 @@
...
-        }
+configurations {
+    bom {
+        extendsFrom compile
     }
 }
 
 bintray { pkg { publications += ['nebulaBom'] } }
+
+afterEvaluate {
```
iamdanfox, line 9: I'm not really comfortable with this much extra complexity in our gradle, for unknown upside.  If the boms are going to be super valuable can we make a concerted effort to make this simpler/encapsulated, otherwise, can we just close?

dansanduleac:
@uschi2000 wc having to re-pin the jersey version (and possibly other versions which could come from the jersey-parent bom) and leaving comments saying "this should be pinned to http-remoting's version" even though the pin isn't even for the same version anymore

PR #789, opened 2018-08-07
publish BOM for jaxrs-client

qinfchen:


iamdanfox:
Verified the pom looks good from the CircleCI artifacts tab: https://2412-43088077-gh.circle-artifacts.com/0/home/circleci/poms/jaxrs-clients/build/publications/nebulaBom/pom-default.xml

iamdanfox (approved):
Approving because `:retrofit2-clients` [already publishes this BOM](https://github.com/palantir/http-remoting/blob/3.41.1/retrofit2-clients/build.gradle#L4), so we might as well publish one for :jaxrs-clients too

PR #794, opened 2018-08-30
Upgrade OkHttp to 3.11.0.

jkozlowski:
Potentially supersedes https://github.com/palantir/http-remoting/pull/747.

## Why?

The 2 releases since 3.9.0 fixed bugs in Http2 handling (especially wrt. flow control).

See:
* https://github.com/square/okhttp/pull/3920
* https://github.com/square/okhttp/pull/3960
* https://github.com/square/okhttp/pull/4033
* https://github.com/square/okhttp/pull/4127

**These bugs have been hit by an internal product and are blocking feature work.** 

## Remaining issues:

- [x] - New: Honor the Retry-After response header. HTTP 503 (Unavailable) responses are retried automatically if this header is present and its delay is 0 seconds. HTTP 408 (Client Timeout) responses are retried automatically if the header is absent or its delay is 0 seconds. @uschi2000 
- [x] 3.10.0 removed support for hostname verification via CN @cakofony 
- [x] There is one test in the repo that checks handling of status code 100, but 100 is kinda special and magical, see linked resources. @jkozlowski
- [x] okhttp.Request#tag can now be associated with multiple values (keyed by Class instances). We reset the default tag in CurrentUrlInterceptor and RoundRobinInterceptor. @uschi2000 I wasn't sure what that is supposed to do so I preserved the behavior, but I'm not sure if we need to adjust it or not.


ferozco (approved):
Seems like a straight forward bump, I don't think any of the issues mentioned would affect us

jkozlowski:
@ferozco what should I do about the OkHttpClientTest#handlesSuccessfulResponseCodesWithSuccessHandler?

iamdanfox:
Note that on the last PR, there were a couple of questions:

from @uschi2000 

What does this mean for us?
```
- New: Honor the Retry-After response header. HTTP 503 (Unavailable) responses are retried automatically if this header is present and its delay is 0 seconds. HTTP 408 (Client Timeout) responses are retried automatically if the header is absent or its delay is 0 seconds.
```

from @cakofony 

Note that 3.10.0 removed support for hostname verification via CN, and only validates SAN values.
This is ideal and matches modern versions of chrome, though if we're worried about breaking things we can supply our own HostnameVerifier instance to okhttp.

ferozco:
I think its fine to omit since it is 100s are such a special case

jkozlowski:
@iamdanfox Yep, I copied those in this PR description, see above. Would like to get the opinions of those people, ideally.

@j-baker had an opinion on the SAN thing (shouldn't affect us), Retry-After I would need to check if we send those on QoS exceptions, but I don't believe so.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/CurrentUrlInterceptor.java
```diff
@@ -44,7 +44,7 @@ public Response intercept(Chain chain) throws IOException {
                 .url(rebasedUrl)
                 // Request.this.tag field by default points to request itself if it was not set in RequestBuilder.
                 // We don't want to reference old request in new one - that is why we need to reset tag to null.
-                .tag(originalRequest.tag().equals(originalRequest) ? null : originalRequest.tag())
+                .tag(originalRequest.equals(originalRequest.tag()) ? null : originalRequest.tag())
```
uschi2000: was this a bug? if so, can you make it a separate PR, please?

j-baker: no, they broke

j-baker: tags are now a map keyed by Class, and the old one is the Object case

j-baker: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Request.java#L137

j-baker: vs https://github.com/square/okhttp/blob/parent-3.9.0/okhttp/src/main/java/okhttp3/Request.java#L42

jkozlowski: and it returns null

j-baker: @jkozlowski this whole block should no longer be necessary, looking at the new code - don't need to unset the tag

j-baker: but maybe we want it to continue working the way it does for people who force their okhttp version, and so this'd be fine

j-baker: cool ignore me :)

uschi2000:
re 503: i'd prefer we continue using our explicit 503 handling and turn off the okhttp one. similar to the existing
```
        client.followRedirects(false);  // We implement our own redirect logic.
```

j-baker:
you don't get to pick

j-baker:
https://github.com/square/okhttp/pull/3753/files

j-baker:
Unless we want to make all of our bodies Unrepeatable, which is its own can of worms

jkozlowski:
Right, I'll clean up the test tomorrow, look into the 503 (although looks like we can't opt out of that behaviour). 

Would be good to understand why we thought we needed to unset the tags and what's the behaviour we want.

okhttp-clients/src/main/java/com/palantir/remoting3/okhttp/CurrentUrlInterceptor.java
jkozlowski: Soo, here's my understanding: in 3.9.0 the Request.Builder#tag method had the following javadoc:

```
    /**
     * Attaches {@code tag} to the request. It can be used later to cancel the request. If the tag
     * is unspecified or null, the request is canceled by using the request itself as the tag.
     */
```

The value, as @j-baker pointed out, is automatically set in Request(Builder) constructor to Request.this, and then propagated across to the next builder when the request is cloned (e.g. for retries). This means it builds a linked list of requests which means they all linger on in memory.

It would seem to me this was an omission and not intentional behavior: this is now gone in OkHttp 3.11 so I think we are good to remove this.

jkozlowski: Going to resolve this point, unless people object.

jaxrs-clients/versions.lock
```diff
@@ -211,13 +211,13 @@
             ]
         },
         "com.squareup.okhttp3:logging-interceptor": {
```
jkozlowski, line 213: Let's discuss SAN vs CN verification here: the only thing that I can think would break were internal ETE tests, but we verified with a standard test certificate and they are already setup with SAN pointing to localhost and internal test domains.

As far as my understanding goes, we should already be generating correct certs in production, so I think we can close this issue?

j-baker: Yes

versions.props
```diff
@@ -9,7 +9,7 @@ com.netflix.feign:feign-*= 8.17.0
 com.palantir.remoting-api:* = 1.9.0
 com.palantir.safe-logging:* = 1.3.0
 com.palantir.tritium:tritium-registry = 0.8.0
-com.squareup.okhttp3:* = 3.9.0
+com.squareup.okhttp3:* = 3.11.0
```
jkozlowski, line 12: Now for Retry-After handling, let's discuss here. The behavior change is as follows:

1) 408 (CLIENT_TIMEOUT) is now automatically retried if Retry-After header is not set or is set to 0, as per above and here's the code: (https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java#L367). I don't think we particularly care about this, but in practice it would mean one extra retry: it will give up if THAT extra retry fails.
2) 503 are retried if Retry-After is set to 0, here's the code: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java#L374. Again, I don't believe we ever set Retry-After in 503s and again it means one extra retry (it will bail if that retry fails). Someone could still do this by throwing javax.ws.rs.ServiceUnavailableException.

So, I think we should be good here too, unless anyone objects?.

jkozlowski:
Can you guys review the reasoning and merge if we think this is good?

jkozlowski:
@iamdanfox, @uschi2000, @ferozco ping, can we get a merge/no-merge?

PR #799, opened 2018-09-06
No more ThreadLocals, delegate everything to 'palantir/tracing-java'

qinfchen:
## Before this PR

People might start adopting our shiny new [palantir/tracing-java](https://github.com/palantir/tracing-java) libraries, while keeping remoting3 ones on the classpath.  This would be **very bad** because there would be two ThreadLocals trying to keep track of the current trace:

- com.palantir.remoting3.tracing.Tracer <-- one ThreadLocal in here
- com.palantir.tracing.Tracer           <-- another ThreadLocal in here!

Concretely, all the clients in Atlas would use the remoting3 tracers but newer user-code would likely use tracing-java.  This would make tracing information far less useful for debugging.

## After this PR

`com.palantir.remoting3.tracing.Tracer` no longer maintains its own ThreadLocal, instead it delegates everything to `com.palantir.tracing.Tracer`.

## Caveats

- 🚨The return type of `Tracer#subscribe(String, SpanObserver)` behaves differently now (it used to preserve reference equality and it no longer does).
- 🚨The `Tracer#getAndClearTrace` returns an incomplete copy of the trace.  I think this is OK because that method was never usable externally because the returned type was not public.
- Should there be a way of detecting this delegating functionality is on the classpath?

--------

When migrating tritium to the rebranded tracing, @j-baker and @schlosna pointed out that that tracers are threadlocal and could potentially break tracing if `remoting-tracers`  and `conjure tracers` were not on the same thread. 
One proposal suggested by @iamdanfox was to delegate remotings tracing calls to conjure tracers. This pr is to demonstrate how the implementation would look like. 
See more details here: https://github.com/palantir/tritium/pull/100.


tracing/src/main/java/com/palantir/remoting3/tracing/AsyncSlf4jSpanObserver.java
```diff
@@ -192,11 +50,4 @@ public static AsyncSlf4jSpanObserver of(
             String serviceName, InetAddress ip, Logger logger, ExecutorService executorService) {
         return new AsyncSlf4jSpanObserver(serviceName, ip, logger, executorService);
     }
-
-    @Override
-    public void doConsume(Span span) {
```
iamdanfox, line 197: this was an implementation detail of the abstract `AsyncSpanObserver` class (which we no longer extend), so I think it's ok to delete.

tracing/src/main/java/com/palantir/remoting3/tracing/AsyncSlf4jSpanObserver.java
```diff
@@ -16,171 +16,29 @@
...
+public final class AsyncSlf4jSpanObserver implements SpanObserver {
 
-    private final Logger logger;
-    private final ZipkinCompatEndpoint endpoint;
+    private final com.palantir.tracing.AsyncSlf4jSpanObserver delegate;
 
-    @JsonSerialize(as = ImmutableZipkinCompatSpan.class)
-    @Value.Immutable
-    @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE)
-    abstract static class ZipkinCompatSpan {
```
iamdanfox, line 51: These classes have been completely deleted because they're an implementation detail that now exists in the new tracing-java class: `com.palantir.tracing.AsyncSlf4jSpanObserver`

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -45,156 +41,74 @@
 
     private Tracer() {}
 
-    // Thread-safe since thread-local
-    private static final ThreadLocal<Trace> currentTrace = ThreadLocal.withInitial(() -> {
```
iamdanfox: This is the critical line! remoting3 no longer maintains any ThreadLocal state, instead it's all delegated to the shiny new tracing-java.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
```diff
@@ -204,74 +113,39 @@ private static Span toSpan(OpenSpan openSpan, Map<String, String> metadata) {
...
+        return com.palantir.tracing.Tracer.getTraceId();
     }
 
     /** Clears the current trace id and returns (a copy of) it. */
     public static Trace getAndClearTrace() {
-        Trace trace = currentTrace.get();
-        currentTrace.remove();
-        MDC.remove(Tracers.TRACE_ID_KEY);
-        return trace;
+        com.palantir.tracing.Trace trace = com.palantir.tracing.Tracer.getAndClearTrace();
```
qinfchen, line 134: @iamdanfox, is this possible that we do something like this instead of having the ExposedTrace that lives inside the tracing-java package?
```diff
-        com.palantir.tracing.Trace trace = com.palantir.tracing.Tracer.getAndClearTrace();
-        return new Trace(ExposedTrace.isObservable(trace), ExposedTrace.getTraceId(trace));
+        boolean isObservable = com.palantir.tracing.Tracer.isTraceObservable();
+        String traceId = com.palantir.tracing.Tracer.getTraceId();
+        com.palantir.tracing.Tracer.getAndClearTrace();
+        return new Trace(isObservable, traceId);
```

iamdanfox: I think there's a bit of a race condition, where we might get the isTraceObservable info, then someone changes stuff beneath our feet, then we get the traceId, then finally we clear it...

qinfchen: yeah, that is not gonna work.

tracing/src/main/java/com/palantir/remoting3/tracing/CloseableTracer.java
```diff
@@ -38,12 +38,12 @@ private CloseableTracer() { }
      * Opens a new {@link SpanType#LOCAL LOCAL} span for this thread's call trace, labeled with the provided operation.
      */
     public static CloseableTracer startSpan(String operation) {
-        Tracer.startSpan(operation);
+        com.palantir.tracing.Tracer.startSpan(operation);
```
j-baker, line 41: can we just be like com.palantir.tracing.CloseableTracer.startSpan?

qinfchen: i kind of prefer using `com.palantir.tracing.Tracer.startSpan(operation);` to match up against `com.palantir.tracing.Tracer.fastCompleteSpan();` few lines below since we can't get the instance of the `com.palantir.tracing.CloseableTracer`.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
```diff
@@ -22,8 +22,6 @@
 
 /** Utility methods for making {@link ExecutorService} and {@link Runnable} instances tracing-aware. */
 public final class Tracers {
-    /** The key under which trace ids are inserted into SLF4J {@link org.slf4j.MDC MDCs}. */
-    public static final String TRACE_ID_KEY = "traceId";
```
iamdanfox: I don't think we should delete public static final fields like this, because if some consumer is depending on `com.palantir.remoting3.tracingTracers.TRACE_ID_KEY` then they'd suddenly break?

qinfchen: it doesn't seem like anyone is using it, https://github.com/search?p=2&q=Tracers.TRACE_ID_KEY&type=Code
i also did a quick search on the key internally and  no internal consumer is using it.

tracing/src/main/java/com/palantir/remoting3/tracing/Tracer.java
j-baker: In every case in this impl we should be delegating all state :)

tracing/src/main/java/com/palantir/remoting3/tracing/Trace.java
```diff
@@ -28,7 +28,7 @@
  * Represents a trace as an ordered list of non-completed spans. Supports adding and removing of spans. This class is
  * not thread-safe and is intended to be used in a thread-local context.
  */
-final class Trace {
+public final class Trace {
```
j-baker: does this need to become public?

iamdanfox: Without this being `public`, one of the test utils fails to compile.

screen shot 2018-09-06 at 17 26 13

I find an alternative that involved some objects and casting too!

qinfchen: it is also kind of odd that we have a public static method that returns `Trace` and `Trace` is private.
https://github.com/palantir/http-remoting/pull/799/files#diff-7ee517b85058e51de91b9bf446ff7585L246

tracing/src/main/java/com/palantir/remoting3/tracing/Tracers.java
iamdanfox: I still don't think we should remove things like this when there isn't a clear motivation - what if there was a user that didn't appear in your search?  Did you check OSS consumers too (e.g. atlasdb?)

tracing/src/main/java/com/palantir/remoting3/tracing/Trace.java
iamdanfox: @qinfchen I don't have the original context, but one possible justification could be that this Trace class is useful for testing but shouldn't be used by consumers??

ferozco (approved):

PR #801, opened 2018-09-12
Stop propagating RemoteExceptions

ellisjoe:
## What is happening

We have concerns that people are introspecting RemoteExceptions that are not actually part of a server's API but instead came from a further upstream server and are propagated by the server's `RemoteExceptionMapper` when one of the server's clients encountered a failed HTTP call.

## What should happen

Instead of blindly propagating uncaught RemoteExceptions, this library should always send 500 responses corresponding to an INTERNAL error.

If devs actually wanted this to be part of the API, they should wrap their client in a `catch` block and re-throw one of their own _ServiceExceptions_.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
@@ -18,8 +18,9 @@ package com.palantir.conjure.java.server.jersey;
 
 import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.conjure.java.api.errors.RemoteException;
-import com.palantir.conjure.java.api.errors.ServiceException;
+import com.palantir.conjure.java.api.errors.SerializableError;
 import com.palantir.logsafe.SafeArg;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.ext.ExceptionMapper;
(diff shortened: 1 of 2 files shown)


jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -40,35 +39,25 @@
...
 @Provider
 final class RemoteExceptionMapper implements ExceptionMapper<RemoteException> {
 
     private static final Logger log = LoggerFactory.getLogger(RemoteExceptionMapper.class);
+    private static final ServiceExceptionMapper mapper = new ServiceExceptionMapper();
 
     @Override
     public Response toResponse(RemoteException exception) {
         Status status = Status.fromStatusCode(exception.getStatus());
+        ServiceException servieException = new ServiceException(ErrorType.INTERNAL, exception);
```
iamdanfox: WOuld prefer to inline the relevant bits of ServiceException mapper to avoid getting another stacktrace

jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/RemoteExceptionMapper.java
```diff
@@ -40,35 +39,25 @@
...
 
-        // log at WARN instead of ERROR because although this indicates an issue in a remote server, it is not
-        log.warn("Forwarding response and status code from remote server back to caller",
+        // log at WARN instead of ERROR because although this indicates an issue in a remote server
+        log.warn("Encountered a remote exception. Wrapping as an internal service exception before responding",
                 SafeArg.of("statusCode", status.getStatusCode()),
-                exception);
+                servieException);
```
iamdanfox: servie

iamdanfox (commented):
Please could we re-target this to develop?  I think the class is now: https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java

iamdanfox (approved):

gregakinman:
The version history of this repo is kind of confusing, can someone confirm that this is only in 4.x?

gregakinman:
Was this released with the major version bump to 4.x?

gregakinman:
@ellisjoe @iamdanfox ping

ellisjoe:
it merged as part of `4.1.1`

gregakinman:
Hmm, ok. Isn't this a breaking change? (It results in different behavior for servers that pick up that version). Also, regardless of if it's a breaking change or not, it's not even listed in the changelog for 4.1.1. At the very least we should be documenting behavior changes...

ellisjoe:
I'd really like to just fix the problem here. It's never been acceptable to rely on transitive service specific exceptions in consumer code and that's what we'd like to ban with the new implementation. There appear to be a few transitive exceptions that are truly the fault of the client and should thus be propagated.

gregakinman:
To clarify: I'm not commenting here on whether this change should have been made or not. That conversation is happening in https://github.com/palantir/conjure-java-runtime/issues/834. I'm commenting here on the manner in which it was made - a behavior change was undocumented and released without a major version bump. I think we can all agree it was at least unacceptable that it was undocumented, given the gravity of the change, and there's also an argument it it's a breaking change, depending on how we want to define "breaking" for this project.

There's nothing we can do post-hoc here - I just want to remind the maintainers of this project to include all notable behavioral changes in release notes in the future.

ellisjoe:
re documenting: yes should have been documented, no argument there, though it would not have been marked as breaking because we didn't believe this would break people. transitive service exceptions are not defined in a service's api and should therefore not be exposed through the service's api.
re breaking: the fact that this change was breaking was unintentional and should be fixed

pnepywoda:
reverting in https://github.com/palantir/conjure-java-runtime/pull/852

PR #809, opened 2018-09-13
Merge remoting into develop

j-baker:
It's a bit goofy because it seems there was some squashing and merging done, but i think this should get it going right!


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -0,0 +1,124 @@
...
+            limiters.replace(name, limiter, newLimiter());
+            return acquireLimiter(name);
+        });
+    }
+
+    private Limiter<Void> newLimiter() {
+        Limiter<Void> limiter = new InstrumentedLimiter(SimpleLimiter.newBuilder()
+                .limit(new RemotingWindowedLimit(VegasLimit.newDefault()))
+                .build(), slowAcquire);
+        return RemotingBlockingLimiter.wrap(limiter, timeout);
```
dansanduleac: can we call this ConjureBlockingLimiter?

j-baker: Done!

dansanduleac (approved):

PR #815, opened 2018-09-17
[fix] Add ServiceClass and hr-path-template params to log.warn line

### LINKED ISSUE #814: Hard to debug "Timed out waiting to get permits for concurrency." message
## What happened?

An internal product upgraded and suddenly started getting lots of warning lines:

```
Timed out waiting to get permits for concurrency.
```

https://github.com/palantir/conjure-java-runtime/pull/786/files#diff-15c82af75a7197b006f2ab86da3abe06R76

These are very hard to debug because they don't contain a stack trace or service name.

Some SafeArg tags should be added to help devs identify which network calls are problematic here!

iamdanfox:
fixes https://github.com/palantir/conjure-java-runtime/issues/814

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -83,7 +83,17 @@ class ConcurrencyLimiters {
     Limiter.Listener acquireLimiterInternal(String limiterKey, int attemptsSoFar) {
         Limiter<Void> limiter = limiters.computeIfAbsent(limiterKey, key -> newLimiter());
         Optional<Limiter.Listener> listener = limiter.acquire(NO_CONTEXT);
-        return listener.orElseGet(() -> {
+
+        if (listener.isPresent()) {
+            if (attemptsSoFar > 0) {
+                log.info("Eventually acquired concurrency permit",
+                        SafeArg.of("serviceClass", serviceClass),
+                        SafeArg.of("limiterKey", limiterKey),
+                        SafeArg.of("attemptsSoFar", attemptsSoFar + 1));
+            }
+
+            return listener.get();
+        } else {
             if (Thread.currentThread().isInterrupted()) {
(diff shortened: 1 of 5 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -76,9 +78,11 @@
             log.warn("Timed out waiting to get permits for concurrency. In most cases this would indicate "
                             + "some kind of deadlock. We expect that either this is caused by not closing response "
                             + "bodies (there should be OkHttp log lines indicating this), or service overloading.",
+                    SafeArg.of("serviceClass", serviceClass),
+                    SafeArg.of("limiterKey", limiterKey),
                     SafeArg.of("timeout", timeout));
-            limiters.replace(name, limiter, newLimiter());
-            return acquireLimiter(name);
+            limiters.replace(limiterKey, limiter, newLimiter());
+            return acquireLimiter(limiterKey);
```
dansanduleac: As discussed offline, please can we add a counter here (and log it) to track how many times this function has been recursed into?
Optionally, if finally `listener.isPresent()` after a number of tries, we could also log that it's succeeded.

dansanduleac: Also, can we add a metric around 'total time waiting for a permit'?
`InstrumentedLimiter` seems to track a single invocation of `Limiter::acquire` into the `slowAcquire` timer, but nothing tracks the total duration of recursive acquire attempts (the total duration of an initial `acquireLimiter` call).
That seems more useful as the former would never record a time higher than the timeout-per-acquire-attempt, so it's harder to see from the metrics how much a particular request is being delayed by this limiter (we instrument the total request time but that includes network time too).

This can be done as part of another PR, but I think it's important to follow-up on.

dansanduleac (approved):
Approved pending adding a counter to acquireLimiter

j-baker:
beautiful

PR #817, opened 2018-09-18
[fix] Empty responses now release concurrency permits 

j-baker:
Sadly I accidentally skipped a line. Would like this merged + released
ASAP; can coordinate internally beyond that.


iamdanfox (approved):
Good spot!

PR #827, opened 2018-09-26
Log error names in exception mappers

pkoenig10:
Alternative to https://github.com/palantir/conjure-java-runtime-api/pull/131

This implementation matches what we currently do for `errorInstanceId` and is less hacky than adding the errorType information to the `ServiceException` args.


iamdanfox (approved):

pkoenig10:
I would much rather add the `errorInstanceId` to the `ServiceException` (see my [comment](https://github.com/palantir/conjure-java-runtime-api/pull/131#issuecomment-424902747)) so we get the args wherever service exceptions are logged. Then it should be removed from here so we don't get duplicate args.

PR #828, opened 2018-09-26
[improvement] upgrade concurrency-limits library

j-baker:
In particular, reduce window size according with Netflix
recommendations, and remove our blocking limiter as per comment.
Also, add a span for how long it took to acquire the limiters!


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConjureWindowedLimit.java
```diff
@@ -38,7 +38,7 @@
     /**
      * Minimum observed samples to filter out sample windows with not enough significant samples.
      */
-    private static final int WINDOW_SIZE = 100;
+    private static final int WINDOW_SIZE = 10;
```
iamdanfox, line 41: stays in sync with the netflix class: https://github.com/Netflix/concurrency-limits/commit/b108aa802a98086117a95afb52e43c865a8c4160

iamdanfox (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConjureBlockingLimiter.java
```diff
@@ -1,122 +0,0 @@
...
-import java.time.Duration;
-import java.time.Instant;
-import java.util.Optional;
-
-/**
- * {@link Limiter} that blocks the caller when the limit has been reached.  The caller is
- * blocked until the limiter has been released.  This limiter is commonly used in batch
- * clients that use the limiter as a back-pressure mechanism.
- *
- * This limiter has been forked from Netflix's {@link com.netflix.concurrency.limits.limiter.BlockingLimiter}.
```
iamdanfox, line 31: Always nice to unfork :)

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -99,8 +103,9 @@
 
             return listener.get();
         } else {
+            // it returns empty if we've timed out, or we've been interrupted
             if (Thread.currentThread().isInterrupted()) {
-                throw new RuntimeException("Thread was interrupted");
+                throw new IOException("Thread was interrupted");
```
iamdanfox, line 108: This is important because async okhttp calls only catch IOExceptions

PR #829, opened 2018-09-27
[improvement] allow returning of ListenableFutures as well as Completable.

j-baker:
There's something goofy where unlike every other CompletableFuture, ours
can be cancelled (because we subclass CompletableFuture).

I would recommend we switch to using ListenableFutures, because dealing
with cancellation properly is the M.O. of ListenableFutures and is super
fragile with CompletableFutures.


conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
```diff
@@ -43,27 +51,87 @@ private AsyncSerializableErrorCallAdapterFactory() {}
...
+
+        @Override
+        public Type responseType() {
+            return responseType;
+        }
+
+        @Override
+        public ListenableFuture<R> adapt(final Call<R> call) {
+            ListenableFutureCallback<R> callback = new ListenableFutureCallback<>(call);
+            call.enqueue(callback);
```
iamdanfox, line 126: this is really cute :)

iamdanfox (approved):

ferozco (approved):

PR #843, opened 2018-10-15
[fix] Preserve OkHostnameVerifier from okhttp 3.9

iamdanfox:
## Before this PR

An earlier PR (https://github.com/palantir/conjure-java-runtime/pull/794) upgraded okhttp from 3.9.0 -> 3.11.0.  This change from [okhttp 3.10.0](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-3100) meant conjure-java-runtime clients were unable to talk to a particular type of internal server (as it is currently only given certs with a CN and not a proper SAN):

> New: Don't fall back to common name (CN) verification for hostnames. This behavior was deprecated with RFC 2818 in May 2000 and was recently dropped from major web browsers.

## After this PR

Jersey and Retrofit clients can be configured to use the old deprecated CN verification fallback behaviour.

TODO:

- is "fallbackToCommonNameVerification" the right boolean option name to add to conjure-java-runtime-api?
- agree on a timeframe when this can be removed

### DIFF
--- a/.baseline/checkstyle/checkstyle-suppressions.xml
+++ b/.baseline/checkstyle/checkstyle-suppressions.xml
@@ -26,4 +26,6 @@
     <suppress files="[/\\].*[/\\]generated.*[/\\]" checks="." />
 
     <suppress files=".*AsyncSerializableErrorCallAdapterFactory\.java" checks="RegexpHeader" />
+    <suppress files=".*DistinguishedNameParser\.java" checks="." />
+    <suppress files=".*Okhttp39HostnameVerifier\.java" checks="." />
 </suppressions>
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DistinguishedNameParser.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DistinguishedNameParser.java
@@ -1,0 +1,414 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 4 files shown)


tpetracca (approved):

robert3005:
@iamdanfox what's latest here? Do we still want to merge this?

iamdanfox:
Yep, if nobody has any concerns about the name then I just need to make a PR to add this extra field to the conjure-java-runtime-api!

darora:
lgtm

carterkozak:
Is there more context somewhere? This seems dangerous and counterproductive. Can we not fix the service with bad certs?

iamdanfox:
We are not intending to merge this as is - we're going to put it behind a feature flag, which is only expected to be enabled on the (3?) internal services which talk to axiom.

j-baker:
We are not intending to merge this as is - we're going to put it behind a feature flag, which is only expected to be enabled on the (3?) internal services which talk to the relevant internal server.

robert3005:
@iamdanfox do you have timeline for adding the feature flag or should I submit the pr for that?

tpetracca:
@iamdanfox @robert3005 is this waiting on something?

kmblake:
@j-baker any updates on the feature flag? This issue became a problem for gemini after we upgraded conjure-java-runtime because certain axiom hosts only include hostname in the CN, so gemini was running into a bunch of `javax.net.ssl.SSLPeerUnverifiedException` errors (see internal issue 2694). We're working on making sure that all axiom instances include hostname in the SAN, but we would definitely like to use this legacy CN verification option in the meantime.

robert3005:
This is superseded by #907

PR #844, opened 2018-10-17
[improvement] change the concurrency algorithm to AIMD

j-baker:
I've come to the conclusion that I don't trust the knobs and levers
available right now to properly tune this and investigate places where
it goes wrong.

This PR flips the algorithm to be AIMD; this means that it only does
stuff in response to explicit 429/503 responses and so makes it highly
unlikely to be problematic.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -334,12 +334,9 @@ supported for media type `application/json`.
 
 Flow control in Conjure is a collaborative effort between servers and clients.
 
-Clients estimate the server's queue size using response delay and proactively throttle outgoing requests to minimize bottleneck queueing.  Servers may also advertise an overloaded state using 429/503 responses, which clients interpret with more drastic throttling. This results in a mechanism similar to TCP/IP flow-control and minimizes both client and server-side errors.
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -334,12 +334,9 @@ supported for media type `application/json`.
 
 Flow control in Conjure is a collaborative effort between servers and clients.
 
-Clients estimate the server's queue size using response delay and proactively throttle outgoing requests to minimize bottleneck queueing.  Servers may also advertise an overloaded state using 429/503 responses, which clients interpret with more drastic throttling. This results in a mechanism similar to TCP/IP flow-control and minimizes both client and server-side errors.
+Servers advertise an overloaded state using 429/503 responses, which clients interpret by throttling the number of in-flight requests they will send (currently according to an additive increase, multiplicative decrease based algorithm). Requests are retried a fixed number of times, scheduled with an exponential backoff algorithm.
```
ferozco, line 337: probably worth while noting the factors that we use, or that we proxy the defaults from netflix

ferozco (approved):

readme.md
j-baker: given the previous description went stale so fast, I'd prefer we keep this to minimal defaults.

PR #850, opened 2018-10-23
Make sure lock verification doesn't run with locks applied

robert3005:
nebula.dependency-lock will check start parameters of your gradle invocation for existence of `generateLock` argument. In cases where we are running `verifyDependencyLocksAreCurrent` or `build` we add `generateLock` to the command line arguments.

https://github.com/nebula-plugins/gradle-dependency-lock-plugin/blob/master/src/main/kotlin/nebula/plugin/dependencylock/DependencyLockPlugin.kt#L129 for reference

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,10 @@ subprojects {
     apply plugin: 'java'
 
     sourceCompatibility = 1.8
+    if (gradle.startParameter.taskNames.find {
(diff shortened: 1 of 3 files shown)


iamdanfox (approved):
Looks like this ran the right task on CI.  It's pretty gross that we have to workaround like this, but I think our own lock-files thing is really the long-term solution here.

screen shot 2018-10-23 at 17 35 06

Issue #834, opened 2018-10-09
Consider propagating certain remote exceptions


ellisjoe:
## What happened?

Service A calls Service B which throws a 403 Unauthorized exception. Currently Service A will rethrow that a an Internal Error (see: https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java) .

## What did you want to happen?

We started mapping `RemoteExceptions` to Internal Errors here: https://github.com/palantir/conjure-java-runtime/pull/801. We'd like to continue to do that, though there seems to be some class of `RemoteExceptions` that may be acceptable to propagate. An initial proposal would be to propagate 403s directly, rather than remapping.

pnepywoda:
Propagating 403 is pretty important, as it should be displayed to users

gregakinman:
From what I can tell, the decision made in #801 was a lesser-of-two-evils choice between:

- clients potentially introspecting RemoteExceptions from an upstream server but having whatever error happened be potentially more clear (you might get a UpstreamServer:SpecificError)
- clients only introspecting RemoteExceptions from the server they're talking to directly but having whatever error happened potentially be obscured if the downstream server doesn't catch the upstream error

Are we sure that the latter situation is better than the former?

Now, every client has to update every time a server adds a new type of error. And before clients update at all, no exceptions get propagated and everything is an opaque internal error. Additionally, you can get situations like the following: a server and a Spark application both live in the same repository and are considered part of the same service, despite being two separate processes. They could share the same error namespace. Previously, the Spark application could throw an error which would then just pass through the server. Now, though, the server will have to catch and rethrow each of these errors to preserve them, which adds a lot of verbosity.

pnepywoda:
There's definitely a certain class of errors you'd want to propagate all the way to the calling user, like NotFound or PermissionDenied. Other than that, we wouldn't want say an OutOfDiskSpace database error returning to the user so anything else would have to be whitelisted if you actually care to keep the error name when displaying to the user

gregakinman:
Why not?

gregakinman:
Guess it could seem out of the blue. But it's more informative than "internal"...

pnepywoda:
Because there's nothing the user can do about it.
Maybe a better example is an endpoint which does paging and the caller passed in larger than is acceptable. In this case you really don't want to show the user an error in the calling service

[...]

PR #852, opened 2018-10-25
Revert "Stop propagating RemoteExceptions (#801)"

pnepywoda:
This was a behavioral change without a major version bump and it's effects are far-reaching (clients will no longer get the exception names/codes they expect)

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
@@ -16,12 +16,12 @@
 
 package com.palantir.conjure.java.server.jersey;
 
-import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableError;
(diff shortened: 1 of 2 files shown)


bkrieger:
+1

iamdanfox (changes requested):
I don't want to prematurely revert this.

Discussion happening on https://github.com/palantir/conjure-java-runtime/issues/834.

pnepywoda:
@iamdanfox not-propagating exceptions is still a regression in functionality and it affects our ability to diagnose issues (when we see `Internal` instead of `RequestedItemNotFound` as the error name). 

I'm still strongly in-favor of reverting this until we've discussed an actual path forward that isn't a regression.

pkoenig10:
I agree with @pnepywoda. Prior to this change there was not clear messaging that clients should not be introspecting upstream remote exceptions. While I don't disagree with the goal of the preventing this behavior, it appears that little consideration was given to the full consequences of this change. 

Upstream exceptions are still incredibly useful when used purely informationally (to provide error messages, etc.). When used in this way, clients aren't introspecting exception they shouldn't be but are still able to provide rich, useful error information.

markelliot:
We understand your position but as discussed on linked issues isn’t something this project is going to revert. We’d continue to note that if you value forwarding specific errors your service can absolutely continue to do so, explicitly.

Issue #813, opened 2018-09-15
Endpoint returning Alias of optional<T> + Http 204 results in NeverReturnNullDecoder exception


nartmal:
## What happened?

An internal service has legacy, non-conjurized, endpoints that return null which breaks with the `NeverNullDecoder` being added as a default. We're pretty low in the stack so major releasing just to address this break is not a feasible option for us. This will prevent us from taking all future updates.

## What did you want to happen?

Either remove it completely or make the `NeverNullDecoder` optional.

iamdanfox:
@nartmal the initial implementation of this was added in #766 (released 3.40.0) as a defensive measure so that all users can be certain that they'll never have to deal with nulls (we always recommend using Optional instead of `null`). However, we quickly realized this had undesirable behaviour when the service should return a container type (List, Set, Optional, Map). This was fixed in #772 (released 3.41.0).

Which version were you experiencing problems with?  I don't really want to remove it completely as it is beneficial to the majority of our users.

If you have defined your API to have nullable methods then it's not really sufficient to stop taking upgrades as I think any consumer that uses modern conjure-java-runtime will be unable to use your nullable endpoints. I'd recommend adding an interface to your API that returns optionals and deprecating your existing endpoints.

nartmal:
This is closable, tliao & jbaker have a path to resolve without any action item here.

iamdanfox:
@nartmal can you share the approach you're planning to take just in case other people find themselves in the same situation?

tengfliao:
@iamdanfox The underlying issue was that there was a type `Type` that was an alias of `Optional<String>`. The alias wasn't being treated the same way as if it were an actual `Optional<String>` and the decoder interpreted it as a null. The approach we're planning on taking is not defining an alias and having java clients take the break.

iamdanfox:
I've encountered this problem myself. 

1. In the Conjure definition, we have an alias of an optional:

```yml
      OptionalAlias: { alias: optional<integer> }
...
      foo:
        http: GET /some-path
        returns: OptionalAlias
```

2. Then the server returns 204 (see below)

```
curl -v http://localhost:8000/some-path
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /some-path HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 204 No Content
< access-control-allow-origin: *
< content-length: 0
[...]

PR #865, opened 2018-11-07
[fix] feign.EmptyContainerDecoder handles aliases of optionals

iamdanfox:
## Before this PR

Users could define ostensibly valid APIs which return aliases of optionals, but feign would throw (see https://github.com/palantir/conjure-java-runtime/issues/813).

## After this PR

HTTP204 will be translated to a usable type, e.g.  `MyAlias.of(Optional.empty())`

----

I'm aware this is a bit gross, so just putting up a proof of concept to see if we can think of something better.  Long-term, if we replace feign with codegen (dialogue) then all of this will go away because we can just codegen the appropriate 204 check.

TODO:

- [x] run the new conjure-verification tests
- [x] hook up to AbstractFeignJaxRsClientBuilder
- [x] be defensive against the infinite recursion thing


conjure-java-jaxrs-client/src/main/java/feign/EmptyContainerDecoder.java
```diff
@@ -0,0 +1,175 @@
...
+ */
+public final class EmptyContainerDecoder implements Decoder {
+
+    private final LoadingCache<Type, Object> blankInstanceCache;
+    private final Decoder delegate;
+
+    public EmptyContainerDecoder(ObjectMapper mapper, Decoder delegate) {
+        this.delegate = delegate;
+        this.blankInstanceCache = Caffeine.newBuilder()
+                .maximumSize(10_000)
```
ferozco: 1000?

ferozco (approved):

carterkozak:
@iamdanfox this commit appears to break conjure-java client verification tests.

PR #879, opened 2018-11-15
Hide implementation dependencies

dansanduleac:
## Before this PR

Clients and servers were exposing their implementation dependencies in the POM.

## After this PR

All projects now hide their implementation dependencies (as in, they show up under scope `runtime` in the POM).

Also, we now use gradle's IMPROVED_POM_SUPPORT to achieve [compile/runtime-scope-separation-in-pom-consumption](https://docs.gradle.org/4.6/release-notes.html#compile/runtime-scope-separation-in-pom-consumption) for the dependencies of this repo's projects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ allprojects {
 }
 
 subprojects {
-    apply plugin: 'java'
+    apply plugin: 'java-library'
 
     sourceCompatibility = 1.8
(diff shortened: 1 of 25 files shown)


iamdanfox (approved):
Directionally I like this, but I wish there was a nice automated tool we could run to verify that we've marked things correctly as api / implementation!

PR #880, opened 2018-11-19
Run java11 tests 

robert3005:
excavator keeps on overwriting #846 

Made https://github.com/Netflix/concurrency-limits/pull/91 to get rid of the fork


versions.props
```diff
@@ -7,24 +7,24 @@ com.google.guava:guava = 23.6.1-jre
...
+io.dropwizard.metrics:metrics-core = 3.2.6
+io.dropwizard:dropwizard-* = 1.3.7
 io.reactivex.rxjava2:rxjava = 2.2.3
 javax.ws.rs:javax.ws.rs-api = 2.0.1
 junit:junit = 4.12
 org.apache.commons:commons-lang3 = 3.8.1
 org.assertj:assertj-core = 3.11.1
 org.bouncycastle:bcpkix-jdk15on = 1.60
-org.glassfish.jersey.*:* = 2.22.2
+org.glassfish.jersey.*:* = 2.25.1
```
iamdanfox, line 27: Apparently 2.26.0 onwards contains breaks (package renames), so this is the highest version that we can use: https://mvnrepository.com/artifact/org.glassfish.jersey.core/jersey-server

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/BlockingLimiter.java
```diff
@@ -0,0 +1,133 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Optional;
+
+/**
+ * {@link Limiter} that blocks the caller when the limit has been reached.  The caller is
+ * blocked until the limiter has been released, or a timeout is reached.  This limiter is
+ * commonly used in batch clients that use the limiter as a back-pressure mechanism.
+ */
+public final class BlockingLimiter<ContextT> implements Limiter<ContextT> {
```
iamdanfox, line 33: Should be safe to remove this once https://github.com/Netflix/concurrency-limits/pull/91 merges

iamdanfox (approved):

PR #890, opened 2018-11-26
[fix] Change rate limits to be asynchronous

robert3005:
On applications making a lot of small requests with rate limiting we have noticed apparent deadlocks where the application was no longer able to make any new requests thus not being able to serve any new incoming requests. This was traced down to okhttp dispatcher pool being full. Apart from the application itself rate limiting it's consumers this change lets us preserve previous limit behaviour and keep the limits.

## Before this PR
Rate limited requests could have occupied all of the okhttp dispatcher pool making it impossible to make any new outbound requests

## After this PR
Rate limited requests are tracked separately and no longer contribute to okhttp dispatcher pool usage.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -26,7 +26,6 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.SettableFuture;
 import com.netflix.concurrency.limits.Limiter;
 import com.netflix.concurrency.limits.limit.AIMDLimit;
-import com.netflix.concurrency.limits.limiter.BlockingLimiter;
 import com.netflix.concurrency.limits.limiter.SimpleLimiter;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.tracing.okhttp3.OkhttpTraceInterceptor;
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -19,27 +19,35 @@ package com.palantir.conjure.java.okhttp;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.SettableFuture;
 import com.netflix.concurrency.limits.Limiter;
 import com.netflix.concurrency.limits.limit.AIMDLimit;
 import com.netflix.concurrency.limits.limiter.BlockingLimiter;
 import com.netflix.concurrency.limits.limiter.SimpleLimiter;
 import com.palantir.logsafe.SafeArg;
-import com.palantir.tracing.CloseableTracer;
 import com.palantir.tracing.okhttp3.OkhttpTraceInterceptor;
 import com.palantir.tritium.metrics.registry.MetricName;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
-import java.io.IOException;
 import java.time.Duration;
+import java.util.ArrayDeque;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import javax.annotation.concurrent.GuardedBy;
 import okhttp3.Request;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@SuppressWarnings("DesignForExtension")
-class ConcurrencyLimiters {
+final class ConcurrencyLimiters {
     private static final Logger log = LoggerFactory.getLogger(ConcurrencyLimiters.class);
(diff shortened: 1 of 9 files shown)


okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/ConcurrencyLimitingInterceptorTest.java
```diff
@@ -1,145 +0,0 @@
...
-import okhttp3.ResponseBody;
-import okio.BufferedSource;
-import org.junit.Before;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
-
-@RunWith(MockitoJUnitRunner.class)
-public final class ConcurrencyLimitingInterceptorTest {
```
robert3005: These tests probably need to exist somewhere but not sure where should they live as that logic is in RemotingOkHttpCall.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -134,4 +135,119 @@ private static String limiterKey(Request request) {
...
+                timeoutCleanup.cancel(true);
+            }
+        }
+
+        private synchronized boolean timeoutScheduled() {
+            return timeoutCleanup != null && !timeoutCleanup.isDone() && !timeoutCleanup.isCancelled();
+        }
+
+        private synchronized void resetLimiter() {
+            log.warn("Timed out waiting to get permits for concurrency. In most cases this would indicate some kind of "
```
ellisjoe, line 174: can't this also happen when you
1. try acquire
2. enqueue
3. complete request
4. try acquire and again fail
5. hit timeout

ellisjoe (commented):
I'm generally still pretty opposed to this in it's current incantation due to the complexity, especially given how easy I find it to reason through the okhttp async code in the [Dispatcher](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Dispatcher.java) and [RealCall](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/RealCall.java).

I think one option is to change the dispatcher to take a generic request limiter, as it already has a notion of limiting the number of in flight request per host (https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Dispatcher.java#L175).

Alternatively I think there's likely a way we can avoid all of the callback madness while still not blocking the dispatcher threads. At the end of the day what we want to do is decouple scheduling from execution and I think we can do that with a less intrusive change.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
robert3005: Yes, this warning is imprecise and has the same semantics as the current one. Do you have suggestions for improvements?

robert3005:
I think I agree with you but I went for something implementable and that we're able to rollout. I don't see a significant difference between this and limiter in dispatcher, same logic at different level. @j-baker and I talked about this and it would have been ideal if you had a fully async api in okhttp instead of only the sync ones. I think the only hard thing to reason about here is what happens in cases where the limits are leaked - otherwise this seems roughly the same to me.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -134,4 +135,119 @@ private static String limiterKey(Request request) {
...
+                            + "deadlock. We expect that either this is caused by not closing response bodies "
+                            + "(there should be OkHttp log lines indicating this), or service overloading.",
+                    SafeArg.of("serviceClass", serviceClass),
+                    SafeArg.of("limiterKey", limiterKey),
+                    SafeArg.of("timeout", timeout));
+            leakSuspected.mark();
```
dansanduleac, line 180: can we tag this with the above args as well? (at least serviceClass and limiterKey)

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
dansanduleac: This seems to behave as before, except that you won't even enqueue if you couldn't acquire a permit.
Can you clarify the failure case you're seeing @ellisjoe?

ellisjoe: It's not that anything will fail, I'm just pointing out that the limiter will get reset under normal operation, as well as the exceptional case where the caller is not leaking responses

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
robert3005: I have a suspicion that there's a reason why it wasn't tagged with all the things before. @j-baker might remember

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpClient.java
```diff
@@ -64,4 +67,11 @@ RemotingOkHttpCall newCallWithMutableState(
                 executionExecutor,
                 maxNumRelocations);
     }
+
+    private Request addRateLimitIdTag(Request request) {
```
robert3005, line 73: we can only mutate request when creating so we had to move acquire to this level. This might have unintended consequence of acquiring limits too early in retrofit clients. We can put a mutable value on the tag so we can set it later but not sure if that's better

dansanduleac: In that case, let's put a SettableFuture here and initialize it in `enqueue` as discussed offline

robert3005: settable future it is

robert3005:
created #898 to remove part of this change

dansanduleac (approved):

PR #898, opened 2018-12-03
Upgrade concurrencylimiters to fix invalid blockinglimiter timeout

robert3005:
Upstream has merged our fix so we can get rid of the fork


dansanduleac:
👍

dansanduleac:
Merge ater adding the comment explaining why we need to truncate the instants to milli precision.

PR #900, opened 2018-12-04
[fix] Rename custom feign Encoders and Decoders to avoid clashing with remoting3

### LINKED ISSUE #810: checkClassUniqueness fails when remoting3 is also on the classpath
## What happened?

I tried to use both the shiny new conjure-java-runtime and the old http-remoting3 in the same project, but baseline's checkClassUniqueness task failed:

```
> Task :some-internal-product:checkClassUniqueness FAILED
4 Identically named classes with differing impls found in [com.palantir.conjure.java.runtime:conjure-java-jaxrs-client:4.0.0-rc6, com.palantir.remoting3:jaxrs-clients:3.43.0]: [feign.TextDelegateEncoder, feign.CborDelegateEncoder, feign.TextDelegateDecoder, feign.CborDelegateDecoder]

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':some-internal-product:checkClassUniqueness'.
> 'runtime' contains multiple copies of identically named classes - this may cause different runtime behaviour depending on classpath ordering.
  To resolve this, try excluding one of the following jars:
  
        (4 classes)   com.palantir.conjure.java.runtime:conjure-java-jaxrs-client:4.0.0-rc6 com.palantir.remoting3:jaxrs-clients:3.43.0        
```

Ideally, this just wouldn't break.  Otherwise, I think we need to offer a recommended approach for how to avoid this.

### LINKED ISSUE #899: feign decoders conflict with remoting3 classes
## What happened?

I'm playing classpath roulette with conjure-java-runtime and remoting3 feign encoders and decoders. We use the `feign` package to use package-private utilities from feign, but when we changed package names for conjure-java-runtime we did not update these packages.

conjure-java-runtime and remoting3 should provide different class names.

Proposal: Prefix class names

carterkozak:
Allows both to exist on the same classpath with reproducible
behavior.

fixes #899
fixes https://github.com/palantir/conjure-java-runtime/issues/810

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -27,21 +27,21 @@ import com.palantir.conjure.java.client.jaxrs.feignimpl.SlashEncodingContract;
 import com.palantir.conjure.java.okhttp.HostEventsSink;
 import com.palantir.conjure.java.okhttp.OkHttpClients;
 import com.palantir.logsafe.Preconditions;
-import feign.CborDelegateDecoder;
-import feign.CborDelegateEncoder;
+import feign.ConjureCborDelegateDecoder;
+import feign.ConjureCborDelegateEncoder;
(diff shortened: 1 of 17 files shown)


iamdanfox (approved):
LGTM - while these classes technically were accessible, I don't think we want to commit to them as public-API.

PR #903, opened 2018-12-05
[improvement] Log remote exception error information

pkoenig10:
## Before this PR
The errorName and errorInstanceId are not logged when mapping a `RemoteException`. Because we don't propagate the error, this information is lost.

## After this PR
The errorName and errorInstanceId are logged when mapping a `RemoteException`. This is consistent with the behavior of `ServiceExceptionMapper` and `JsonExceptionMapper`.


dansanduleac:
👍

PR #904, opened 2018-12-05
Propagate 403s

ellisjoe:
See https://github.com/palantir/conjure-java-runtime/issues/834


carterkozak:
If this merges, please update the conjure-undertow implementation as well:
https://github.com/palantir/conjure-java/blob/develop/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptionHandler.java

gregakinman (approved):

gregakinman:
@ellisjoe merge?

pnepywoda (approved):

ferozco:
👍

pnepywoda:
@ferozco would you be willing to propagate NOT_FOUND exceptions as well?

ferozco:
Whats the use case you had in mind? It would be surprising to have internal services hit 404s and want to propagate them up

pnepywoda:
We throw Not_found when objects or rids aren’t found. This isn’t about
wrong urls

PR #905, opened 2018-12-10
Disable automatic agent-based ALPN installation

markelliot:
## Before this PR

The jetty-http2-agent package provided a Java agent that would install https://github.com/jetty-project/jetty-alpn-agent, and an `Http2Agent.install()` method which allowed users to invoke that functionality themselves.

## After this PR

The `Http2Agent` class is now no-op.

We're removing this because it relies on very brittle logic that imposes a hidden constraint on the maximum JDK patch version, hasn't seen almost any adoption internally due to performance limitations in Java 8's Security stack, and there are better alternatives out there.

Callers should remove `Http2Agent.install()` invocations and may still obtain http2 based connections by using a different Java Security Provider such as [Conscrypt](https://github.com/google/conscrypt), or upgrading to Java 9+.


dansanduleac:
👍

dansanduleac:
@cakofony fair point! outputting the message along with a stacktrace to stderr instead

jetty-http2-agent/src/main/java/com/palantir/conjure/java/http2/Http2Agent.java
```diff
@@ -16,41 +16,33 @@
...
-        }
-
-        try {
-            AgentLoader.loadAgentClass(Http2Agent.class.getName(), "");
-            hasBeenInstalled = true;
-        } catch (Exception e) {
-            log.warn("Unable to dynamically install jetty-alpn-agent via ea-agent-loader, proceeding anyway...", e);
-        }
+        // no-op
+        new Exception("Http2Agent.install() no longer does anything, please remove all invocations from your codebase")
```
markelliot: @dansanduleac I think it's much safer to deprecate to no-op than deprecate and begin throwing exceptions/this is not a reasonable fallback behavior.

markelliot:
@cakofony I think deprecation warnings should help developers chase occurrences here/logging is likely unnecessary

jetty-http2-agent/src/main/java/com/palantir/conjure/java/http2/Http2Agent.java
carterkozak: This does not throw

carterkozak:
I agree, my point was that printing the same warning N times without additional context is not helpful.

jetty-http2-agent/src/main/java/com/palantir/conjure/java/http2/Http2Agent.java
markelliot: I see. given that we're going to produce deprecation warnings I don't think logging an exception here is necessary.

dansanduleac:
👍

jetty-http2-agent/src/main/java/com/palantir/conjure/java/http2/Http2Agent.java
iamdanfox: Realistically, these warnings are mostly only going to appear in gradle test output (as they're not structured JSON logs). Given that the gradle compilation output should already contain a deprecation warning I don't really want to start a precedent of adding exceptions every time we deprecate something.

iamdanfox (approved):

uschi2000:
Make sure this ends up in the release notes, please.

PR #908, opened 2018-12-11
Prevent npes if you don't get remotingokhttpclient

robert3005:
The error happens if you call newBuilder on RemotingOkhttpClient since then that will lose our override of newCall

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpClient.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpClient.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.java.okhttp;
 
 import com.google.common.util.concurrent.SettableFuture;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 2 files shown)


onursatici (approved):

ellisjoe:
Should we be overriding `newBuilder` in `RemotingOkhttpClient` then?

robert3005:
Even if you override newBuilder you can't override the builder itself

dansanduleac:
Should we log something when this happens? Sounds like it's indicative of the user doing something fishy.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpClient.java
```diff
@@ -57,6 +60,13 @@ public RemotingOkHttpCall newCall(Request request) {
         return newCallWithMutableState(addRateLimitIdTag(request), backoffStrategyFactory.get(), MAX_NUM_RELOCATIONS);
     }
 
+    @Override
+    public Builder newBuilder() {
+        log.warn("Attempting to copy RemotingOkHttpClient. Some of the functionality like rate limiting and qos will "
+                + "not be available to the new client");
```
dansanduleac: could we pass a `new SafeRuntimeException("stacktrace")` to this as well, so we know where this was called and track it?

dansanduleac:
👍

PR #909, opened 2018-12-12
[break] Exclude empty optionals during serialization

### LINKED ISSUE #398: Jackson ObjectMapper config changes
I propose we:
 - if a field is Optional.absent/Optional.empty we omit the field vs. explicit null
 - if a field is an empty map, set or list we omit the field vs. explicit empty

pkoenig10:
## Before this PR
Empty optional fields are serialized as `null`.

## After this PR
Empty optional fields are omitted during serialization

---

Fixes #398

~This change should be safe. The only test that fails without the object mapper changes is `serializeJdk8ModuleEmptyOptionalField`, which is expected. Object mappers created both before and after this change can correctly deserialize all forms (present, null, absent).~

I don't think it is possible to omit empty collections without requiring code changes in objects themselves. Many objects require that collection fields are present and don't provide a default value. This includes the default configuration of all immutables objects, which are used extensively with these object mappers.

We can still make this change for conjure objects by adding the relevant `@JsonInclude` annotation.

---

EDIT: This actually would be a break for clients mapping objects that do not provide a default value for optional objects. Both conjure and immutables objects do, so this would only affect objects with hand-rolled jackson annotations. But that is still a break.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ bin/
 .idea/
 out/
 generated_src/
+generated_testSrc/
 
 # Codegen
 .generated
--- a/conjure-java-jackson-serialization/build.gradle
+++ b/conjure-java-jackson-serialization/build.gradle
@@ -1,3 +1,4 @@
+apply plugin: "org.inferred.processors"
 apply from: "${rootDir}/gradle/publish-jar.gradle"
 
 dependencies {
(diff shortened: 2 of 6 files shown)


iamdanfox (changes requested):
While this does look a lot prettier, I don't think it's really safe to do it just yet because typescript clients may well be expecting a JSON response containing an explicit `null` or `[]` (both are distinct from the concept of a missing field, which is the javascript value `undefined`).

I think we really need a serde layer in conjure-typescript before I'd be confident that this will not cause widespread frontend disruption.

PR #916, opened 2019-01-09
[improvement] Remove unnencessary complexity passing trace operations to okhttp

carterkozak:
Unnecessary for us to un-escape trace operation names for every client request

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/PathTemplateHeaderEnrichmentContract.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/PathTemplateHeaderEnrichmentContract.java
@@ -10,7 +10,9 @@ import feign.MethodMetadata;
 import java.lang.reflect.Method;
 
 public final class PathTemplateHeaderEnrichmentContract extends AbstractDelegatingContract {
+    @Deprecated
     public static final char OPEN_BRACE_REPLACEMENT = '\0';
+    @Deprecated
     public static final char CLOSE_BRACE_REPLACEMENT = '\1';
 
     public PathTemplateHeaderEnrichmentContract(Contract delegate) {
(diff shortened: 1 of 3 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/PathTemplateHeaderEnrichmentContract.java
```diff
@@ -19,10 +29,11 @@ public PathTemplateHeaderEnrichmentContract(Contract delegate) {
...
+
         metadata.template().header(OkhttpTraceInterceptor.PATH_TEMPLATE_HEADER,
-                metadata.template().method() + " "
-                        + metadata.template().url()
-                            .replace('{', OPEN_BRACE_REPLACEMENT)
-                            .replace('}', CLOSE_BRACE_REPLACEMENT));
+                metadata.template().method() + " " + metadata.template().url()
+                        // escape from feign string interpolation
+                        // See RequestTemplate.expand
+                        .replace("{", "{{"));
```
dansanduleac, line 37: do we have any tests to verify that this doesn't break anything?

dansanduleac (commented):
Why are we making this change? Have we established that this is causing slowness somewhere?
I don't have much context on this bit of code so would require a bit more digging to make sure thsi doesn't break any expectations.

carterkozak:
> Why are we making this change?

We've got multiple projects which implement exactly this functionality two different ways, I'd like to standardize on one of them, so I opted for the simpler solution.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/PathTemplateHeaderEnrichmentContract.java
carterkozak: Yes, this is tested by com.palantir.conjure.java.client.jaxrs.TracerTest

iamdanfox: :+1: I checked this out locally and verified that without that `.replace("{", "{{"));` line `com.palantir.conjure.java.client.jaxrs.TracerTest` fails!

iamdanfox (approved):

PR #917, opened 2019-01-09
[improvement] Don't ACCEPT_FLOAT_AS_INT

dansanduleac:
## Before this PR

Integer fields could be deserialized from floating point values e.g. `123.0` -> `123`.

## After this PR

Floating point JSON values are no longer allowed when an integer is expected.


ferozco:
👍

iamdanfox:
Since this touches object mappers, can we make sure to put a very clear explanation of the implications in the release?

PR #924, opened 2019-01-11
[improvement] Lower read/write timeouts to 5 minutes

ellisjoe:
## Before this PR
Generally requests that take >5 minutes are highly likely to be taking so long due to talking to a bad server rather than actually requiring that much time.

## After this PR
We timeout in half the time which means we now expect to get a successful response in half the time.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -38,8 +38,8 @@ public final class ClientConfigurations {
 
     // Defaults for parameters that are optional in ServiceConfiguration.
     private static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
-    private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(10);
-    private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(10);
+    private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(5);
+    private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(5);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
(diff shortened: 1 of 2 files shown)


carterkozak:
I'm curious about specific failures that motivated this change. Have we observed server errors which result in clients hanging for 10 minutes?

carterkozak:
More concretely, are we confident we're not seeing something like https://github.com/palantir/conjure-java-runtime/issues/925?

dansanduleac:
How did we reach the conclusion that `5 minutes` is what the majority of successful requests take?

iamdanfox:
From in-person chat with @ellisjoe:

1. The existing 10 minute timeout does not apply to streaming uploads or downloads, so lowering to 5 minutes won't affect any huge uploads or downloads.
2. The existing 10 minute timeout isn't our long-term desired state for any user-initiated actions (aka RPCs), ideally this would be O(seconds) by default. Cutting this in half is an incremental way to get closer to O(seconds) by default, with the goal of doing further cuts in future.
3. This _is_ intended to be a forcing function for service design, where very long calls should be turned into an asynchronous job that has an ID and then is polled for results, or a truly streaming thing (with the goal of giving the user better progress information).  If the 10min -> 5min change causes short-term problems at deployments, it can be configured in YAML back to 10 minutes as a stopgap fix.

markelliot:
I think we might also (after this change) add an annotation which allows the limit to be mutated upwards for particular endpoints so that we can think of this as the global default.

carterkozak:
I'm mostly aligned with this, but would like to clarify: Are we suggesting immediate response polling over long-polling? Connections are incredibly cheap, threads are not, and we should be careful not to confuse the two.

ellisjoe:
Don't think we're against long polling, but do think we'd want to treat that as the special case rather than the default. As mark mentioned, you could imagine we allow you to explicitly annotate the methods which are expected to take longer than the default timeout period.

carterkozak:
Yep, I'm in agreement, I think Mark and I posted concurrently.

ellisjoe:
Ah great! Good with merging this then?

carterkozak (approved):
Go for it :-)

PR #929, opened 2019-01-21
[fix] Downgrade to retrofit 2.4.0

dansanduleac:
## Before this PR

#871  updated the version of retrofit2 we were using, but this introduced a regression in path parameter parsing.

## After this PR

Revert back to retrofit2 2.4.0. This is a workaround for https://github.com/palantir/conjure-java-runtime/issues/930


versions.props
```diff
@@ -14,7 +14,9 @@ com.palantir.safe-logging:* = 1.7.0
 com.palantir.tracing:* = 2.1.0
 com.palantir.tritium:tritium-registry = 0.11.2
 com.squareup.okhttp3:* = 3.12.0
-com.squareup.retrofit2:* = 2.5.0
+# dependency-upgrader:OFF
```
iamdanfox, line 17: Link to https://github.com/palantir/conjure-java-runtime/issues/930 pls?

carterkozak:
lgtm, thanks!

iamdanfox (approved):

PR #939, opened 2019-02-04
[improvement] enqueue for concurrency limits in round robin fifo not fifo

j-baker:
Fifo allows for serious starvation issues; i.e. if a single thread
enqueues 50k requests, it'll be scheduled as the next 50k requests to be
sent; a thread which enqueues in the meantime will have to wait.

This PR means that we will round robin between threads, serving the
requests in fifo order.

For feign clients this will change absolutely nothing, since they are
synchronous. For retrofit clients (of which we have maybe 2) this will
dramatically improve _fairness_; a background task spamming requests
will not be able to totally starve the limiters.


robert3005 (approved):
LGTM - one thing that was confusing was when does a thread get removed from activeQueuedThreads. Would add one more test to verify that for given thread id queued -> empty -> queued behaves correctly

ferozco (approved):

PR #940, opened 2019-02-05
[improvement] okhttp 3.12.0 -> 3.13.1, to pick up perf improvements to okhttp3.Dispatcher

iamdanfox:
## Before this PR

We recently experienced two P0 issues that seemed to be caused by locking inside the okhttp 3.12.0 dispatcher.  I opened a PR to okhttp which should hopefully reduce the amount of work done inside `synchronized` blocks (https://github.com/square/okhttp/pull/4581).

One internal service forced okhttp to 3.11.0 as a temporary workaround.

## After this PR

Hopefully that 3.11.0 downgrade can be reverted, and P0s won't re-occur!

_Full changelog https://github.com/square/okhttp/blob/master/CHANGELOG.md_

Do not merge until:

- [x] ~TLSv1 and TLSv1.1 are no longer enabled by default. - do we need to opt-in to `ConnectionSpec.COMPATIBLE_TLS`~ seems like we've always been overriding the connection spec to [only specify TLS 1.2](https://github.com/palantir/conjure-java-runtime/blob/9d88bfa3c3e1de74ea57f158734186628d1bb8c9/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java#L225).


dansanduleac:
:+1:

PR #941, opened 2019-02-07
[improvement] Improve exceptions and logging when retrying

pkoenig10:
## Before this PR
- Request failures were logged at the `debug` level
- Nothing was logged when handling a `RetryOther` QoS response
- The message "Rescheduling call after backoff" was logged when retrying requests for different reasons
- `IOException` was used to indicate failures

## After this PR
- Request failures are logged at the `info` level
- Log when handling any QoS response
- Use unique, descriptive messages when retrying requests
- Use `SafeIoException` to indicate failures

Logging request failures at the info level is the primary motivation of this PR. Without this there is no indication when requests fail.

cc @pnepywoda

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -26,6 +26,8 @@ import com.palantir.conjure.java.api.errors.QosException;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.client.config.ClientConfiguration;
 import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
+import com.palantir.logsafe.exceptions.SafeIoException;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


pnepywoda (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -175,24 +177,31 @@ public void onFailure(Call call, IOException exception) {
...
-                            + request().url() + "' and base URLs " + urls.getBaseUrls()));
+                    callback.onFailure(call, new SafeIoException(
+                            "Failed to determine valid failover URL",
+                            exception,
+                            UnsafeArg.of("requestUrl", call.request().url().toString()),
+                            UnsafeArg.of("baseUrls", urls.getBaseUrls())));
                     return;
                 }
 
+                log.info("Retrying call after failure",
```
iamdanfox, line 197: @pkoenig10 why do we think all services want to see this log line all the time? If you're interested in something specific at one service, why not dial up debug logging for that service?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -273,50 +285,65 @@ public Void visit(QosException.Throttle exception) {
...
+                    callback.onFailure(call, new SafeIoException(
+                            "Failed to determine valid redirect URL after receiving QoS retry other",
+                            exception,
+                            UnsafeArg.of("requestUrl", call.request().url().toString()),
+                            UnsafeArg.of("redirectToUrl", exception.getRedirectTo().toString()),
+                            UnsafeArg.of("baseUrls", urls.getBaseUrls())));
+                    return null;
                 }
+
+                log.debug("Retrying call after receiving retry other",
```
iamdanfox: Might as well use the precise class name here? s/retry other/QosException.RetryOther`?

iamdanfox:
Big fan of all the IOException -> SafeIoException fixes, thanks @pkoenig10! LGTM pending the debug->info question

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
pkoenig10: Sure, I was trying to use class name agnostic messages but happy to change if you prefer using the class name

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
pkoenig10: When requests fail there is no indication in any service or request logs of the failure. When debugging these kind of failures, it can appear that the request was not retried the correct number of times or that the request was retried on the same host.

The majority of debugging is done after the fact, so debug logs are often not very useful.

When we receive QoS responses, there is at least a request log in the remote service. So those logs probably should stay debug.

Also QoS responses are relatively common and expected while failures are generally rare and unexpected. So it seems acceptable to log when they occur.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -175,24 +177,31 @@ public void onFailure(Call call, IOException exception) {
...
+                    callback.onFailure(call, new SafeIoException(
+                            "Failed to determine valid failover URL",
+                            exception,
+                            UnsafeArg.of("requestUrl", call.request().url().toString()),
+                            UnsafeArg.of("baseUrls", urls.getBaseUrls())));
                     return;
                 }
 
+                log.info("Retrying call after failure",
+                        SafeArg.of("backoffMillis", backoff.get().toMillis()),
```
dansanduleac, line 198: we should log the `redirectTo.get()` url as well, so it's obvious what server it will try nextt

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -273,50 +285,65 @@ public Void visit(QosException.Throttle exception) {
...
                 }
+
+                log.debug("Retrying call after receiving QosException.Unavailable",
+                        SafeArg.of("backoffMillis", backoff.get().toMillis()),
+                        exception);
+                Request redirectedRequest = request().newBuilder()
+                        .url(redirectTo.get())
+                        .build();
+                scheduleExecution(() -> client.newCallWithMutableState(redirectedRequest, backoffStrategy,
+                        maxNumRelocations).enqueue(callback), backoff.get());
```
dansanduleac: I know you just copied code but do you mind making the method chaining more readable?
The formatting is super weird across this whole repo sadly

ferozco:
👍

PR #952, opened 2019-02-14
[improvement] Decode null response body with text/plain as empty string

saurabhshukla92:
## Before this PR
Conjure JaxRS client throws exception if the response with type text/plain has null body. 

## After this PR
Response with type text/plain with null body is now decoded as empty string.


dansanduleac:
Talked offline, but to summarize:

This is dealing with endpoints returning String where you get a 204 with `content-type: text/plain`, which is not a kind of endpoint conjure would generate.
While we want to support this for historical reasons, I think it might be surprising to users expecing a String to get an empty string when the server returned null.

Instead, we should pass the null through, and make the client acknowledge / specially handle this case, rather than let the client believe that an empty string was returned.

I've opened https://github.com/palantir/conjure-java-runtime/pull/953 to try to implement that.

iamdanfox (approved):
I think this looks fine to me - empty string seems a perfectly reasonable way to represent a response with no body.

iamdanfox:
If time allows, it would be awesome to see if we can capture the same behaviour in a retrofit test too (just to make sure things don't diverge), but if there's time pressure on this then no worries.

dansanduleac:
👍 given that feign handles even a 200 with `content-length: 0` as body=null (a tad unexpected!), I think in the end this approach is less surprising so let's go with this

PR #953, opened 2019-02-14
[fix] Move ConjureTextDelegateDecoder first, allowing it to return null

dansanduleac:
## Before this PR

conjure-java-runtime currently has some special handling for endpoints marked with `@Produces(MediaType.TEXT_PLAIN)` that return String.
This is a kind of endpoint that is not generated by conjure, but supported by the feign client for historical reasons.

Such an endpoint might respond with a 204 (NO CONTENT), but that would currently throw because the `ConjureNeverReturnNullDecoder` sits at the top of the decoder stack.

## After this PR

Don't apply the `ConjureNeverReturnNullDecoder` to such endpoints as described above.

These are not endpoints that conjure would generate, so it seems acceptable to relax the null safety on them.

We still want the null safety to apply to endpoints that could be generated by conjure, since we know that conjure servers should not permit unexpected null/empty responses, but not to other kinds of endpoints that the feign client supports for historical reasons.

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -119,17 +119,17 @@ abstract class AbstractFeignJaxRsClientBuilder {
     }
 
     private static Decoder createDecoder(ObjectMapper objectMapper, ObjectMapper cborObjectMapper) {
-        // The ConjureTextDelegateDecoder is allowed to return null
-        // This is to support older endpoints that are not conjure-generated
-        return new ConjureTextDelegateDecoder(
-                new ConjureNeverReturnNullDecoder(
-                new ConjureJava8OptionalAwareDecoder(
-                        new ConjureGuavaOptionalAwareDecoder(
-                                new ConjureEmptyContainerDecoder(
-                                        objectMapper,
-                                        new ConjureInputStreamDelegateDecoder(
-                                                new ConjureCborDelegateDecoder(
-                                                        cborObjectMapper,
-                                                        new JacksonDecoder(objectMapper))))))));
+        return new ConjureJava8OptionalAwareDecoder(
+                new ConjureGuavaOptionalAwareDecoder(
+                        // The ConjureTextDelegateDecoder is allowed to return null
+                        // This is to support older endpoints that are not conjure-generated
+                        new ConjureTextDelegateDecoder(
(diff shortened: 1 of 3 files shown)


saurabhshukla92 (approved):

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -119,13 +119,15 @@ private Contract createContract() {
...
-                                new ConjureEmptyContainerDecoder(
-                                        objectMapper,
-                                        new ConjureInputStreamDelegateDecoder(
-                                                new ConjureTextDelegateDecoder(
+        return new ConjureJava8OptionalAwareDecoder(
+                new ConjureGuavaOptionalAwareDecoder(
+                        // The ConjureTextDelegateDecoder is allowed to return null
+                        // This is to support older endpoints that are not conjure-generated
+                        new ConjureTextDelegateDecoder(
+                                new ConjureNeverReturnNullDecoder(
```
dansanduleac, line 127: I swapped the order round this way because:

* optional decoders need to first check if a 204 is compatible with an optional type
* conjure text delegate decoder _only_ handles 204 + text/plain + **return type you can assign string to** (String, CharSequence, Object) -- there are tests for these. the return type check has been added in this PR
* never null decoder handles all other unexpected nulls coming from jackson

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -119,13 +119,15 @@ private Contract createContract() {
     }
 
     private static Decoder createDecoder(ObjectMapper objectMapper, ObjectMapper cborObjectMapper) {
-        return new ConjureNeverReturnNullDecoder(
```
iamdanfox, line 122: I really don't want to move this because I think it's a really nice guarantee to say the conjure-java-runtime clients _can never return null_.  Part of our whole story about conjure is that you shouldn't have to think about the concept of `null` - we emphasise optionals instead.  I don't think we should reintroduce edgecases here.

iamdanfox:
Not really a fan of this approach - would prefer to run with https://github.com/palantir/conjure-java-runtime/pull/952

dansanduleac:
Yeah, problem is.. I made the assumption that if you return a 200 with an empty string (`content-length: 0`) feign would give you a empty string for the body, but instead it gives you null. In that case, #952 is a less suprising behaviour. Closing.

PR #960, opened 2019-02-27
Only select a URL once per request attempt

### LINKED ISSUE #942: Round robin selection strategy will advance URL selector twice on retry
## What happened?

When a failure happens in a client using the round robin node selection strategy, the URL is advanced twice.

The first time will happen in the request callback, either in the [failure handler](https://github.com/palantir/conjure-java-runtime/blob/4.14.0/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java#L182) or the [QoS unavailable handler](https://github.com/palantir/conjure-java-runtime/blob/4.14.0/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java#L308).

The second time will happen [in the round robin interceptor](
https://github.com/palantir/conjure-java-runtime/blob/4.14.0/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RoundRobinUrlInterceptor.java#L44) for the retried request.

If the number of urls is odd then this behavior is acceptable. However if the number of urls is even then we may try a second time on some hosts before we've tried once on other hosts.

When retrying a request with a round robin selection strategy we should only advance the URL selector once.

pkoenig10:
Fixes #942

## Before this PR
When using the round robin node selection strategy, retried requests would advance the current URL twice - once in the response handler and again in the URL selector interceptor.

## After this PR
Retried requests will only advance the current URL once, as expected. This PR also simplifies the URL selection by removing the URL selector interceptor and choosing the correct URL when a new request is first attempted.

The URL selector interceptor seems broken in a number of other ways. For example, the current URL can change between when a redirect URL is selected on failure and when the retried request passes through the interceptor, causing a different URL to be used than the one selected in the failure handler.


iamdanfox (approved):
Looks very sensible, thanks @pkoenig10!

PR #962, opened 2019-03-01
[improvement] add bean validation + test that we handle `@NotNull` request bodies appropriately

bavardage:
## Before this PR
In some places where we take a dependency on conjure-java-jersey-server we layer resource classes. Some of these layers make assumptions that the request body parameters are not null (and fail if this is not the case).

## After this PR
By adding jersey-bean-validation we ensure that request handler methods annotating their parameters as `@NotNull` do not get called with null parameters. Instead the request is rejected at the jersey layer.


carterkozak:
I like this. Thoughts @iamdanfox @dansanduleac @ferozco?

carterkozak (approved):

iamdanfox:
Looks great, thanks @bavardage!

PR #966, opened 2019-03-04
[improvement] Lower read and write timeouts

ellisjoe:
## Before this PR
Single faulty nodes continue to cause consumers to grind to halt as they wait for requests to timeout. 

## After this PR
Lowering this again means we are more quickly able to detect faulty nodes and recover by failing over to healthy nodes.

Previous PR lowering timeouts: https://github.com/palantir/conjure-java-runtime/pull/924

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -38,8 +38,8 @@ public final class ClientConfigurations {
 
     // Defaults for parameters that are optional in ServiceConfiguration.
     private static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(10);
-    private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(5);
-    private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(5);
+    private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(2);
+    private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(2);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
     private static final Duration DEFAULT_FAILED_URL_COOLDOWN = Duration.ZERO;
     private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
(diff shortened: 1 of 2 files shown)


iamdanfox:
Looking at aggregate request logs, there are _a lot_ of `duration:>120` requests today.  While the [previous PR](https://github.com/palantir/conjure-java-runtime/pull/924) did have the desired effect (requests started failing and @saurabhshukla92 dialled up the timeout as a stop-gap fix and then worked on parallelization as a longer-term fix), we received feedback that it wasn't well enough communicated.

I think teams that are already tracking SLOs won't protest at these shorter timeouts, but other teams that perhaps _do_ rely on >2 minute requests working will likely be angry.  Before merging, let's try to:

1. Identify which services will be affected
1. Send comms suggesting they should be defining SLOs as a first priority
1. Suggest recommended approaches to refactor RPC calls that _can't_ complete within seconds

carterkozak:
Also note that this will affect services which _do not_ report aggregate statistics to shared infrastructure. Will you be available to help individual teams work through timeouts on 2-5 minute requests which previously succeeded?

Alexis-D:
An alternative proposal would be to make conjure-java-runtime ship w/o defaults, and require clients to explicitly define their own timeouts. Thoughts?

iamdanfox:
@ellisjoe what do you think of this alternative approach that might tick some of the same outcomes?

## Server-defined adaptive timeouts

- On the server-side, resourced annotated with `@ServiceLevelObjective(p95 = "200ms", p99 = "220ms", p999 = "240ms")`  would happen to include this information in a _Response header_, e.g. `SLO: {p95: 200ms, p99: 220ms, p999: 240ms}`
- On the client-side, any client with no explicitly-configured timeouts would detect these headers and update their internal representation of the SLOs advertised by each node. Clients with _explicit_ timeouts set in YAML would just use these and ignore the server-advertised SLO, as an escape hatch for dev teams.

A nice property is that dev teams which experience pain caused by a slow upstream service would encourage the upstream service to define SLOs.  Then, if the SLOs are met, then the client will get nice fast behaviour.  If the SLOs are not met, then the team responsible for the upstream service will be alerted.

ellisjoe:
So I think at this point any change here is likely to introduce more pain than anything, and I also don't see fixing timeouts as the things that's going to particularly move the needle in terms of our client's resiliency. I'd propose we hold off on changes here and keep track of situations where a lower timeout would have helped us avoid slowdowns more quickly.

Few thoughts we've discussed off thread:
- SLO based limits on a per endpoint basis seem reasonable. We'd need to figure out a way for service owners to define and expose those to consumers, which conjure doesn't support today
- Another thing we could do is try to learn what a reasonable timeout is, in a similar way to how we learn the concurrency limit: as requests succeed we lower the timeout, as they fail we increase it

iamdanfox:
Closing out as we're not going to merge this PR in this current state - I think @Alexis-D has a "conjure-java-runtime's timeouts" quip going for further discussion.

PR #967, opened 2019-03-05
Switch to gradle-consistent-versions

dansanduleac:
## Before this PR

We used the nebula lock plugin to source all our versions - this had a few problems:

- very verbose PRs
- excavator PRs didn't run `./gradlew generateLock saveLock`
- `./gradlew idea` could be very slow

## After this PR

We use the new https://github.com/palantir/gradle-consistent-versions plugin instead.  The task to remember is: **`./gradlew --write-locks`**


okhttp-clients/versions.lock
```diff
@@ -1,432 +0,0 @@
...
-            "locked": "2.1.0"
-        },
-        "com.palantir.tritium:tritium-registry": {
-            "locked": "0.11.2"
-        },
-        "com.squareup.okhttp3:logging-interceptor": {
-            "locked": "3.12.0"
-        },
-        "com.squareup.okhttp3:okhttp": {
-            "locked": "3.12.0",
```
iamdanfox, line 167: Validated this version is unchanged :+1:

build.gradle
```diff
@@ -11,16 +11,20 @@ buildscript {
...
 
 apply plugin: 'com.palantir.baseline'
 apply plugin: 'com.palantir.git-version'
+apply plugin: 'com.palantir.consistent-versions'
 
 allprojects {
     version gitVersion()
     group 'com.palantir.conjure.java.runtime'
 
+    apply plugin: 'nebula.dependency-recommender' // necessary only to construct the BOM
```
iamdanfox: As a follow-up, we might be able to delete the bom publishing entirely as I think it's unused!

iamdanfox:
:+1:

PR #969, opened 2019-03-06
Gradle dependencies are defensive against known issues

iamdanfox:
## Before this PR

Robots might come round and try to upgrade our dependencies, but we know that we specifically don't want to take a couple of upgrades.

## After this PR

Gradle will fail builds if you try to upgrade outside our known safe ranges of retrofit and dropwizard metrics


iamdanfox:
@dansanduleac I think we're gonna have to be smarter here because all our tests rely on the new metrics jar!!

dansanduleac (approved):

PR #973, opened 2019-03-10
[break][improvement] Deserialize List, Map, and Set to guava immutable implementations

carterkozak:
Guava collections provide many of the same benefits that we encode into our generated beans, which were previously not applied to conjure objects outside of bean values: immutability, and values contained in collections may not be `null`.

### DIFF
--- a/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
+++ b/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
@@ -43,6 +43,7 @@ client:
    - 'null'
    receiveListAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
    receiveSetBearerTokenAliasExample:
    - 'null'
    receiveSetBinaryAliasExample:
@@ -65,13 +66,11 @@ client:
    - 'null'
    receiveSetAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
 
    receiveMapBinaryAliasExample:
    - '{}'
    - '{"SGVsbG8sIFdvcmxk": true}'
-   receiveMapDoubleAliasExample:
-   - '{"10": true, "10e0": false}'
-   - '{"10": true, "10.0": false}'
 
   singleHeaderService: {}
 
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ImmutableCollectionsModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ImmutableCollectionsModule.java
@@ -1,0 +1,40 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Values deserialized into {@link List}, {@link Set}, and {@link Map}, use Guava implementations of
+ * {@link ImmutableList}, {@link ImmutableSet}, and {@link ImmutableMap} respectively.
+ */
+final class ImmutableCollectionsModule extends SimpleModule {
+    private static final long serialVersionUID = 1L;
+
+    ImmutableCollectionsModule() {
+        super(ImmutableCollectionsModule.class.getCanonicalName());
+        addAbstractTypeMapping(List.class, ImmutableList.class);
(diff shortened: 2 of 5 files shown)


carterkozak:
Labeled as a break because it's possible folks rely on null values or modifying root level collections.

rubenfiszel:
Tests are a bit too specific around the mapper. We probably want to spec that Encoding should return immutable collections and parametrize the tests around picking an Encoding.

[edit: nvm the above, wrong repo]

conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
```diff
@@ -43,6 +43,7 @@ client:
    - 'null'
    receiveListAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
```
carterkozak, line 46: Fix: https://github.com/palantir/conjure-verification/pull/141

rubenfiszel (approved):

conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
dansanduleac: Surprised this was working before.. was it just deserializing the `null` to `Optional.empty()`?

carterkozak: Nope, `Alias<X>` isn't any different from deserializing a raw `X`, so it used an `ArrayList<Object>` which does not do any validation, and ends up with `null` values.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ImmutableCollectionsModule.java
```diff
@@ -0,0 +1,40 @@
...
+final class ImmutableCollectionsModule extends SimpleModule {
+    private static final long serialVersionUID = 1L;
+
+    ImmutableCollectionsModule() {
+        super(ImmutableCollectionsModule.class.getCanonicalName());
+        addAbstractTypeMapping(List.class, ImmutableList.class);
+        addAbstractTypeMapping(Map.class, ImmutableMap.class);
+        addAbstractTypeMapping(Set.class, ImmutableSet.class);
+    }
+}
```
dansanduleac, line 40: This seems like a different change that should be done in a different PR.
Also, @iamdanfox raises a good point that if we want to standardize on these collections, why not make the change in conjure-java?
That way, users could also benefit from the added type-safety (it's obvious that these values are immutable, so never necessary to copy them).

conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
dansanduleac: Argh.. I really dislike this magic inversion of control inherent in reflection-based serde.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ImmutableCollectionsModule.java
carterkozak: using interface types here gives us the freedom to move to another implementation without breaking all conjure consumers.
If we update conjure-java instead, we will break every generated service which consumes or return a list/set/map because the method signature will change.

carterkozak: At the moment conjure-undertow generated code is the only place where our generated code depends on guava, and that's for a single `ImmutableList.of()` that I'm considering removing for simplicity.

dansanduleac: We can keep returning List etc in the interface then, and only use immutables in the implementation

carterkozak: Then we'd need this change anyhow, otherwise we'd face the same problem for wrapper types used as the request/response body.

conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
```diff
@@ -65,13 +66,11 @@ client:
...
    receiveSetAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
 
    receiveMapBinaryAliasExample:
    - '{}'
    - '{"SGVsbG8sIFdvcmxk": true}'
-   receiveMapDoubleAliasExample:
-   - '{"10": true, "10e0": false}'
-   - '{"10": true, "10.0": false}'
```
iamdanfox, line 74: Sadly, I don't think we can make a change like this without a major rev of conjure-java-runtime (which has historically been _super_ expensive). I think people will get super angry if we impose this change on them, because if they have odd looking JSON already persisted in their database they would suddenly be read it out again.

I think the way to make progress in this direction would be to allow users to opt-in to stricter inner fields in our builders:

```diff
   @Generated("com.palantir.conjure.java.types.BeanBuilderGenerator")
    @JsonIgnoreProperties(ignoreUnknown = true)
    public static final class Builder {
-       private List<String> value = new ArrayList<>();
+       private ImmutableList.Builder<String> value = ImmutableList.builder();

        private Builder() {}
```

This itself is a bit contentious, as we have previously tried to make sure our conjure objects don't rely on guava at all.  Maybe long-term we could let people use the JDK-native `List` https://docs.oracle.com/javase/9/docs/api/java/util/List.html#immutable?

iamdanfox (changes requested):
I want to be really cautious about changing the behaviour of our object mappers - either we've got to make a whole new object mapper, or come up with some way that people can responsibly opt-in to this and not have it imposed on them.

carterkozak:
I've considered retrofitting this change onto the mappers we use for the conjure-undertow runtime, since it requires you to opt into using the new framework, however that runs slightly counter to the goal of providing a drop-in replacement. Thoughts?

iamdanfox:
@cakofony I actually quite like that idea, because the whole story we're pitching with the conjure-generated undertow handlers is that they're more precise and more performant... when people opt-in to them, I think it's pretty reasonable to ask them to confront any previous imprecision they were relying on, and crucially they can do this at their leisure because the old jersey approach keeps working.

carterkozak:
Great, I'll open a PR there

PR #974, opened 2019-03-10
[don't merge without broader conversation] Fix scalar coerscion tests by disabling afterburner

carterkozak:
MapperFeature.ALLOW_COERCION_OF_SCALARS _will_ save us
...if we disable afterburner. We can fix the afterburner bug,
however I've been unable to produce a benchmark showing clear
wins using the afterburner module.

### DIFF
--- a/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
+++ b/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
@@ -1,14 +1,7 @@
 client:
   autoDeserialize:
-
-   receiveBooleanExample:
-   - '{"value":0}' # jackson is casting 0 -> false and 1 -> true... MapperFeature.ALLOW_COERCION_OF_SCALARS);) in 2.9 will save us
-   - '{"value":"true"}' # jackson is casting 0 -> false and 1 -> true... MapperFeature.ALLOW_COERCION_OF_SCALARS);) in 2.9 will save us
-
    receiveStringExample:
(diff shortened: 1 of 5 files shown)


dansanduleac:
Well, that's one way to do it!
I've already fixed this in afterburner btw (https://github.com/FasterXML/jackson-modules-base/pull/70), but it won't be in until 2.9.9 which seems to be taking a while to release.

carterkozak:
Oh fun, would you mind adding that link to the ignore line?

carterkozak:
Closing given https://github.com/FasterXML/jackson-modules-base/pull/70

conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
```diff
@@ -1,14 +1,7 @@
 client:
   autoDeserialize:
-
-   receiveBooleanExample:
-   - '{"value":0}' # jackson is casting 0 -> false and 1 -> true... MapperFeature.ALLOW_COERCION_OF_SCALARS);) in 2.9 will save us
-   - '{"value":"true"}' # jackson is casting 0 -> false and 1 -> true... MapperFeature.ALLOW_COERCION_OF_SCALARS);) in 2.9 will save us
-
    receiveStringExample:
    - '{"value":8}' # jackson coerces things to other types
```
dansanduleac, line 4: it's interesting that this is still allowed even with ALLOW_COERCION_OF_SCALARS on

PR #978, opened 2019-03-12
Fix comments re: jackson coercion of scalars

dansanduleac:
## Before this PR

Ignore file was mentioning the wrong jackson flag - we already use ALLOW_COERCION_OF_SCALARS, but it doesn't work with afterburner. See https://github.com/FasterXML/jackson-modules-base/pull/70

## After this PR

Fixed comments to more accurately describe the ignores that will be fixed by jackson 2.9.9


iamdanfox:
:+1:

PR #979, opened 2019-03-12
[improvement] log.warn if deserialized raw collections contain null

carterkozak:
Tests will fail, however in production where assertions are not
enabled, warnings will be logged and functionality is unmodified.

### DIFF
--- a/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
+++ b/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
@@ -75,9 +75,6 @@ client:
    receiveMapBinaryAliasExample:
    - '{}'
    - '{"SGVsbG8sIFdvcmxk": true}'
-   receiveMapDoubleAliasExample:
-   - '{"10": true, "10e0": false}'
-   - '{"10": true, "10.0": false}'
 
   singleHeaderService: {}
 
--- a/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
+++ b/conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
@@ -43,6 +43,7 @@ client:
    - 'null'
    receiveListAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
    receiveSetBearerTokenAliasExample:
    - 'null'
    receiveSetBinaryAliasExample:
@@ -65,6 +66,7 @@ client:
    - 'null'
    receiveSetAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
 
    receiveEnumExample:
    - '"!!!"'
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
@@ -137,14 +137,19 @@ final class CollectionValidationModule extends SimpleModule {
 
     @Nullable
     private static <T> T logIfNull(@Nullable T value) {
-        // Log noisily in production
-        if (value == null
-                // Avoid the performance cost of Throwable.fillInStackTrace if WARN has been disabled
-                && log.isWarnEnabled()) {
+        if (value == null) {
+            onNull();
+        }
+        return value;
+    }
+
+    private static void onNull() {
+        // Avoid the performance cost of Throwable.fillInStackTrace if WARN has been disabled
+        if (log.isWarnEnabled()) {
+            // Log noisily in production
             log.warn("Detected a null value, null values are not allowed by Conjure", new IllegalArgumentException());
         }
         // Tests should fail, but we're not ready to cause breaks in deployed systems yet.
-        assert value != null : "Null values are not allowed by Conjure";
-        return value;
+        assert false : "Null values are not allowed by Conjure";
     }
 }
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
@@ -1,0 +1,128 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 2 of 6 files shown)


conjure-java-client-verifier/src/test/resources/ignored-test-cases.yml
```diff
@@ -43,6 +43,7 @@ client:
    - 'null'
    receiveListAnyAliasExample:
    - 'null'
+   - '[null, 0, "content", true, [1,2,3], {"key":3}]' # null is not a valid any value, see conjure-verification#141
```
carterkozak, line 46: This only applies when asserts are enabled

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Nullable
+    private static <T> T logIfNull(@Nullable T value) {
+        // Log noisily in production
+        if (value == null) {
+            log.warn("Detected a null value, null values are not allowed by Conjure", new IllegalArgumentException());
+        }
+        // Tests should fail, but we're not ready to cause breaks in deployed systems yet.
+        assert value != null : "Null values are not allowed by Conjure";
```
carterkozak: This may be contentious, interested in feedback.

I'd like tests which rely on unexpected behavior to fail rather than making their way to production and filing the logs -- the added logging should be considered a fallback.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
```diff
@@ -0,0 +1,128 @@
...
+ * Validates that collections are deserialized into data that Conjure allows. This module provides <i>soft</i>
+ * validation, logging instead of breaking functionality. Over time this will transition to strict validation,
+ * throwing exceptions when incorrect data is received.
+ * This implementation allows us to discover code which leverages incorrect behavior and remediate without
+ * breaking functionality.
+ *
+ * What is validated?
+ * <ul>
+ *     <li>Collection values must not be null</li>
+ *     <li>TODO: Values are not mutated (after deserialization)</li>
```
carterkozak: Open question: Do we want hard, or soft validation on mutability? Unlike null values, this is related entirely to implementation, not wire format.

markelliot (commented):
Thanks! Think this will help generate sufficient signal for us to make an informed decision and/or help existing services improve their messages

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
iamdanfox: This is a pretty cool solution :) 

I think we'd need to excavate something out (or bake something into a plugin) to make sure tests run with `-ea`, because I just tried putting `assert false` in a test and it still passed!

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -102,6 +102,7 @@ public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
                 .registerModule(new Jdk8Module().configureAbsentsAsNulls(true))
                 .registerModule(newSafeForNewJdksAfterburnerModule())
                 .registerModule(new JavaTimeModule())
+                .registerModule(new CollectionValidationModule())
```
iamdanfox, line 105: if someone objects to any performance implications of this, can they register their own CollectionModule to override this and deserialize to ImmutableLists directly?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
carterkozak: I think our intellij configuration is wonky, if gradle _ever_ runs tests without assertions enabled, that's a huge bug.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
carterkozak: We don't make mappers configurable in most places, if this causes performance regressions (which I don't expect, unless the case we're validating against is hit) we should fix them.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CollectionValidationModule.java
```diff
@@ -0,0 +1,146 @@
...
+            if (previousValue != null) {
+                // Avoid the performance cost of Throwable.fillInStackTrace if WARN has been disabled
+                if (log.isWarnEnabled()) {
+                    // TODO(ckozak): Log key and values as unsafe arguments. This would require us to take a
+                    // dependency on safe-logging.
+                    log.warn("Detected duplicate map keys which are not allowed by Conjure",
+                            new IllegalArgumentException());
+                }
+                assert false : "Duplicate values for the same key are not allowed by Conjure. Key '"
+                        + key + "' values ['" + value + "', '" + previousValue + "']";
```
schlosna: performance nit: might want to pull the logic inside the if block out to a separate method so that the main `put` method bytecode is small and gets JITed & inlined, while the duplicate logging & assert is out-lined in separate slow path.

carterkozak: Good idea.
I'm curious if you have any tricks for measuring whether a specific method hits the 35 byte sweet spot?

schlosna: `javap` for the classfile is quickest, but I also have been maintaining a [fork of the `jitwatch-intellij` plugin](https://github.com/schlosna/jitwatch-intellij) that you can run your unit tests/sample load.

Before (`put` was bytecode size 100) :
image

After (`put` is byte code size 23 so can get inlined):
image

I think we may want to extract out the `onNull` handling as well to optimize happy path:
```
    private static <T> T logIfNull(@Nullable T value) {
        // Log noisily in production
        if (value == null) {
            onNull();
        }
        return value;
    }

    private static void onNull() {
        // Tests should fail, but we're not ready to cause breaks in deployed systems yet.
        assert false : "Null values are not allowed by Conjure";

        // Avoid the performance cost of Throwable.fillInStackTrace if WARN has been disabled
        if (log.isWarnEnabled()) {
            log.warn("Detected a null value, null values are not allowed by Conjure", new IllegalArgumentException());
        }
    }
```

```
private static Object logIfNull(Object)
    0: aload_0         
    1: ifnonnull       7    
    4: invokestatic    #12  // Method onNull:()V
    7: aload_0         
    8: areturn         

private static void onNull()
     0: getstatic       #13  // Field $assertionsDisabled:Z
     3: ifne            16   
     6: new             #14  // class java/lang/AssertionError
     9: dup             
    10: ldc             #15  // String Null values are not allowed by Conjure
    12: invokespecial   #16  // Method java/lang/AssertionError."<init>":(Ljava/lang/Object;)V
    15: athrow          
    16: getstatic       #1   // Field log:Lorg/slf4j/Logger;
    19: invokeinterface #17, 1// InterfaceMethod org/slf4j/Logger.isWarnEnabled:()Z
    24: ifeq            44   
    27: getstatic       #1   // Field log:Lorg/slf4j/Logger;
    30: ldc             #18  // String Detected a null value, null values are not allowed by Conjure
    32: new             #19  // class java/lang/IllegalArgumentException
    35: dup             
    36: invokespecial   #20  // Method java/lang/IllegalArgumentException."<init>":()V
    39: invokeinterface #21, 3// InterfaceMethod org/slf4j/Logger.warn:(Ljava/lang/String;Ljava/lang/Throwable;)V
    44: return          
```

carterkozak: This is wonderful, thanks! Opened a PR with the changes I made to make the build succeed.

iamdanfox (approved):
:+1: - neat performance optimizations too! I think merging this makes sense because it'll just give us additional signal if people are doing bad things.

Will need to make IntelliJ do `-ea` somewhere else (possibly in baseline)

PR #983, opened 2019-03-15
[break] Preliminary support for the Conjure integer64 type

carterkozak:
For information describing the new type:
https://github.com/palantir/conjure/blob/rfc/007-integer64/docs/rfc/007-integer64-type.md

**BREAK** Any code which serializes a java long in a
Conjure `any` object will produce JSON string values as defined
in the specification for Conjure `integer64`.

This implementation adds a dependency from
`conjure-java-jackson-serialization` onto `conjure-lib` in order
to avoid writing `SafeLong` using the `integer64` serializer.

Deserialization is lenient, supporting deserialization of both
quoted and unquoted values for backwards compatibility with
existing java clients.


conjure-java-jackson-serialization/build.gradle
```diff
@@ -8,8 +8,12 @@ dependencies {
     compile "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"
     compile "com.fasterxml.jackson.dataformat:jackson-dataformat-cbor"
 
+    implementation 'com.palantir.conjure.java:conjure-lib', {
```
carterkozak, line 11: The SafeLong serializer needs to live in this project, rather than with SafeLong to avoid wire breaks from mixed conjure-java/conjure-java-runtime versions.

markelliot:
Regarding the break, as with some other recent discussions, please first add logging that will allow us to track the pervasiveness of the break prior to making the break.

carterkozak:
Unfortunately this is not quite as straightforward to test largely due to the total lack of specification around “any” serialization and deserialization.
Validation occurs viabtypr checking on the parsing side, which may not even be java.

This change is meant to fix serialization and deserialization of externally imported java.lang.Long objects in a way that can be leveraged for https://github.com/palantir/conjure/blob/rfc/007-integer64/docs/rfc/007-integer64-type.md
We could instead revert coercion changes, however that would likely be considered a larger break than this proposal.

Shorter term we can break out the lenient deserializer and merge without the serializer component to allow java clients to comminicate with servers that leverage long external type imports, even if the values are not encoded as json strings.

ferozco:
I quickly look through internal API definitions and it seems like the longs are not used within Foundry. Because of the lack of usage, I would be comfortable merging the break as is

carterkozak:
That's fair, based on #917 as precedent this should be fine to merge since we provide a clear explanation of the implications.

chrismale:
+1 to this change.  I am needing to use the OSS conjure client to talk to a cluster of _large internal product_, removing a bunch of legacy cruft around how we choose which node to talk too.  Having this change would enable that.

ferozco:
👍

j-baker:
Hey - I don't think this is reasonable to merge and I think we should revert it. This is a straight up wire format break, which is one of the only contracts we hold dear in Conjure. People internally use Conjure object mappers for long term serialization; if you've stored a long in your AtlasDB you'll now be unable to deserialize it while blue-green deploying and other things. I'm sorry - I meant to make this comment a few days ago before this merged, I think this will internally just cause an unplanned outage.

PR #984, opened 2019-03-17
[fix] correctly set traceids on retries, ConcurrencyLimiter considers host

j-baker:
I was in the area, and stopped by to fix a few bugs.

The first is that the tracing is not set correctly on retries. This
means that RPCs are associated with incorrect traces. We've seen
symptoms of this internally. Use the DeferredTracer to fix this.

The second is that we use a SettableFuture and setFuture on it within
the limiter listener request tag. If the future is already done (e.g.
this is a retry) then this will already be done, and we will not react
properly; we will attempt to close out the old listener and not the new one.

Interestingly this actually has no effect because the Netflix library
does not validate that you didn't call it twice and behaves as though
it was a different listener. But this could change at any time, and
then we'd be in a tough spot.

Lastly, factor the host you're talking to into the concurrency limiter
key. This means that if you're round robining one bad host can't hurt
your collective performance.

I don't think anyone is negatively affected by this internally - it can only really affect you if you rate limit globally and I don't know we do that internally.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -41,6 +41,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import javax.annotation.concurrent.GuardedBy;
 import okhttp3.Request;
+import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,7 +49,6 @@ final class ConcurrencyLimiters {
     private static final Logger log = LoggerFactory.getLogger(ConcurrencyLimiters.class);
     private static final Duration DEFAULT_TIMEOUT = Duration.ofMinutes(1);
     private static final Void NO_CONTEXT = null;
-    private static final String FALLBACK = "";
     private static final MetricName SLOW_ACQUIRE =
             MetricName.builder().safeName("conjure-java-client.qos.request-permit.slow-acquire").build();
     private static final MetricName LEAK_SUSPECTED =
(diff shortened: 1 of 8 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -101,12 +101,12 @@ private ConcurrencyLimiter newLimiter(String limiterKey) {
...
     }
 
-    private static String limiterKey(Request request) {
+    private String limiterKey(Request request) {
         String pathTemplate = request.header(OkhttpTraceInterceptor.PATH_TEMPLATE_HEADER);
         if (pathTemplate == null) {
             return FALLBACK;
         } else {
-            return request.method() + " " + pathTemplate;
+            return request.url().host() + " " + request.method() + " " + pathTemplate;
```
carterkozak: I think it’s time to key on a proper object

j-baker: heh, yes

j-baker: done

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -150,12 +151,16 @@ private static ResponseBody buffer(MediaType mediaType, byte[] bodyBytes) {
...
     public void enqueue(Callback callback) {
+        DeferredTracer tracer = new DeferredTracer();
         ListenableFuture<Limiter.Listener> limiterListener = limiter.acquire();
-        request().tag(ConcurrencyLimiterListener.class).limiterListener().setFuture(limiterListener);
+        request().tag(ConcurrencyLimiterListener.class).setLimiterListener(limiterListener);
         Futures.addCallback(limiterListener, new FutureCallback<Limiter.Listener>() {
             @Override
             public void onSuccess(Limiter.Listener listener) {
-                enqueueInternal(callback);
+                tracer.withTrace(() -> {
```
iamdanfox, line 160: :+1: good spot! Would be a shame to regress here - could you change one of the tests to capture this desired behaviour?

j-baker: done

iamdanfox:
:+1:

PR #965, opened 2019-03-03
[fix] Stop producing illegal reflective access warnings on JDK9+

markelliot:
## Before this PR
illegal reflective access warnings on JDK9+

## After this PR
no warnings

fixes #964

PR #987, opened 2019-03-18
[fix] Revert "[fix] Stop producing illegal reflective access warnings on JDK9+ (#965)"

carterkozak:
This reverts commit 5d945f40a302167524b4a2e5edf4df57704be03e.

This fixes IllegalAccessExceptions while deserializing immutables
generated classes.


iamdanfox:
:+1: Sounds like we'll have to find an alternative way to suppress those warnings :/

markelliot:
would be ideal to add the missing tests here?

iamdanfox:
Yep - just gonna get the fixed release out first, then have another go at silencing these warnings (and also add some coverage)

Issue #988, opened 2019-03-19
Make threads daemons


carrino:
https://github.com/palantir/conjure-java-runtime/pull/792

Don't keep the jvm alive if everything else has shut down.  This is producing a regression on one of our internal services that fails to start and never shuts down so doesn't get restarted and just hangs.

iamdanfox:
@carrino why not just revive that PR you linked and continue the discussion there?

The one case undesirable case I can think of is if a dying server wanted to kick off some kind of POST request, e.g. to ship some diagnostics somewhere or release some resources, a request might exist _only_ on the Okhttp Dispatcher threads, which if they were daemon, would then just be dropped...

Other than that, it sounds like a reasonable safe change that shouldn't hurt Foundry services...

carterkozak:
This sounds reasonable to me, I imagine our services are taking longer than necessary to stop while they wait the full timeout for idle threads to go away.

carrino:
I think the case you laid out isn't really useful in practice.  I think if I call post, i would wait for the result to ensure it's complete and not rely on the implementation detail that we are using threads under the covers.

carterkozak:
In practice it could be helpful, you would enqueue an action which triggers a callback. If you're not blocking on a future, the application could exit. The problem is the daemon threads live a full minute while idle, and don't differentiate blocking shutdown between running vs idle.
That said, I think this should use daemon threads because we're not worried about that case when the server shuts down. Our clients tend to use futures to either batch awaiting results or simply wait on the future, not to asynchronously trigger execution.
Additionally spark requires us to block the main thread until we're done, in which case it doesn't matter at all whether we call these threads daemons or not, so this could not break anything that is not already broken on spark.

carrino:
```
    /**
     * The {@link ScheduledExecutorService} used for scheduling call retries. This thread pool is distinct from OkHttp's
     * internal thread pool and from the thread pool used by {@link #executionExecutor}.
     * <p>
     * Note: In contrast to the {@link java.util.concurrent.ThreadPoolExecutor} used by OkHttp's {@link
     * #executionExecutor}, {@code corePoolSize} must not be zero for a {@link ScheduledThreadPoolExecutor}, see its
     * Javadoc.
     */
    private static final ScheduledExecutorService schedulingExecutor = Tracers.wrap(Executors.newScheduledThreadPool(
[...]

PR #989, opened 2019-03-19
[fix] ScheduledExecutors use daemon threads, to avoid blocking shutdowns

carrino:
## Before this PR

This library leaves non-daemon threads behind so services never shut down

## After this PR
==COMMIT_MSG==
ScheduledExecutors use daemon threads, so the JVM isn't prevented from shutting down
==COMMIT_MSG==

## Possible downsides?
Very minimal downsides for uses who fire and forget requests and shut down their server and expect the library to keep the JVM alive and retry under the covers.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -61,10 +61,14 @@ public final class OkHttpClients {
                                     + "which requires debugging.",
                             uncaughtException))
             .setNameFormat("remoting-okhttp-dispatcher-%d")
+            // The Dispatcher must *not* use daemon threads otherwise it would become impossible to make
+            // outgoing network requests in a shutdown hook.
+            .setDaemon(false)
             .build();
     /**
      * The {@link ExecutorService} used for the {@link Dispatcher}s of all OkHttp clients created through this class.
-     * Same as OkHttp's default, but with a logging uncaught exception handler.
+     * Same as OkHttp's default, but with a logging uncaught exception handler. In a cachedThreadPool, threads that
+     * have not been used for sixty seconds are terminated, so this won't block shutdown indefinitely.
      */
     private static final ExecutorService executionExecutor =
             Tracers.wrap(Executors.newCachedThreadPool(executionThreads));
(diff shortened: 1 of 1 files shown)


carrino:
https://github.com/palantir/conjure-java-runtime/issues/988

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -114,7 +114,7 @@
      * Javadoc.
      */
     private static final ScheduledExecutorService schedulingExecutor = Tracers.wrap(Executors.newScheduledThreadPool(
-            NUM_SCHEDULING_THREADS, Util.threadFactory("conjure-java-runtime/OkHttp Scheduler", false)));
+            NUM_SCHEDULING_THREADS, Util.threadFactory("conjure-java-runtime/OkHttp Scheduler", true)));
```
j-baker, line 121: there are other executors in this class; you probably want to daemonize all of them!

carrino: maybe, but this is the one that is causing me troubles

carterkozak: Please either update all executors, or add comments describing why they need to use non-daemon threads

carterkozak:
Can you please update the title and message describing what has changed for a better commit log.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
carrino: ok, better?

carrino:
comment added

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
carterkozak: Why doesn't this apply to the cached executor which also uses non-daemon threads? If anything has made a remote call I imagine that would block shutdown for a full minute as well even if actions are not running.

If we haven't made calls using conjure-java-runtiem clients, it's odd that we would create this executor. Should creation of these executors be deferred until a client is created using memoized suppliers?

iamdanfox:
I think we've now accounted for all the executors constructed in this library.

Screenshot 2019-03-26 at 18 12 00

ferozco:
👍

PR #990, opened 2019-03-19
Revert "[break] Preliminary support for the Conjure integer64 type"

j-baker:
Reverts palantir/conjure-java-runtime#983

> Hey - I don't think this is reasonable to merge and I think we should revert it. This is a straight up wire format break, which is one of the only contracts we hold dear in Conjure. People internally use Conjure object mappers for long term serialization; if you've stored a long in your AtlasDB you'll now be unable to deserialize it while blue-green deploying and other things. I'm sorry - I meant to make this comment a few days ago before this merged, I think this will internally just cause an unplanned outage.

and beyond this I found an internal usage which this would break (and cause DB corruption).


iamdanfox:
Hang on - the original PR does give us [lenient deserialization](https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/Integer64Module.java#L95-L99) (ie. both strings and numbers are accepted as longs).

If the concern is that a server might upgrade to this new version and start producing string longs straight away while clients are still on an old version of c-j-r so don't understand them yet, why don't we keep the lenient deserializer and _only revert_ the serializer changes introduced in #983?

Can you elaborate on how this could cause DB corruption?

j-baker:
I'll link you internally.

j-baker:
but like, to be clear, try figuring out how to ship this thing if you have an endpoint `echo` which takes `any` and returns the same `any`.

ferozco:
Even keeping the lenient deserialization would not work for `map<string, any>` as the values would be interpreted as strings. We should revert, and reconsider how to move forward with int64

j-baker:
For another thing, the break relates to the 'any' type only for Conjure defined types. Things like Timelock would be broken by this; they return longs from their endpoints which would now be quoted which would break their clients.

https://github.com/palantir/atlasdb/blob/develop/timelock-impl/src/main/java/com/palantir/atlasdb/timelock/AsyncTimelockResource.java#L66

iamdanfox:
OK - given we have some clear cases that this would be problematic, let's get this revert in so develop stays releasable and we can continue the discussion without risk of pushing out breaks prematurely.

:+1:

carterkozak:
Given this revert we need to either revert #917 or bring un-revert the lenient long deserializer to fix the break introduced in #917 (serialization would be unchanged).

Issue #985, opened 2019-03-18
Introduce a way to propagate 429


jonsyu1:
## What happened?

When the server returns a 429, the `RemotingOkHttpCall` uses its backoff strategy to determine how to retry the request and throws a `SafeIoException` when it gives up.

## What did you want to happen?

When using `conjure-java-runtime` in a service that proxies user requests, we want to propagate the 429 to the user so they can decide how to handle it. Interactive clients and long running jobs will want to handle the 429 differently. Propagating the 429 puts the upstream client in control of the retry behavior and eliminates the proxy service's need to understand how to retry the request.

pkoenig10:
It would also be useful to immediately return a 429 when we know a client request will take a long time due to a large number of already queued requests.

In the current implementation that client request would complete, but it would be slow due to being queued behind the existing requests. Immediately returning a 429 to provide back pressure will force clients to back off and prevent us from queueing requests indefinitely.

markelliot:
@pkoenig10 that's already a thing you can accomplish by throwing the appropriate exception

jonsyu1:
Fixed in https://github.com/palantir/conjure-java-runtime/pull/1041

PR #991, opened 2019-03-19
[improvement] Provide a way to propagate 429s

jonsyu1:
Repurpose BackoffStrategy to also indicate whether or not an operation should be retried at all. This allows us to propagate 429s instead of always throwing a SafeIoException on failure.

Fixes: #985 

## Before this PR
A 429 always results in either a successful retry or a `SafeIoException`. Sometimes the client does not have enough context to decide how to handle a 429 and wants to propagate the 429 to an upstream client.

## After this PR
When `ClientConfiguration.maxNumRetries` is set to 0, 429s are propagated instead of rethrown as `SafeIoException`s.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -63,7 +63,7 @@ final class RemotingOkHttpCall extends ForwardingCall {
     private static final ResponseHandler<IOException> ioExceptionHandler = IoExceptionResponseHandler.INSTANCE;
     private static final ResponseHandler<QosException> qosHandler = QosExceptionResponseHandler.INSTANCE;
 
-    private final BackoffStrategy backoffStrategy;
+    private final RetryStrategy retryStrategy;
     private final UrlSelector urls;
     private final RemotingOkHttpClient client;
     private final ScheduledExecutorService schedulingExecutor;
@@ -74,7 +74,7 @@ final class RemotingOkHttpCall extends ForwardingCall {
 
     RemotingOkHttpCall(
             Call delegate,
-            BackoffStrategy backoffStrategy,
+            RetryStrategy retryStrategy,
             UrlSelector urls,
             RemotingOkHttpClient client,
             ScheduledExecutorService schedulingExecutor,
@@ -82,7 +82,7 @@ final class RemotingOkHttpCall extends ForwardingCall {
             ConcurrencyLimiters.ConcurrencyLimiter limiter,
             int maxNumRelocations) {
         super(delegate);
-        this.backoffStrategy = backoffStrategy;
+        this.retryStrategy = retryStrategy;
         this.urls = urls;
         this.client = client;
         this.schedulingExecutor = schedulingExecutor;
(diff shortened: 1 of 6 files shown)


jonsyu1:
Looks like the gradle task timed out the circle builds. My guess is that some end to end test is hanging.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -273,7 +273,11 @@ private void scheduleExecution(Runnable execution, Duration backoff) {
         return new QosException.Visitor<Void>() {
             @Override
             public Void visit(QosException.Throttle exception) {
-                Optional<Duration> nonAdvertizedBackoff = backoffStrategy.nextBackoff();
+                if (!retryStrategy.shouldRetry()) {
+                    throw exception;
```
uschi2000: Don't remember 100% of the details, but I believe you should only throw IOException within the okhttp callstack. Maybe an alternative: throw a more special SafeIoException below?

jonsyu1: Instead of introducing a new `SafeIoException` for the user to handle, how about we return a 429 `Response`? This does not introduce any additional API since the caller already gets back a `Response`. The caller just needs to check the status code and throw a `QosException` if it's 429.

bavardage: @jonsyu1 not sure I understand what you're proposing - surely it's possible we could get a real 404 and wouldn't want that to be turned into a 429..? Would a real 404 show up differently somehow?

jonsyu1: Sorry, I don't know why I typed 404 when I meant to type 429. I updated my comment.

bavardage: I believe it would be better to pass an exception out via `callback.onFailure(call, ...)` rather than throwing at this point in the code. I think that would make sure that the exception bubbles up in the right way up in https://github.com/palantir/conjure-java-runtime/blob/266a35b5550ff4d50cd69e53e645aa6b609f6723/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java#L119

On the other hand I don't know this code at all well, so could be talking nonsense.

I feel it would be worth getting to the point where the tests pass with your initial refactors (the `BackoffStrategy` -> `RetryStrategy` piece) but then catching up with someone on the @palantir/infrastructure team in a higher bandwidth setting to figure out the correct approach here.

jonsyu1: What I'm trying to suggest above is passing a 429 via `callback.onResponse(call, response)` because this is easier for the caller to handle.

For retrofit clients, this means returning a `Call` whose `Response.code()` returns 429. This is more intuitive than catching `IOException` and checking if it is a special `SafeIoException` because it is already part of retrofit's normal API.

For feign clients, this means throwing a `FeignException` whose `status()` returns 429. This is more intuitive than checking the `FeignException`'s cause for a special `SafeIoException` because it is already part of feign's normal API.

In its current form, the `Response` is turned into a `QosException` and the visitor does not get access to the `Response`. We could have the `qosHandler` return the `Response` in addition to the `QosException` but that seems to leak the `Response` from the `ResponseHandler` abstraction. Alternatively, we could construct a new `ResponseHandler` that uses the `qosHandler` to turn the `Response` into a `QosException` and then invokes the `callback.onResponse` with the `Response`.

I'm happy to iterate on this over VTC or slack.

jonsyu1:
I opted for a different approach outlined by my [comment above](https://github.com/palantir/conjure-java-runtime/pull/991#discussion_r271034401) in #1041.

PR #992, opened 2019-03-20
[fix] java long deserialization regression introduced in #917

carterkozak:
Java clients and servers will once again support deserialization
of long values encoded as JSON strings.

## Before this PR
Disabling `MapperFeature.ALLOW_COERCION_OF_SCALARS` in #917 broke existing services which used java long values encoded as strings (base-type: string) for interoperability with javascript clients.

## After this PR
==COMMIT_MSG==
Java clients and servers once again support deserialization of `java.lang.Long` values encoded as JSON strings.
==COMMIT_MSG==

## Possible downsides?
None known

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
@@ -56,6 +56,11 @@ final class LenientLongModule extends SimpleModule {
             throw new IOException("Expected a long value");
         }
 
+        @Override
+        public boolean isCachable() {
+            return true;
+        }
+
         private static Long parseLong(JsonParser jsonParser) throws IOException {
             try {
                 return Long.valueOf(jsonParser.getValueAsString());
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
@@ -49,6 +49,8 @@ final class LenientLongModule extends SimpleModule {
                     return jsonParser.getLongValue();
                 case VALUE_STRING:
                     return Long.valueOf(jsonParser.getValueAsString());
+                case VALUE_NULL:
+                    return null;
             }
             throw new IOException("Expected a long value");
         }
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import java.io.IOException;
+
+/**
+ * Provides support for the {@link Long} deserialization from JSON string and numeric values regardless of
+ * <pre>MapperFeature.ALLOW_COERCION_OF_SCALARS</pre> configuration.
+ */
+final class LenientLongModule extends SimpleModule {
+
+    LenientLongModule() {
+        super("lenient long");
+        // Register to both Long.TYPE and Long.class
+        this.addDeserializer(long.class, new LongAsStringDeserializer())
+                .addDeserializer(Long.class, new LongAsStringDeserializer());
+    }
+
+    private static final class LongAsStringDeserializer extends StdDeserializer<Long> {
+
+        private LongAsStringDeserializer() {
+            super(Long.TYPE);
+        }
+
+        @Override
+        public Long deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+            switch (jsonParser.currentToken()) {
+                case VALUE_NUMBER_INT:
+                    // Lenient implementation for compatibility with existing clients
+                    return jsonParser.getLongValue();
+                case VALUE_STRING:
+                    return Long.valueOf(jsonParser.getValueAsString());
+            }
+            throw new IOException("Expected a long value");
+        }
+    }
+}
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.serialization;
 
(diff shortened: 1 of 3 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        @Override
+        public Long deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+            switch (jsonParser.currentToken()) {
+                case VALUE_NUMBER_INT:
+                    // Lenient implementation for compatibility with existing clients
+                    return jsonParser.getLongValue();
+                case VALUE_STRING:
+                    return Long.valueOf(jsonParser.getValueAsString());
```
schlosna: Do we need to handle empty string?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
```diff
@@ -0,0 +1,56 @@
...
+            }
```
schlosna, line 54: Do we need to handle `VALUE_NULL`?

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -136,6 +136,26 @@ public void testMapWithDuplicateKeys() {
...
+
+    @Test
+    public void testLongDeserializationFromJsonNumber() throws IOException {
+        assertThat((Long) MAPPER.readValue("1", new TypeReference<Long>() {})).isEqualTo(1L);
+    }
+
+    @Test
+    public void testLongTypeDeserializationFromJsonNumber() throws IOException {
+        assertThat(MAPPER.readValue("1", Long.TYPE)).isEqualTo(1L);
+    }
```
schlosna, line 171: Can we test with some additional cases (`Integer.MAX_VALUE + 1L`, `Long.MAX_VALUE`, empty, null, etc.)? QuickTheories would be nice

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
carterkozak: Empty string should cause an exception because the java type expects a `long`, this is similar to `OffsetDateTime` deserialization failures given invalid date strings.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
carterkozak: I suppose it's possible for java code to expect a boxed `Long` in a nullable place, though in an ideal world we would expect `OptionalLong`. Will update.

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
carterkozak: Good call, updated, though I have not added the quicktheories dep in this PR.

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -136,6 +140,91 @@ public void testMapWithDuplicateKeys() {
...
+        assertThatThrownBy(() -> MAPPER.readValue("" + large, Long.TYPE))
+                .isInstanceOf(JsonParseException.class)
+                .hasMessageContaining("out of range of long");
+    }
+
+    @Test
+    public void testLongAsStringOverflowDeserialization() {
+        BigInteger large = BigInteger.valueOf(Long.MAX_VALUE).add(BigInteger.ONE);
+        assertThatThrownBy(() -> MAPPER.readValue("\"" + large + "\"", Long.TYPE))
+                .isInstanceOf(NumberFormatException.class);
```
schlosna: Should this be a JsonParseException constricted with deserialization context?

carterkozak: That's a good question. Existing validation for primitive values does produce JsonParseException, however other input value validation (OffsetDateTime, ResourceIdentifier, etc) does not produce JsonParseException.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
```diff
@@ -0,0 +1,58 @@
...
+final class LenientLongModule extends SimpleModule {
+
+    LenientLongModule() {
+        super("lenient long");
+        // Register to both Long.TYPE and Long.class
+        this.addDeserializer(long.class, new LongAsStringDeserializer())
+                .addDeserializer(Long.class, new LongAsStringDeserializer());
+    }
+
+    private static final class LongAsStringDeserializer extends StdDeserializer<Long> {
```
schlosna, line 39: Not sure we need the full complexity of `LongDeserializer` from 
https://github.com/FasterXML/jackson-databind/blob/d2c083a6220f2875c97c29f4823d9818972511dc/src/main/java/com/fasterxml/jackson/databind/deser/std/NumberDeserializers.java#L536 but may want to reference the implementation

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
ferozco: Should we not fail if a long is expected but the value is absent?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
```diff
@@ -0,0 +1,72 @@
...
+
+        private LongAsStringDeserializer() {
+            super(Long.TYPE);
+        }
+
+        @Override
+        public Long deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+            switch (jsonParser.currentToken()) {
+                case VALUE_NUMBER_INT:
+                    // Lenient implementation for compatibility with existing clients
```
j-baker: think this comment is stale, right?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
carterkozak: We should, but that's a separate concern from this parsing function. It's possible that existing human generated objects are expected to be deserialized as `@Nullable Long value` which we shouldn't cause to throw in this change.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
carterkozak: Yep, fixed.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/LenientLongModule.java
```diff
@@ -0,0 +1,71 @@
...
+        public Long deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException {
+            switch (jsonParser.currentToken()) {
+                case VALUE_NUMBER_INT:
+                    return jsonParser.getLongValue();
+                case VALUE_STRING:
+                    return parseLong(jsonParser);
+                case VALUE_NULL:
+                    return null;
+            }
+            throw new IOException("Expected a long value");
```
iamdanfox, line 55: Might as well do a SafeIoException?

carterkozak: That would be better, but requires adding a safe-logging:preconditions dependency to this library. Are we okay with that?

iamdanfox:
:+1: Seems like this will provide clear upside in allowing an internal product to ship Long IDS as strings over the wire.  Given that we previously had this behaviour in https://github.com/palantir/conjure-java-runtime/pull/917, seems like it should be a safe change.

carterkozak:
policy-bot is sad

PR #996, opened 2019-03-21
[fix] Relax CollectionValidationModule map validation to match generated objects

carterkozak:
Null map values no longer trigger WARN logging or assertions.
Tracking conjure-java failure to validate null map values here:
https://github.com/palantir/conjure-java/issues/291

## After this PR
==COMMIT_MSG==
Null map values no longer trigger WARN logging or assertions.
==COMMIT_MSG==


ferozco:
👍

PR #1000, opened 2019-03-25
[improvement] Allow use of a no-op concurrency limiter via code configuration

markelliot:
## Before this PR

It's not possible to disable sympathetic client-side QoS, even when it interacts poorly with misbehaving servers.

## After this PR
==COMMIT_MSG==
Allow use of a no-op concurrency limiter via code configuration.
==COMMIT_MSG==

## Possible downsides?
Users might abuse this setting to remove a feature we generally want -- this PR attempts to mitigate that by:
1. Requiring that configuration be performed in code
2. Adding the text `forExpertsOnly` to the beginning of the method and clear JavaDoc to indicate the limitations under which this configuration should be applied.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -101,6 +101,13 @@ public interface ClientConfiguration {
      */
     Duration backoffSlotSize();
 
+    /**
+     * Disables the client-side sympathetic QoS. Consumers should almost always set this value to its default of false,
+     * except where there are known issues with the QoS interaction. Please consult project maintainers if flipping
+     * this value to true.
+     */
+    boolean forExpertsOnlyDisableSympatheticClientQoS();
+
     @Value.Check
     default void check() {
         if (meshProxy().isPresent()) {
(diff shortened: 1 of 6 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -101,6 +101,13 @@
      */
     Duration backoffSlotSize();
 
+    /**
+     * Disables the client-side sympathetic QoS. Consumers should almost always set this value to its default of false,
+     * except where there are known issues with the QoS interaction. Please consult project maintainers if flipping
+     * this value to true.
+     */
+    boolean forExpertsOnlyDisableSympatheticClientQoS();
```
iamdanfox: Rather than using a boolean, what about an enum something like the following:

```
enum ClientSideQoS {
  ENABLED, 
  DANGEROUS_DISABLE_CLIENT_QOS
}
``` 

This would mean you don't need the `private static final boolean DEFAULT_DISABLE_SYMPATHETIC_QOS = false;` thing below?

iamdanfox:
Sweet happy with the user-facing API here - we can always tinker with the internal classes later if we need to!

:+1:

PR #1002, opened 2019-03-26
Revert "[improvement] log.warn if deserialized raw collections contain null (#979)"

carterkozak:
The Conjure ObjectMapper is used too many places internally which are not directly
connected to Conjure. It is infeasible to implement this sort of validation without
a hook from generated code to opt in.

## Before this PR
Applications attempt to deserialize lists with null values using Conjure ObjectMapper instances.

## After this PR
==COMMIT_MSG==
Revert collection null validation.

The Conjure ObjectMapper is used too many places internally which are not
directly connected to Conjure. It is infeasible to implement this sort of
validation without a hook from generated code to opt in.

This reverts commit 78bde41444f90730d9e46e9b4cc4f8bd06e95e15.
==COMMIT_MSG==


iamdanfox:
:+1: I've also seen a few cases where people were relying on deserializing user-authored JSON containing nulls.

Perhaps we could hit the same outcome by making conjure-java use ImmutableMaps and ImmutableLists internally? I don't know how valuable it really is to make our API objects guava-free :/

PR #1005, opened 2019-03-26
[fix] Add back acquireLimiter span

robert3005:
## Before this PR
after #890 we no longer get acquireLimiter span 

## After this PR
==COMMIT_MSG==
Add span for acquireLimiter operation
==COMMIT_MSG==

## Possible downsides?
N/A

@ellisjoe @markelliot


conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/TracerTest.java
```diff
@@ -54,18 +54,25 @@ public void before() {
...
 
+        AtomicInteger called = new AtomicInteger();
         Tracer.subscribe(TracerTest.class.getName(), span -> {
-            assertThat(span.getOperation(), either(is("acquireLimiter")).or(is("GET /{param}")));
+            if (span.type().equals(SpanType.CLIENT_OUTGOING)) {
+                assertThat(span.getOperation(), is("GET /{param}"));
+            } else {
+                assertThat(span.getOperation(), is("acquireLimiter"));
+            }
+            called.getAndIncrement();
```
carterkozak: nit: I'd prefer to push the spans into a collection and verify after. As written this will pass if the `acquireLimiter` span is created twice (also true of the previous implementation).

robert3005: Ok

robert3005:
This pr will break with tracing-java 2.9.0 after https://github.com/palantir/tracing-java/pull/97

robert3005:
This needs a more thorough change to tracing java, will reopen once that's done

PR #1006, opened 2019-03-26
Simplify threadworkqueue

j-baker:
No behaviour change, simpler implementation.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ThreadWorkQueue.java
```diff
@@ -36,36 +34,29 @@
...
 
     T remove() {
-        long id = nextThread();
-        Queue<T> workQueue = queuedRequests.get(id);
-        T result = workQueue.remove();
-        if (workQueue.isEmpty()) {
-            queuedRequests.remove(id);
-        } else {
-            activeQueuedThreads.add(id);
+        Map.Entry<Long, Queue<T>> workQueue = queuedRequests.get(id);
```
iamdanfox: do you just want to getfirst here?

iamdanfox (approved):

PR #1010, opened 2019-03-26
[improvement] OkHttpClients memoizes scheduled executors

carterkozak:
## Before this PR
Previously we created two threads on OkHttpClients class
initialization.

## After this PR
==COMMIT_MSG==
OkHttpClients memoizes scheduled executors, avoid thread creation on static initialization.
==COMMIT_MSG==


iamdanfox:
Ok the outcome here is that if someone has c-j-r on the classpath, but never actually constructs a client, then their JVM will run with two fewer threads?

I wonder if this should even be turned into an errorprone check?

:+1:

carterkozak:
That’s correct. An error prone check could be helpful in some cases. This sort of thing is fine for services, but should be avoided in library code.

PR #1011, opened 2019-03-26
[improvement] OkHttpClients uses non-daemon threads

carterkozak:
## Before this PR
Non-daemon threads prop up the JVM for a full minute after any OkHttpClients client completes a request.

## After this PR
==COMMIT_MSG==
[improvement] OkHttpClients uses non-daemon threads

Using Conjure OkHttpClients no longer blocks JVM shutdown for one
minute following the last usage.
==COMMIT_MSG==

## Possible downsides?
OkHttpClient **asynchronous** usage will no longer block JVM shutdown. Usage with feign is unaffected, usage with retrofit is only affected if callbacks are registered to responses and no threads are blocking waiting for completion. Any scenarios affected by this change would also fail on spark.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -63,14 +63,21 @@ public final class OkHttpClients {
                                     + "which requires debugging.",
                             uncaughtException))
             .setNameFormat("remoting-okhttp-dispatcher-%d")
-            // The Dispatcher must *not* use daemon threads otherwise it would become impossible to make
-            // outgoing network requests in a shutdown hook.
-            .setDaemon(false)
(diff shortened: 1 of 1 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -61,14 +61,21 @@
                                     + "which requires debugging.",
                             uncaughtException))
             .setNameFormat("remoting-okhttp-dispatcher-%d")
-            // The Dispatcher must *not* use daemon threads otherwise it would become impossible to make
-            // outgoing network requests in a shutdown hook.
```
carterkozak, line 67: This comment was incorrect. Requests will still work fine in hooks as long as we wait for our futures. A call being in flight without anything waiting for the result will no longer block shutdown.

iamdanfox:
:+1: for nice descriptive comments.

PR #1014, opened 2019-03-27
[improvement] Provide span operation when using deferred traces

pkoenig10:
## Before this PR
When using versions of tracing-java with https://github.com/palantir/tracing-java/pull/97, the spans created by the concurrency limiter and the dispatcher have the same name as the parent span. This is confusing since the parent span name does not accurately describe the work of the concurrency limiter or the dispatcher.

## After this PR
The concurrency limiter and the dispatcher provide a unique span operation when creating spans.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandler.java
```diff
@@ -44,7 +44,7 @@
                     : "<empty>";
 
             return Optional.of(new SafeIoException(
-                    String.format("Error %s. (Failed to parse response body as SerializableError.)", response.code()),
+                    "Failed to parse response body as SerializableError",
```
pkoenig10, line 47: This needed to change as a result of upgrading `safe-logging` which requires these messages to be compile-time constants.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -73,7 +73,7 @@
      * have not been used for sixty seconds are terminated, so this won't block shutdown indefinitely.
      */
     private static final ExecutorService executionExecutor =
-            Tracers.wrap(Executors.newCachedThreadPool(executionThreads));
+            Tracers.wrap("dispatcher", Executors.newCachedThreadPool(executionThreads));
```
carterkozak: nit: Can we provide more context by prefixing these with `OkHttp: `?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,7 +151,7 @@ private static ResponseBody buffer(MediaType mediaType, byte[] bodyBytes) {
 
     @Override
     public void enqueue(Callback callback) {
-        DeferredTracer tracer = new DeferredTracer();
+        DeferredTracer tracer = new DeferredTracer("acquire-limiter");
```
carterkozak: Would be nice if we prefixed this as well

okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandlerTest.java
```diff
@@ -51,7 +51,7 @@ public void extractsIoExceptionForAllErrorCodes() {
         for (int code : ImmutableList.of(300, 400, 404, 500)) {
             IOException exception = decode(MediaType.APPLICATION_JSON, code, "body").get();
             assertThat(exception.getMessage()).isEqualTo(
```
carterkozak, line 53: nit: could use `assertThatLoggableExceptionThrownBy` here, even though safe-logging wasn't introduced by your change.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandler.java
carterkozak: 👍

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
pkoenig10: What do you want to prefix this with? `OkHttp: `?

carterkozak: Yep, we should group by component, not necessarily class

okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandlerTest.java
pkoenig10: We're not throwing an exception here and the `decode` method returns a regular `IOException`. So we can either create a callable that throws this exception or cast it to a `SafeIoException` but both of these options seems a bit gross. Happy to do this if you would prefer though.

carterkozak:
lgtm 👍

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,7 +151,7 @@ private static ResponseBody buffer(MediaType mediaType, byte[] bodyBytes) {
 
     @Override
     public void enqueue(Callback callback) {
-        DeferredTracer tracer = new DeferredTracer();
+        DeferredTracer tracer = new DeferredTracer("OkHttp: acquire-limiter");
```
robert3005: I don't think this name is accurate as you need to open span now and close it in the success while this only make sure that there's same trace around retries

pkoenig10: What alternative name would you suggest?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,7 +151,7 @@ private static ResponseBody buffer(MediaType mediaType, byte[] bodyBytes) {
 
     @Override
     public void enqueue(Callback callback) {
-        DeferredTracer tracer = new DeferredTracer("OkHttp: acquire-limiter");
+        DeferredTracer tracer = new DeferredTracer("OkHttp: enqueue");
```
carterkozak, line 154: @robert3005 Are you happy with this?

robert3005: LGTM

robert3005:
👍

abless:
Note that this broke the Salt client: https://github.palantir.build/foundry/salt/blob/develop/salt-client/src/main/java/com/palantir/salt/client/Salt.java#L1368

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/IoExceptionResponseHandler.java
abless: This broke clients that previously tried parsing the error code.

PR #1023, opened 2019-03-29
Bring back old exception message

ellisjoe:


carterkozak:
I prefer this over the alternative, but I would prefer to avoid control flow based on exception message parsing. Too late in this case I suppose.

carterkozak:
👍

iamdanfox:
I've fixed up the internal product that was regexing exception messages.

carterkozak:
Thanks Dan!

ellisjoe:
fixed this in the client lib

PR #1024, opened 2019-03-29
Use AIMD limit in http-remoting concurrency limiter

pkoenig10:
## Before this PR
The latest version of http-remoting used the vegas limit in the concurrency limiter.

## After this PR
==COMMIT_MSG==
The latest version of http-remoting now uses the AIMD limit with no timeout in the concurrency limiter.
==COMMIT_MSG==


iamdanfox:
@pkoenig10 I'm a bit uncomfortable about backporting stuff like this - who is still on old remoting and not made the jump to conjure-java-runtime?

iamdanfox:
Ok after chatting to @markelliot, I'm ok with getting this in as it is really intended for a very small number of internal products.

Will tag a 3.46.0

iamdanfox (approved):

PR #1026, opened 2019-03-29
[fix] Add back acquire limiter span

robert3005:
## Before this PR
There was no way to measure time spent in acquire limiter

## After this PR
==COMMIT_MSG==
OkHttp: enqueue span is renamed to OkHttp: execute and gets OkHttp: execute-enqueue for time spent acquiring limiter and OkHttp: execute-run for execution of the call
==COMMIT_MSG==

PR #1039, opened 2019-04-03
[fix] Downgrade okhttp 3.14.0 -> 3.13.1 to avoid UnrepeatableRequestBody NoSuchMethodErrors 

dansanduleac:
## Before this PR

https://github.com/palantir/conjure-java-runtime/pull/1026 (which went into 4.20.1) inadvertently brought in a bump in okhttp to 3.14.0.
One of our internal libraries relied on `UnrepeatableRequestBody` which got removed in github.com/square/okhttp/pull/4676, and a solution using `RequestBody#isOneShot` couldn't be made that exhibits the previous behaviour.

## After this PR

Downgrading okhttp to 3.13.1, and adding a constraint to prevent accidental upgrades to 3.14.0 for the time being.


iamdanfox:
:+1:

PR #1050, opened 2019-04-09
[improvement] addition of new node strategy enum

ghost:
## Before this PR
Only two options are available for node selection which limits the control that a client has over which node handles a particular request. 

## After this PR
==COMMIT_MSG==
Creates the enum for a new strategy which will allow the selection of nodes based on a header value otherwise defaults to round robin selection.
==COMMIT_MSG==

## Possible downsides?
Strategy is not yet implemented - this will cause breaks if anyone tries to use the enum in the mean time.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
@@ -30,7 +30,7 @@ public enum NodeSelectionStrategy {
     ROUND_ROBIN,
 
     /**
-     * Try to pin to a particular node based on HEURISTIC_NODE_VALUE header otherwise round robin.
+     * Try to pin to a particular node based on NODE_PIN_VALUE header otherwise round robin.
      */
-    HEURISTIC_PIN
+    PIN_FROM_REQUEST
 }
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
@@ -30,7 +30,7 @@ public enum NodeSelectionStrategy {
     ROUND_ROBIN,
 
     /**
-     * Try to pin to a particular node based on NODE_PIN_VALUE header otherwise pin to a node.
+     * Try to pin to a particular node based on Node-Pin-Value header otherwise use PIN_UNTIL_ERROR strategy.
      */
     PIN_FROM_REQUEST
 }
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
@@ -30,7 +30,7 @@ public enum NodeSelectionStrategy {
     ROUND_ROBIN,
 
     /**
-     * Try to pin to a particular node based on NODE_PIN_VALUE header otherwise round robin.
+     * Try to pin to a particular node based on NODE_PIN_VALUE header otherwise pin to a node.
      */
     PIN_FROM_REQUEST
 }
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
@@ -27,5 +27,10 @@ public enum NodeSelectionStrategy {
     /**
      * For each new request, select the "next" node (in some undefined order).
      */
-    ROUND_ROBIN
+    ROUND_ROBIN,
+
+    /**
+     * Try to pin to a particular node based on HEURISTIC_NODE_VALUE header otherwise round robin.
+     */
+    HEURISTIC_PIN
 }
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConsistentHashRing.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConsistentHashRing.java
@@ -16,9 +16,9 @@
 
 package com.palantir.conjure.java.okhttp;
 
-import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableList;
 import com.google.common.hash.Hashing;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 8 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/NodeSelectionStrategy.java
```diff
@@ -27,5 +27,10 @@
     /**
      * For each new request, select the "next" node (in some undefined order).
      */
-    ROUND_ROBIN
+    ROUND_ROBIN,
+
+    /**
+     * Try to pin to a particular node based on NODE_PIN_VALUE header otherwise round robin.
+     */
+    PIN_FROM_REQUEST
```
uschi2000, line 35: how is this different from PIN_UNTIL_ERROR together with a 308+Location?

ghost: @ellisjoe @iamdanfox 

The context is that we want to have conjure select which node to run on and balance the load, but we need to be able to guarantee that a set of requests will be executed on the same node. So we don't want the request to specify the actual node, just ensure it will run on the same node as requests with the same 'NODE_PIN_VALUE' header value.

ellisjoe: yeah, we want to be able to allow the client to specify a header value and for conjure to send all requests with the same header value to the same remote host (eg: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing/load_balancers#ring-hash)

I think the implementation of this should be small enough to stick in this PR? @anderoo

uschi2000: @anderoo Can you please write up the full protocol for how this is intended to work, including the first request, and what happens in case of failure (e.g., a server responds with 503 or connection timeout)

iamdanfox:
Don't think we can merge this node selection strategy until it's implemented!

If possible, I'd love to try and align terminology with envoy (as that's what we're pushing for long-term): https://www.envoyproxy.io/docs/envoy/v1.7.0/intro/arch_overview/load_balancing

Might be interesting to match the ring hashing thing they have too - how would the product you work on behave if a new host is added part way through a transaction?

ghost:
@iamdanfox Yes, the plan is to implement some ring hash.

Adding / Removing hosts during a transaction is tricky, it's a trade off between consistency of hash values and load balancing. If we want to focus on load balancing during this phase then we'd have to sacrifice all current transactions and have them retry with the updated ring (If the add/remove impacted the hash value of these transactions).

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConsistentHashRing.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.hash.Hashing;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import okhttp3.HttpUrl;
+
+final class ConsistentHashRing {
+    private final List<HttpUrl> nodes = new ArrayList<>();
```
jkozlowski: This needs to be threadsafe, look at the usages.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConsistentHashRing.java
```diff
@@ -0,0 +1,53 @@
...
+    Optional<HttpUrl> getNode(String nodePinValue) {
+        if (nodes.isEmpty()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(nodes.get(hash(nodePinValue)));
+    }
+
+    private int hash(String nodePinValue) {
+        return Hashing.consistentHash(Hashing.crc32c().hashString(nodePinValue, StandardCharsets.UTF_8), nodes.size());
```
jkozlowski: I guess I need to read up here, but why crc? Curious what do things like envoy use?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -155,12 +157,26 @@ private static String switchWsToHttp(String url) {
...
+
     @Override
     public void markAsFailed(HttpUrl failedUrl) {
         if (useFailedUrlCache) {
             Optional<Integer> indexForFailedUrl = indexFor(failedUrl);
-            indexForFailedUrl.ifPresent(index ->
-                    failedUrls.put(baseUrls.get(index), UrlAvailability.FAILED)
+            indexForFailedUrl.ifPresent(index -> {
+                        failedUrls.put(baseUrls.get(index), UrlAvailability.FAILED);
+                        consistentHashRing.removeNode(baseUrls.get(index));
```
jkozlowski, line 178: I'm not sure this is right, the URLs never reenter the ring? I think you just want to use the failedUrls thing in redirectToHash to somehow select next node to try?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -155,12 +157,26 @@ private static String switchWsToHttp(String url) {
         return redirectTo(current, baseUrls.get((currentUrl.get() + 1) % baseUrls.size()));
     }
 
+    @Override
+    public Optional<HttpUrl> redirectToHash(HttpUrl current, Optional<String> nodePinValue) {
```
jkozlowski, line 161: We are also missing code changes in #redirectToNext above: RemotingOkHttpCall uses that for retries for requests that fail in #enqueueInternal; as it stands it would just use the currentUrl pinning, which is not what we want.

What we actually want is not clear to me: @ellisjoe what is the retry behavior here, because we have multiple retry paths (at the okhttp layer and application layer in RemotingOkHttpCall).

My intuition tells me we actually don't want to retry on a different host in this setup, instead we should fail the entire chain of requests and make the whole thing retry?

That would certainly be my preference for a sticky system where the chain of requests acumulates some state that is tied to a particular node.

iamdanfox:
Interested in exploring this alternative approach: https://github.com/palantir/conjure-java-runtime/pull/1056.

It doesn't guarantee that all requests in one transaction hit the same node, but then I think neither does the ring hashing?

ghost:
Closing as we will try to experiment internally on this first.

PR #1052, opened 2019-04-09
[fix] Handle ClientQoS cases explicitly

jonsyu1:
Explicitly enumerate ClientQoS cases instead of only checking for the
ENABLED case.

FLUP to https://github.com/palantir/conjure-java-runtime/pull/1041#discussion_r273649953


iamdanfox (approved):
Sure, I guess error-prone's exhaustiveness checker will help us now

PR #1053, opened 2019-04-10
[deprecation] Move feign encoders and decoders out of the "feign" package

carterkozak:
Our style guide states that our code must not live in other packages:
> The package name of internal as well as open source projects
> starts with `com.palantir.<project name>`
https://github.com/palantir/gradle-baseline/blob/develop/docs/java-style-guide/readme.md#package-statement

Deprecation: Classes which were previously in the `feign` package
have been migrated into `com.palantir.conjure.java.client.jaxrs.feignimpl`.
These are implementation details of the conjure-java-jaxrs-client library,
however they are still accessable to avoid blocking existing consumers.
We have provided deprecated classes in the `feign` package which shim to
the new location for now.

Classes originally existed in the feign package to use the
package-private `Types.getRawType` utility. This change updates
those uses to a new internal utility `RawTypes.get` which is
implemented using guava TypeToken.getRawType.

## Before this PR
We failed to follow our style rules. This resulted in classpath conflicts moving from `remoting3` to `conjure-java-runtime` because only `com.palantir` packages were renamed.

## After this PR
==COMMIT_MSG==
Move feign encoders and decoders out of the "feign" package.
==COMMIT_MSG==

## Possible downsides?
Code which used these classes directly will get deprecation warnings.

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/RawTypes.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/RawTypes.java
@@ -1,0 +1,32 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.conjure.java.client.jaxrs.feignimpl;
+
+import com.google.common.reflect.TypeToken;
+import java.lang.reflect.Type;
+
+/**
+ * Utility functionality to detect raw types.
+ */
+final class RawTypes {
+
+    /** Attempts to extract the raw type class from a {@link Type}. */
+    static Class<?> get(final Type type) {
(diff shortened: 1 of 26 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/RawTypes.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+final class RawTypes {
+
+    /** Attempts to extract the raw type class from a {@link Type}. */
+    static Class<?> get(final Type type) {
+        return TypeToken.of(type).getRawType();
+    }
+
+    private RawTypes() { }
+}
```
carterkozak, line 32: This class is new

carterkozak:
**Methodology**
Classes in `feign` were moved to our `feignimpl` package, and renamed to strip the `Conjure` class name prefix. Otherwise unmodified.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/RawTypes.java
iamdanfox: Do you think there's any realistic danger that the old `feign.Types#getRawType` method could return different results than guava's TypeToken class?

I think this repo's testing is limited to pretty vanilla usages unfortunately... just wondering if this is going to break some strange service out there...

iamdanfox (approved):
If we're confident that the feign method and the guava method behave the same, then I'm all for this! (Should allow us to switch on baseline-class-uniqueness more widely!)

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/RawTypes.java
carterkozak: I have more confidence in the google implementation than `feign.Types`. I've run OpenFeign tests with their implementation replaced with this one without failures.

PR #1056, opened 2019-04-11
[improvement] Randomize URLs every 10 minutes

ellisjoe:
## Before this PR
When we use the `PIN_UNTIL_ERROR` node selection strategy, a rolling restart of an upstream service causes all downstream clients to become pinned to the same host (this is particularly bad when there are only 2 upstream hosts).

## After this PR
==COMMIT_MSG==
Randomize the url ordering every 10 minutes to prevent clients from clustering around an upstream host during restarts/failures.
==COMMIT_MSG==

## Possible downsides?
If things are heavily cached on a single upstream host, every 10 minutes the client may start using a new host with cold caches. In general we have to deal with this problem anyway because an upstream host may become unavailable (503) or just outright die.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -54,6 +57,16 @@ private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, Duration failedUrlCoold
...
+            List<T> shuffledList = new ArrayList<>(list);
+            Collections.shuffle(shuffledList);
+            return Collections.unmodifiableList(shuffledList);
+        } else {
+            return list;
+        }
+    }
+
     /**
      * Creates a new {@link UrlSelector} with the supplied URLs. The order of the URLs may be randomized by setting
```
markelliot: update the docs to indicate URLs are randomized every 10 minutes when randomizeOrder is true

markelliot: also worth noting that except for testing we expect randomizeOrder to generally be set to true

markelliot (approved):
👍 modulo nits

carterkozak:
This breaks stateful services. Should it be configurable?

jkozlowski:
Kinda my thinking too

ellisjoe:
@carterkozak can you elaborate? all services today need to be able to handle single node failures, so in the worst case this just exacerbates any issues that clients would otherwise already experience.

carterkozak:
Let's consider a local task executing on a single node. It may not be persisting state to the database. Because this task may take >30 seconds we want to poll for results.
In the worst case today, runs with inputs which result in 11 minutes of execution may fail over to another node once or twice in the case of overloaded instances, but assuming we land on a node which continues to be healthy, it will complete successfully.

This proposal will cause us to kick off N(=retries) jobs for every attempt, and only complete successfully if we get lucky enough to randomize back to the current node.

There are a number of issues with my example but it makes closely to several of our services. We should fix those, but it will neither be quick nor easy.

jkozlowski:
For any sort of transactional service, which maintains per transaction state on the pinned node, all requests in flight will fail and have to be retried.

So this is basically a question of how much work you're wasting and retrying: if all transactions are quick and small, this is fine. But for anything that's running something a bit bigger, we'll need to redo the work.

I think it's ok, but I wouldn't be opposed to rolling out behind a flag. That said it will be hard, because it's client side.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -30,17 +31,19 @@
...
+    private final Supplier<List<HttpUrl>> baseUrls;
     private final AtomicInteger currentUrl;
     private final Cache<HttpUrl, UrlAvailability> failedUrls;
     private final boolean useFailedUrlCache;
 
-    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, Duration failedUrlCooldown) {
-        this.baseUrls = baseUrls;
+    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, boolean randomizeOrder, Duration failedUrlCooldown) {
+        this.baseUrls = Suppliers.memoizeWithExpiration(
+                () -> randomize(baseUrls, randomizeOrder), 10, TimeUnit.MINUTES);
```
carterkozak: We create most of our clients at the same time when a service is started. Assuming relatively heavy use, expiration will trigger almost exactly every ten minutes. This means every ten minutes we should expect degraded performance and higher error rates as all clients simultaneously rotate to new nodes.

ellisjoe: Updated to include some jitter in the period that we refresh at

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -28,19 +29,25 @@
...
     private final Cache<HttpUrl, UrlAvailability> failedUrls;
     private final boolean useFailedUrlCache;
 
-    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, Duration failedUrlCooldown) {
-        this.baseUrls = baseUrls;
+    private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, boolean randomizeOrder, Duration failedUrlCooldown) {
+        this.baseUrls = Suppliers.memoizeWithExpiration(
+                () -> randomize(baseUrls, randomizeOrder),
+                // Add jitter to avoid mass node reassignment when multiple nodes of a client are restarted
+                600 + ThreadLocalRandom.current().nextLong(-30, 30),
```
iamdanfox: I think we should justify the choice of 10 minutes here (especially as our read timeouts are in the same order of magnitude).  It seems to me that we just want clients to randomize more frequently than services are restarted... should we model this as roughly daily?

In the case of a two-node service, after one node switches off, the other node can be expected to receive 100% of traffic from all clients until half of this duration has elapsed.  Given that many our services already seem to handle the degenerate case of 99% load on one node and 1% on the other, maybe we could actually be more conservative with this?

What do you think about using something like 6 hours instead of 10 minutes?

iamdanfox:
@jellis thoughts on using 6 hours initially instead of 10 minutes? We can always dial this down further, just want to be cautious initially.

I think Carter's concern is quite reasonable though - currently, a stateful service running an expensive computation might know that it's not safe to shut down and could plausibly have some graceful request draining logic to ensure that a client's long-running request isn't interrupted by a shutdown.

With this new logic, clients effectively perceive a server shutdown every 10 minutes, and servers can do nothing about this.

What is the recommendation for services that fit into this category?

markelliot:
@iamdanfox I think there's a bad assumption in that reasoning that nodes are stable, and generally the PIN_UNTIL_ERROR strategy is susceptible to this line of reasoning. In practice, our services are widely resilient to this sort of problem, and where they are not resilient, should track the hard/complex fixes that @carterkozak describes as long-term goals. Practically, this PR makes a _rare_ event _typical_ and we have in this library (and in practice in our ecosystem) generally subscribed to the Chaos Engineering mentality of making rare events _normal_ so that we deal with them.

Because we're still on a journey to good, it seems reasonable that we'd feature flag this off, perhaps indirectly via a `@Deprecated` enum option of `PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE`...but either way I think having a shuffle that's on the order of our read/write timeouts in duration is overall positive for our ecosystem.

I'd also note that the long-request mode you describe is not affected by this change, this change impacts the _next_ request made, open requests will remain open until they complete or timeout, and are otherwise unaffected. (The polling that Carter describes is something we can't help with, though our general advice is that nodes of that service should probably talk something other than load-balanced HTTP, share state using one of our storage systems, share events using one of our eventing systems, or to more gracefully handle duplicative work.)

ellisjoe:
@carterkozak @iamdanfox plumbed through the configuration.

iamdanfox:
Ok so an additional piece of context is that apparently MP's heaviest users are already set to ROUND_ROBIN, so we're confident they're not relying on hitting warm caches.

If any client _is_ relying on hitting warm caches, they will need to release a version that explicitly opts out of this shuffling using `PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE`, and separately try to eliminate this requirement.

This doesn't preclude us from making improvements to how servers can rebalance load too, so :+1: from me.

PR #1062, opened 2019-04-15
[improvement] Make http2 configurable

### LINKED ISSUE #1058: Provide an option to disable http2 in client
It's possible to [create an okhttp3 client that only supports HTTP/1.1](https://stackoverflow.com/questions/41925501/java-okhttp3-only-use-http-1-1-or-http-2), but this configuration is is not exposed at the JaxRsClient level. Can we make the JaxRsClient similarly configurable to disable HTTP/2?

kmblake:
Fixes #1058 
## Before this PR

The okhttp library used by conjure-java-runtime allows configuring which protocols the client will support, which is useful in situations where a server doesn't behave correctly when HTTP/2 is enabled. However, conjure-java-runtime's ClientConfiguration does not expose this configuration.
## After this PR
==COMMIT_MSG==
Allow http2 to be disabled via code configuration
==COMMIT_MSG==
HTTP/2 remains enabled by default, but there is now a field in ClientConfiguration to disable it if necessary.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -105,6 +105,9 @@ public interface ClientConfiguration {
     /** Indicates whether client-side sympathetic QoS should be enabled. */
     ClientQoS clientQoS();
 
+    /** Indicates whether the client should support the HTTP/2 protocol. */
+    boolean http2();
+
     @Value.Check
     default void check() {
         if (meshProxy().isPresent()) {
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -48,6 +48,7 @@ public final class ClientConfigurations {
     private static final int DEFAULT_MAX_NUM_RETRIES = 4;
     private static final ClientConfiguration.ClientQoS DEFAULT_DISABLE_SYMPATHETIC_QOS =
             ClientConfiguration.ClientQoS.ENABLED;
+    private static final boolean DEFAULT_HTTP_2 = true;
 
     private ClientConfigurations() {}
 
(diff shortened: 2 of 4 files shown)


uschi2000:
This type of configurability is anti-strategy, because know clients have to know about internals of the server they wish to talk to. Can we instead fix the servers so it "just works"?

iamdanfox:
@uschi2000 I think this PR is intended to help diagnose if there's something wrong with the upstream server or not (the upstream server in this case is written in golang).

How would you feel about publishing a release candidate from this branch so that people can use it internally for debugging?

carterkozak:
Would it be reasonable to make this configurable on the upstream server instead?

jdhenke:
@carterkozak we need the performance gains of http2 and other clients seem to work fine but the one using this library does not.

uschi2000:
@iamdanfox for that type of diagnosis, why not just build a local jar on a dev machine and use that?

carterkozak:
@jdhenke I'm curious what http2 performance gains you're seeing using http/2, I'm not aware of any backend components which benefit from it outside of those with misconfigured connection pools. Not terribly relevant to this issue though.

kmblake:
Did some more debugging, and it turned out that the client and server were unable to agree on a cipher for HTTP/2. Enabling GCM cipher suites fixed it, so this change is no longer needed from my end.

schlosna:
This highlights an observability pain-point that applies to both Conjure Java clients and servers — how can one tell when you’re hitting potential badness with TLS such as:
* negotiations failing to negotiate http2 (typically due to one side missing the [TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 cipher suite requires by the spec.](https://http2.github.io/http2-spec/#rfc.section.9.2.2)
* negotiating a slow cipher suite (sometimes requires for compatibility, but may be indicator of poor config/out of date client/server version)
* not reusing sessions (likely indicator of other issues such as connection pool saturation, or failing to properly consume/reset HTTP connections before returning them to the OkHTTP pool)

One potentially valuable mechanism would be to wrap the `SSLSocketFactory` used for Conjure client and server to [add a SSLSocket handshake completed listener](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSocket.html#addHandshakeCompletedListener-javax.net.ssl.HandshakeCompletedListener-) to any created sockets and capture metrics around TLS socket creation rate, negotiated cipher suites selection frequency distribution (especially tagged by client/service version), and TLS protocol version (breakdown of TLS 1.2 vs 1.3).

PR #1066, opened 2019-04-16
Convert empty optional responses in retrofit clients

### LINKED ISSUE #668: Retrofit clients should treat HTTP 204 as empty
Our jax-rs clients already interpret a http 204 as Optional.empty, but the retrofit clients return it as `null`!

```java
    @Test
    public void retrofit2_client_can_retrieve_an_optional_empty() throws Exception {
        assertThat(client.optionalEmpty(AuthHeader.valueOf("authHeader")).execute().body())
                .isEqualTo(Optional.empty());
    }
```

```
org.junit.ComparisonFailure: 
Expected :Optional.empty
Actual   :null
 <Click to see difference>
```

pkoenig10:
Fixes #668

## Before this PR
Retrofit clients convert 204s, empty response bodies, and `null` response bodies to `null` rather than an empty optional for endpoints that return optional values.

## After this PR
Retrofit clients now convert 204s, empty response bodies, and `null` response bodies to an empty optional for endpoints that return optional values.

## Possible downsides?
This may behave incorrectly for nested optional values, but these have been disallowed since https://github.com/palantir/conjure/pull/44 because of this ambiguity.

### DIFF
--- a/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/CoerceNullValuesCallAdapterFactory.java
+++ b/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/CoerceNullValuesCallAdapterFactory.java
@@ -13,6 +13,10 @@ import java.lang.reflect.Type;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
 import java.util.Set;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


iamdanfox:
This looks like a strict improvement over the current world.  I had to do a ton of crazier things on the jaxrs side to handle aliases of optionals etc (see https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EmptyContainerDecoder.java), but I'd be OK with just merging this to unblock for now.

:+1:

carterkozak:
I think conjure-java has failing tests when we take the feign upgrade that introduced that decoder, so it's pinned to a lower version. I may be confusing that with a different decoder though.

iamdanfox:
Hmmm good point, looks like we still haven't root caused that!

pkoenig10:
The `JacksonConverterFactory` already correctly deserializes `null` as an empty optional, so we don't need the changes to `CoerceNullCollectionsConverterFactory` . Would you prefer I revert that change or keep it to be explicit here and not rely on behavior of libraries?

iamdanfox:
DW I think the consistency of having all the Optional* things together is fine!

PR #1069, opened 2019-04-18
[improvement] Correctly nest spans when tracing okhttp requests

robert3005:
## Before this PR
OkHttp: execute-enqueue would measure time in dispatcher pool and acquire-limiter

## After this PR
==COMMIT_MSG==
OkHttp tracing spans get correctly nested where acquire-limiter and okhttp: execute don't overlap and okhttp: execute-enqueue measures just the queuing time. 
==COMMIT_MSG==

## Possible downsides?
Current spans are not easy to understand - this attempts to at least make sure they measure the right thing.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DispatcherTraceTerminatingInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DispatcherTraceTerminatingInterceptor.java
@@ -24,7 +24,7 @@ public final class DispatcherTraceTerminatingInterceptor implements Interceptor 
     @Override
     public Response intercept(Chain chain) throws IOException {
         AsyncTracerTag tracerTag = chain.request().tag(AsyncTracerTag.class);
-        if (tracerTag == null && tracerTag.asyncTracer() == null) {
+        if (tracerTag == null || tracerTag.asyncTracer() == null) {
             return chain.proceed(chain.request());
         }
 
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -30,6 +30,7 @@ import com.palantir.logsafe.UnsafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalStateException;
 import com.palantir.logsafe.exceptions.SafeIoException;
 import com.palantir.tracing.AsyncTracer;
+import com.palantir.tracing.Tracer;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.net.SocketTimeoutException;
@@ -165,8 +166,15 @@ final class RemotingOkHttpCall extends ForwardingCall {
         Futures.addCallback(limiterListener, new FutureCallback<Limiter.Listener>() {
             @Override
             public void onSuccess(Limiter.Listener listener) {
-                tracer.withTrace(() -> null);
-                enqueueInternal(callback);
+                tracer.withTrace(() -> {
+                    // terminate acquire-limiter-run span
+                    Tracer.fastCompleteSpan();
+                    request().tag(AsyncTracerTag.class).setAsyncTracer(new AsyncTracer("OkHttp: execute"));
+                    enqueueInternal(callback);
+                    // Need to recreate a span to make sure withTrace will not close parent spans
+                    Tracer.startSpan("ignored-span");
+                    return null;
+                });
             }
 
             @Override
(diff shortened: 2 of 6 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DispatcherTraceTerminatingInterceptor.java
```diff
@@ -16,19 +16,18 @@
...
 import okhttp3.Interceptor;
 import okhttp3.Response;
 
 public final class DispatcherTraceTerminatingInterceptor implements Interceptor {
     @Override
     public Response intercept(Chain chain) throws IOException {
-        AsyncTracer tracerTag = chain.request().tag(AsyncTracer.class);
-        if (tracerTag == null) {
+        AsyncTracerTag tracerTag = chain.request().tag(AsyncTracerTag.class);
+        if (tracerTag == null && tracerTag.asyncTracer() == null) {
```
carterkozak: npe

robert3005: 🤦‍♂️

iamdanfox: I was surprised errorprone didn't catch this, but then I remembered we don't ahve uber's nullaway enabled!

iamdanfox:
I'm a bit late to this tracing game so I added some comments to try to capture the state of the world.  Also here is the 'before this PR' state of the world:

Screenshot 2019-04-18 at 18 16 51

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -155,14 +156,30 @@ private static ResponseBody buffer(MediaType mediaType, byte[] bodyBytes) {
...
         request().tag(ConcurrencyLimiterListener.class).setLimiterListener(limiterListener);
+
         Futures.addCallback(limiterListener, new FutureCallback<Limiter.Listener>() {
             @Override
             public void onSuccess(Limiter.Listener listener) {
-                tracer.withTrace(() -> null);
-                enqueueInternal(callback);
+                // tracer.withTrace completes the 'acquire-limiter-enqueue'
+                // tracer.withTrace starts the 'acquire-limiter-run' span
+                tracer.withTrace(() -> {
```
iamdanfox, line 169: I think before merging this it would be nice to come up with a solution that doesn't require us to include the `acquire-limiter-run` and `ignore` span hacks

robert3005: That was my thought as well - this pr is here to point to intended behaviour

carterkozak: https://github.com/palantir/tracing-java/pull/124

pkoenig10:
@iamdanfox in your example, the `execute-run` is not actually nested under `acquire-limiter-run`. The alignment occurs on the left side of the div rather than the left side of the text.

pkoenig10:
This implementation still nests all async spans directly under the caller's span. This makes it difficult to group spans for a request together when multiple requests are being made simultaneously.

I've put up an alternative and (IMO) simpler implementation in #1073. In that PR I updated the test before making the change to clearly show the span behavior before and after. I would appreciate if someone could take a look.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
diogoholanda: Why was the hack here needed?
It seems like leaving the old behavior would suffice to leave the `acquire-limiter-run` span empty and the other spans with the appropriate parents:
```
tracer.withTrace(() -> null);
enqueueInternal(callback);
```

robert3005:
Unfortunately the implementations aren't equivalent - this implementation tracks time spent in dispatcher queue while the other doesn't. I am happy to add another top level span but we originally thought it's not helpful, seems we have been wrong.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
robert3005: It won't work - the callback is executed on different thread and different trace. You need to add spans to the original trace

robert3005:
If you feel this is useful feel free to reopen - it seems that we aren't concerned with this issue right now

PR #1070, opened 2019-04-18
Rename to ServerQoS#{AUTOMATIC_RETRY, PROPAGATE_429_and_503_TO_CALLER}

iamdanfox:
I know I suggested the last `AutomaticRetryOnQoS` name, but it seemed kinda gross and too much indirection on further thought.  

I hope the new name will be more obvious

Follow-up to https://github.com/palantir/conjure-java-runtime/pull/1041

cc @jonsyu1

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -106,7 +106,7 @@ public interface ClientConfiguration {
     ClientQoS clientQoS();
 
     /** Indicates whether QosExceptions (other than RetryOther) should be propagated. */
-    AutomaticRetryOnQoS automaticRetryOnQoS();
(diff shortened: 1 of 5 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -158,6 +159,6 @@ static Builder builder() {
          * but the backend is not, it makes no sense to redirect the caller to a new backend. The client will still
          * follow redirects.
          */
-        DANGEROUS_DISABLE_AUTOMATIC_RETRY_ON_QOS;
```
iamdanfox, line 161: It's not actually dangerous in the same way that `DANGEROUS_DISABLE_SYMPATHETIC_CLIENT_QOS` is (that has the risk of overloading servers).

jonsyu1 (approved):

dansanduleac (approved):

PR #1079, opened 2019-04-24
Separate QoS responses in host metrics

pkoenig10:
## Before this PR
429 and 503 responses are included in the host metrics `get4xx` and `get5xx` timers, respectively. This can result in false positives in the `SERVICE_DEPENDENCY` health check.

## After this PR
429 and 503 are not included in the host metrics `get4xx` and `get5xx` timers. These responses are counted in a separate `getQos` timer.

### DIFF
--- a/changelog/@unreleased/pr-1079.v2.yml
+++ b/changelog/@unreleased/pr-1079.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Host metrics now measures QoS responses separately from other responses.
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/1079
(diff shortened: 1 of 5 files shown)


markelliot:
Not commenting one way or the other on whether this should be adopted, but metrics changes _must_ be reflected with readme updates.

pkoenig10:
The readme does not currently have any information about host metrics. I'm happy to add them but would prefer to do it in a separate PR.

ferozco:
The change seems sensible. I do agree with mark though, even if we didn't previously mention the host metrics in the README we should add something now so that its very clear to users what each metric means.

pkoenig10:
@ferozco Docs have been added for this change, any other blockers to merging this?

iamdanfox (approved):
LGTM, although I think you might need to manually push a changelog to get your autorelease I'm afraid

pkoenig10:
Moved to #1262

PR #1081, opened 2019-04-25
Add jaxrs and retrofit tests for query param serialization

uschi2000:


conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientQueryParamHandlingTest.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public final class JaxRsClientQueryParamHandlingTest extends TestBase {
```
carterkozak, line 34: We also have a `JaxRsClientCollectionHandlingTest`, but it only tests response body handling

carterkozak:
👍 

We may want to add a test for a more interesting type than string, that feign and retrofit wouldn't test themselves. Perhaps rid or safelong?

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientQueryParamHandlingTest.java
uschi2000: yeah, I had put it there first, put it's a weird mix, in particular with the parametrization

uschi2000:
for this test I'm mostly interested in asserting that the serialization is as `p1=v1&p1=v2` and not as `p1=[v1,v2]`, so I think string is sufficient.

PR #1085, opened 2019-04-26
[improvement] Do not retry on read timeout

### LINKED ISSUE #1084: Retrying timed out requests leads to retry storms
## What happened?
An expensive request that always times out was retried at multiple levels of the stack, causing a retry storm.

For example, the client makes an expensive request that goes through services `A -> B -> C`. `A` times out and retries the request 3 times. Each attempt causes `B` to time out and retry its request to `C` 3 times.

```
Caused by: java.net.SocketTimeoutException: {throwable3_message}
	at okio.Okio$4.newTimeoutException(Okio.java:232)
	at okio.AsyncTimeout.exit(AsyncTimeout.java:286)
	at okio.AsyncTimeout$2.read(AsyncTimeout.java:241)
	at okio.RealBufferedSource.indexOf(RealBufferedSource.java:358)
	at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:230)
	at okhttp3.internal.http1.Http1Codec.readHeaderLine(Http1Codec.java:226)
	at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:202)
```
Do not retry requests that time out. Retrying on other `IOException`s is still desirable because it eventually produces a `SafeIoException` that results in a 500. Timeout failures are different because the caller will time out before the callee can produce a 500. We do not want to set the retry count to 0 because we still want other `IOException`s to be retried.

jonsyu1:
Fixes #1084 

## Before this PR
An expensive request that always times out was retried at multiple levels of the stack, causing a retry storm.

For example, the client makes an expensive request that goes through services `A -> B -> C`. `A` times out and retries the request 3 times. Each attempt causes `B` to time out and retry its request to `C` 3 times.

## After this PR
==COMMIT_MSG==
Do not retry requests that time out on reading the response headers.
==COMMIT_MSG==

Retrying on other `IOException`s is still desirable because it eventually produces a `SafeIoException` that results in a 500. Timeout failures are different because the caller will time out before the callee can produce a 500. We do not want to set the retry count to 0 because we still want other `IOException`s to be retried.

## Possible downsides?
In some pathological cases, a timed out request could work the next try, but it's not worth the common case of causing retry storms.


markelliot:
I'm supportive of this change, but given sensitivity to the implementation and breadth of use, it might make sense to give people an escape-hatch to re-enable this behavior. Possible to plumb through a feature flag that allows consumers a way to still retry on timeouts?

jonsyu1:
The feature flag sounds good. I assume we want this to be flag to be accessible via runtime config so I'll open a PR on https://github.com/palantir/conjure-java-runtime-api to add this to `ServiceConfiguration`.

ferozco:
defer to @iamdanfox since he probably has strong opinions on this than I do

jonsyu1:
@iamdanfox ping ^

jonsyu1:
I wired up a compile time configuration for re-enabling retry on timeout. @markelliot is there anything else you'd like to see on this PR?

markelliot:
can we add a test to check that we still retry on connect timeout? We certainly want to retry when the server we are talking to does not open a connection (since, e.g., it might be down)

markelliot:
other than that, lgtm

jonsyu1:
Looks like connect timeouts also come in the form of `SocketTimeoutException` so I had to parse the exception message.

jonsyu1:
@markelliot does the test and updated exception handling look okay to you?

gatesn:
Why are we disabling all retries instead of just blacklisting the single "bad" URI?

markelliot:
@gatesn: prior to this PR, when hitting a read timeout, the library interpreted the socket exception signal as "this server is broken"; after this PR, the library will interpret the signal as "this request is broken" -- I think that's generally a move in the right direction for our empirical experience and our defaults.

okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/OkHttpClientsTest.java
```diff
@@ -509,6 +510,73 @@ public void handlesIoExceptions_obeysMaxNumRetries() throws Exception {
...
+        Call call = client.newCall(new Request.Builder().url(url + "/foo?bar").build());
+        assertThatThrownBy(() -> call.execute())
+                .isInstanceOf(SafeIoException.class)
+                .hasMessageContaining("Failed to complete the request due to an IOException")
+                .hasCauseInstanceOf(SocketTimeoutException.class);
+    }
+
+    @Test
+    public void handlesConnectTimeouts_alwaysRetry() throws IOException, InterruptedException {
+        String urlConnectTimeout = "http:/10.255.255.1";
```
markelliot: is the missing slash here intentional?

jonsyu1: No, it's not. Fixed.

jonsyu1:
Is there anything else we want to address before this merges?

markelliot (approved):

pnepywoda:
We hit this recently, is there a reason we don't retry a different host when one times out? Our situation was one where 1 host was GC-ing, high-system-load, high-thread-usage, but the other 2 were fine and would have been able to handle the request

jonsyu1:
@pnepywoda retrying a different host sounds reasonable in that scenario. Taking a step back, shouldn't the load balancer do a better job of distributing traffic and/or expensive requests?

Timed out requests often have no chance to finish before the timeout because the request itself is exorbitantly expensive. Without request cancellation, these requests can continue take up resources for each retry for the remainder of the requests. This is especially bad if the service is sharded and the requests keep getting proxied to the same overloaded shard.

I'd feel a lot more comfortable retrying a different host if we could:
1) cancel the request before retrying a different host
2) put in an escape hatch for disabling the retry on timeout for sharded services

But ultimately, the application will have a lot more context on whether or not the request has a chance at completing, so I'd prefer to leave that decision to the application.

pnepywoda:
@jonsyu1 what drove the decision to do this globally? (for all conjure clients). We're still seeing trouble when clients don't retry another (healthy-looking) node and is failing jobs.

jonsyu1:
> what drove the decision to do this globally? (for all conjure clients)

The risk of retry storms drove the decision to do this globally and our experience with bad (expensive) requests. One bad request should not be retried exponentially on timeout. One bad request should not starve other requests.

I'm open to changing the default behavior if this is a common point of friction but my concern remains.

abless:
We are seeing a case where remoting doesn't retry another host because the connection is refused. Sounds like this PR is the culprit?

EDIT: nvmd, read the code wrong, please disregard.

PR #1095, opened 2019-05-07
[fix] enforce timeout precision

dansanduleac:
## Before this PR

We require connect / read / write timeouts to have [millisecond precision](https://github.com/palantir/conjure-java-runtime/blob/6ce94d16aeb2f11cf7e72fc779d07f8929b31c69/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java#L195-L200), silently truncating them to 0 if they are smaller than 1ms.
Sadly, okhttp / feign treats a timeout of 0 as "no timeout", leading to user confusion.

## After this PR
==COMMIT_MSG==
Throw an error when attempting to create a service with a timeout with sub-millisecond precision. 
==COMMIT_MSG==

## Possible downsides?
Services configured with such a timeout will now fail to start.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -16,11 +16,13 @@
 
 package com.palantir.conjure.java.client.config;
 
-import static com.google.common.base.Preconditions.checkArgument;
+import static com.palantir.logsafe.Preconditions.checkArgument;
 
 import com.google.common.net.HostAndPort;
 import com.palantir.conjure.java.api.config.service.BasicCredentials;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.UnsafeArg;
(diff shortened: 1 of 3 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -118,6 +120,18 @@ default void check() {
...
         }
+        // Assert that timeouts are in milliseconds, not any higher precision, because feign only supports millis.
+        checkTimeoutPrecision(connectTimeout(), "connectTimeout");
+        checkTimeoutPrecision(readTimeout(), "readTimeout");
+        checkTimeoutPrecision(writeTimeout(), "writeTimeout");
+    }
+
+    default void checkTimeoutPrecision(Duration duration, String timeoutName) {
+        checkArgument(duration.minusMillis(duration.toMillis()).isZero(),
+                "Timeout should be a multiple of milliseconds",
```
iamdanfox: Timeouts with sub-millisecond precision are not supported

iamdanfox (approved):

lycarter (approved):
Thanks for the fix!

PR #1097, opened 2019-05-07
[improvement] Allow a non-global tagged metric registry

iamdanfox:
## Before this PR

All metrics (e.g. `client.response`) get written to a single shared TaggedMetricRegistry instance, supplied by tritium's `DefaultTaggedMetricRegistry.getDefault()`.

This has been tolerable so far, but means we are unable to carefully roll out a change to the underlying registry reservoir without affecting _all_ products at once, as just having tritium 0.12.0 on the classpath would impose this change on users.

## After this PR
==COMMIT_MSG==
Users can now supply a TaggedMetricRegistry rather than relying on a global singleton.

No behaviour changes for users relying on `ClientConfigurations.of()`.
==COMMIT_MSG==

e.g.

```groovy
     ClientConfiguration clientConfiguration = ClientConfiguration.builder()
                .from(ClientConfigurations.of(serviceConfig))
                .taggedMetricRegistry(registry)
                .build();
```

This is also nicer from a testing perspective, as we no longer clobber global state in each unit test.

## Possible downsides?

- ClientConfiguration has a new field now, so anyone _manually_ building these field-by-field would experience runtime failures until they also initialize the taggedMetricRegistry
- The four dispatcher-related metrics are now scoped under the tag `from: DispatcherMetricSet`, which is a bit pointless (just tritium's API).  It shouldn't increase our DD spend because this metrics are already uniquely identified by the product version!
- Strange things could happen if users provide a TaggedMetricRegistry with an odd equals/hashcode implementation.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -23,6 +23,7 @@ import com.palantir.conjure.java.api.config.service.BasicCredentials;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
 import java.net.ProxySelector;
 import java.time.Duration;
 import java.util.List;
@@ -103,12 +104,14 @@ public interface ClientConfiguration {
      */
     Duration backoffSlotSize();
 
-
     /** Indicates whether client-side sympathetic QoS should be enabled. */
     ClientQoS clientQoS();
 
     /** Indicates whether QosExceptions (other than RetryOther) should be propagated. */
     ServerQoS serverQoS();
+
+    /** Both per-request and global metrics are recorded in this registry. */
+    TaggedMetricRegistry taggedMetricRegistry();
 
     @Value.Check
     default void check() {
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -21,6 +21,7 @@ import com.google.common.net.HostAndPort;
 import com.palantir.conjure.java.api.config.service.ProxyConfiguration;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.config.ssl.SslSocketFactories;
+import com.palantir.tritium.metrics.registry.DefaultTaggedMetricRegistry;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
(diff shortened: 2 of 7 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -77,6 +78,7 @@ public static ClientConfiguration of(ServiceConfiguration config) {
                 .backoffSlotSize(config.backoffSlotSize().orElse(DEFAULT_BACKOFF_SLOT_SIZE))
                 .clientQoS(CLIENT_QOS_DEFAULT)
                 .serverQoS(PROPAGATE_QOS_DEFAULT)
+                .taggedMetricRegistry(DefaultTaggedMetricRegistry.getDefault())
```
carterkozak, line 81: May want to upgrade and use the non-deprecated method

iamdanfox: Aha so I actually don't want to impose tritium 0.12.0 on everyone _just_ yet because it changes the global singleton from EDR -> sliding window, and the intention of this PR is to let us roll out that change in a well communicated way!

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -103,13 +104,15 @@
...
 
-
     /** Indicates whether client-side sympathetic QoS should be enabled. */
     ClientQoS clientQoS();
 
     /** Indicates whether QosExceptions (other than RetryOther) should be propagated. */
     ServerQoS serverQoS();
 
+    /** Both per-request and global metrics are recorded in this registry. */
+    TaggedMetricRegistry taggedMetricRegistry();
```
iamdanfox, line 114: Feels a tiny bit weird to put this here as most the of the other fields are pure data, but this isn't the first 'behavioury' object as we've already got a `java.net.ProxySelector`, `SSLSocketFactory` and `X509TrustManager`.

dansanduleac (approved):

uschi2000:
I don't think it's a good idea to mix injection and global singletons. So far the contract has been to just grab the global singleton (which is fine), but now you have to always first check whether there is an injected instance flying around somewhere?

iamdanfox:
Not sure what you mean by "check whether there is an injected instance flying around"?

uschi2000:
Say I want to log some metrics deep down in some class in this library. By convention, I would grab a `DefaultTaggedMetricRegistry.getDefault()` and add my metrics. However, that is no longer the correct thing to do after this change and the only way to realize is by careful code review by something who knows about this change and remembers it.

iamdanfox:
I don't think grabbing the DefaultTaggedMetricRegistry has ever really been a best practise.  I don't think there's too much danger of people adding that as it is pretty much impossible to unit test (both of the other usages of tagged metric registries already had an injected registry).

uschi2000:
OK, can we then deprecate DefaultTaggedMetricRegistry.getDefault()?

PR #1098, opened 2019-05-07
Set higher cooldown time by default

hsaraogi:
## Before this PR
Say we have 3 nodes of service A and service B, and one node of service B is in a state that it is accepting connections but not processing requests, all requests will timeout after [IDLE_TIMEOUT which is 60s](https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt#L256).
So for the first 60s, serviceA sends 1/3rd requests to all three nodes, then one node is blacklisted for cooldown-interval and again service A starts sending 1/3rd requests to all nodes. One service set the cooldown interval to be 10s and a lot of the others set it to less that a minute, thus a lot of the requests will fail. If the cooldown time is set to a higher value, like 5mins, this would be better.

We need to follow-up with PRs so that products remove the hard-coded default unless they have a good reason to not do so.
## After this PR
Blacklist nodes that caused an error for 5mins to prevent a lot requests from hitting a bad node.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -41,7 +41,7 @@ public final class ClientConfigurations {
     private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(5);
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(5);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
-    private static final Duration DEFAULT_FAILED_URL_COOLDOWN = Duration.ZERO;
+    private static final Duration DEFAULT_FAILED_URL_COOLDOWN = Duration.ofMinutes(5);
     private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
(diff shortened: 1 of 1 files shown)


markelliot:
I'm pretty sure we only read this value if the mode is switched to ROUND_ROBIN, and in the place where you do that, you're required to set failedUrlCoolDown explicitly.

hsaraogi:
> set failedUrlCoolDown explicitly

But products usually copy this config or choose a value out of thin air. We had this as 10s for example. Making this default as 5mins does no harm if using PIN_UNTIL_ERROR and can prevent outages if the client is ROUND_ROBIN.

markelliot:
It seems like there's something formulaic we could do to either set or verify failedUrlCoolDown, could we do that instead of changing an unused default?

hsaraogi:
I could change the check to verify that the failed cooldown is greater than 5mins instead of checking that the value is positive if that makes sense.

markelliot:
I guess we could be more formulaic? Sounds like you're proposing that failedUrlCoolDown is equal to the read timeout?

hsaraogi:
yeah, I was originally proposing making it greater than 60s (the [Idle timeout](https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt#L256)), so that we have a very small percentage of requests hit a node that is not responding, but think making the cooldown greater than read timeout be even more robust.

PR #1101, opened 2019-05-09
[improvement] make sure failedUrlCooldown is at least readTimeout

hsaraogi:
## Before this PR
Say we have 3 nodes of service A and service B, and one node of service B is in a state that it is accepting connections but not processing requests, all requests will timeout after IDLE_TIMEOUT which is 60s.
So for the first 60s, serviceA sends 1/3rd requests to all three nodes, then one node is blacklisted for cooldown-interval and again service A starts sending 1/3rd requests to all nodes. One service set the cooldown interval to be 10s and a lot of the others set it to less that a minute, thus a lot of the requests will fail. If the cooldown time is set to a higher value, like 5mins, this would be better.

## After this PR
Do not send any requests to nodes that caused a connection error for 5mins to prevent a lot requests from hitting a bad node.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -115,8 +115,9 @@ public interface ClientConfiguration {
             checkArgument(uris().size() == 1, "If meshProxy is configured then uris must contain exactly 1 URI");
         }
         if (nodeSelectionStrategy().equals(NodeSelectionStrategy.ROUND_ROBIN)) {
-            checkArgument(!failedUrlCooldown().isNegative() && !failedUrlCooldown().isZero(),
-                    "If nodeSelectionStrategy is ROUND_ROBIN then failedUrlCooldown must be positive");
+            checkArgument(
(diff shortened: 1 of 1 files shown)


pkoenig10:
Note that if the failedUrlCooldown is equal to the readTimeout, then we are still sending requests to a possibly bad node 50% of the time.

hsaraogi:
Not really, requests to the node will timeout after 60s due to the idle timeout. and then we dont send any requests for 5 minutes, so its 1/6~16.7% requests to the bad node.

pkoenig10:
Sure, if you assume that there is a one minute idle timeout that will get hit first. I'm not entirely sure how this read timeout and the default okio idle timeout relate. I've definitely seen requests fail after the read timeout so I'm curious why those didn't hit the idle timeout first.

pkoenig10:
Looking at the code, the [one minute idle timeout in `AsyncTimeout`](https://github.com/square/okio/blob/okio-parent-1.17.2/okio/src/main/java/okio/AsyncTimeout.java#L50) is just the default value. The raw socket timeout and okio source timeouts are [set to the read timeout](https://github.com/square/okhttp/blob/parent-3.13.1/okhttp/src/main/java/okhttp3/internal/connection/RealConnection.java#L526-L528).

PR #1104, opened 2019-05-13
Test failed URLs before allowing them to be used again

pkoenig10:
Another way to mitigate the problem described in #1101.

## Before this PR
When a URL expires from the failed URL cache we immediately began using it again for requests. However the first request made to a previously failed URL has a much higher likelihood of failing than a typical request. Furthermore, we may make many requests to this URL before we can determine the host is still unavailable, all of which will fail.

## After this PR
When a URL fails, we limit its use to at most one request per `failedUrlCooldown`. If a request to a previously failed URL succeeds, then we mark the URL as available and it is used normally again.

## Possible downsides?

If a client using the `PIN_UNTIL_ERROR` node selection strategy has all of its URLs marked as failed, then requests will effectively be round robined until one URL is successful, at which point all requests will use that URL. While this is a behavior change, it seems like the more correct behavior.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -176,30 +177,40 @@ private static String switchWsToHttp(String url) {
         return redirectTo(current, httpUrls.get((currentUrl.get() + 1) % httpUrls.size()));
     }
 
+    @Override
+    public void markAsSucceeded(HttpUrl failedUrl) {
+        if (useFailedUrlCache) {
```
pkoenig10, line 202: Can we just remove this? Updating this cache should be pretty inexpensive

dansanduleac: seems fine, it's just disabling it in cases where it doesn't make sense.
Although it's a bit weird we accept a negative value here...

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -176,30 +177,40 @@ private static String switchWsToHttp(String url) {
...
+            indexFor(failedUrl).ifPresent(failedUrls::invalidate);
```
dansanduleac: I wonder if O(1) -> O(n) here might be a problem since this happens on every successful request. I think at our current load though this will be negligible.

pkoenig10: Yeah this is not ideal, but I wasn't sure how else to get the desired behavior.

dansanduleac:
👍 ok, makes sense
I put in a change to use a `Clock` instead of `Instant.now` as I'm not a fan of Thread.sleep in tests if it can be avoided.

pkoenig10:
👍 LGTM

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -41,10 +44,12 @@
 
     private final Supplier<List<HttpUrl>> baseUrls;
     private final AtomicInteger currentUrl;
-    private final Cache<HttpUrl, UrlAvailability> failedUrls;
+    private final LoadingCache<Integer, Instant> failedUrls;
```
iamdanfox: Doesn't the `List<HttpUrl>` get reshuffled (L57)? using indexes here would mean they end up pointing at something else after the resuffle.

pkoenig10: You're right, although this creates potential race conditions at multiple places in this class where we compute a URL index and then get the URL using potentially different orderings.

It seems to me that instead of randomizing the URL list we should just randomly select a new current URL every 10 minutes. Would you be open to that change in a follow up?

dansanduleac: I'd like to see a follow-up that saves the base url into the request as a tag, so we wouldn't have to do any of this indexOf nonsense at all.

pkoenig10:
@iamdanfox @dansanduleac can I get a +1?

dansanduleac:
@pkoenig10 Sorry we seemed to forget about this PR. I've rebased it onto the most recent changes to UriSelectorImpl.

However, since this involves some extra complexity, I'd like to understand how often was this case actually encountered since making this PR. The default connect timeout is 10s so it's quite likely that bad requests won't take more than that, unless they do somehow manage to connect but the server is then very busy and fails to respond for a long time. Have you seen this be a problem lately?

Also, chatted with @ferozco and we realized that this is not actually behaving as intended for PIN_UNTIL_ERROR clients. Each new Call will select the URL via `redirectToCurrent()` which doesn't look at `failedUrls`, so even after refreshing a URL as cooling down, new requests would continue to use that URL until a failure / 503 forces a `redirectToNext` call.
 If we do want to merge this in the end, then I think `redirectToCurrent()` should also use `getNext` if the current node is in cooldown mode.

pkoenig10:
The issue that prompted this was particularly odd. We were sending requests and the connection was accepted but the server either never responded or was taking over 60s to respond. The root cause was that the server had gotten into some bad state and wasn't properly handling requests. While this scenario is unlikely, this still seems like a reasonable improvement.

> If we do want to merge this in the end, then I think `redirectToCurrent()` should also use `getNext` if the current node is in cooldown mode.

Can we do this in a follow up? The behavior of `PIN_UNTIL_ERROR` will be the same as it was before this PR but we get the improvements for `ROUND_ROBIN` clients.

dansanduleac:
Sorry for the delay here. I think that for consistency, and since we consider this to be a positive change no matter what `NodeSelectionStrategy` you use, we'd like the same behaviour in both cases.

pkoenig10:
@dansanduleac I've updated the behavior to be consistent, this should be ready for another review

dansanduleac:
So, I added tests to validate that we are actually reverting to round robin in the default PIN_UNTIL_ERROR case if all nodes have been marked as failed, and it turns out it wasn't doing that.
This is because the new method `redirectToFirstNotFailed` wasn't flexible enough to force-select the *next* node rather than the start node (in the case of `redirectToCurrent` but all nodes are marked as failed), so I brought back the `getNext` method (renamed to `getNextHealthy`) and then inlined the 3 calls to `redirectToFirstNotFailed` (which are now 2 instructions each) to account for this difference.

dansanduleac:
👍

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -64,11 +68,11 @@ private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, boolean reshuffle, Dura
...
         this.lastBaseUrl = new AtomicReference<>(baseUrls.get(0));
 
-        long coolDownMillis = failedUrlCooldown.toMillis();
+        this.clock = clock;
         this.failedUrls = Caffeine.newBuilder()
+                .executor(MoreExecutors.directExecutor())
                 .maximumSize(baseUrls.size())
-                .expireAfterWrite(coolDownMillis, TimeUnit.MILLISECONDS)
-                .build();
+                .build(key -> clock.instant().plus(failedUrlCooldown));
```
iamdanfox: This cache looks quite strange to me now - I don't really understand why it's desirable to have a fixed expiry time for every entry?? What if a UrlSelectorImpl isntance lives for a long time?

pkoenig10: > I don't really understand why it's desirable to have a fixed expiry time for every entry??

Isn't this exactly the meaning of `coolDownMillis`? What are you suggesting as an alternative?

pkoenig10: Right, we don't want this anymore otherwise we might mark previously failed urls as healthy again. The maximum size should be sufficient to limit the size of this cache.

pkoenig10: This cache no longer does any expiry. We also limit the size of the cache to the size of the base urls list. What's the concern here?

iamdanfox:
Honestly I'm pretty uncomfortable about changing the current UrlSelectorImpl in its current state.  The methods on the interface barely make sense to me, and there's now so many different possible codepaths inside this object depending on what order they're called, whether the failedUrlCooldown is provided etc etc.

The fact that we have a NodeSelectionStrategy but it doesn't correlate to different implementations of UrlSelectorImpl feels a bit weird to me too.

pkoenig10:
I agree that this interface is pretty non-intuitive. I considered refactoring the `UrlSelectorImpl` to contain the node selection strategy as part of this PR.

I'm happy to do the work to refactor this to either take a node selection strategy or have different implementations. We can do it before, in, or after this PR.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
dansanduleac: Well in any case this doesn't need to be a cache, turns out. So I've made it a map.

dansanduleac:
Quick update: I think we're still not super happy with the current implementation here, seems like the UrlSelector methods are too confusing (too similar but subtly different) and we might be able to simplify the "API" surface there.
Was going to iterate further but @iamdanfox wanted to prototype implementing this behaviour change in a different way, so blocking on that for the time being.

iamdanfox (approved):
Ok so I'm still uncomfortable with the whole UrlSelector interface, but I'm gonna drop an approve on this because @dansanduleac is confident it's better than the current world.

PR #1107, opened 2019-05-15
[improvement] Upgrade concurrency-limits to 0.2.2

jonsyu1:
## Before this PR
Instances of `SimpleLimiter` lack the in flight metric. Java projects built on top of `conjure-java-runtime` that force a newer version of `conjure-java-runtime` to pick up this feature inadvertently pick up behavioral changes because the `AIMDLimit` defaults have changed.

## After this PR
==COMMIT_MSG==
Upgrade concurrency-limits dependency without changing behavior. AIMDLimit defaults have changed since 1.0.9 so we explicitly configure the old behavior. This partially future-proofs us from future changes to AIMDLimit.
==COMMIT_MSG==

## Possible downsides?
This change comes with the normal risks of upgrading dependencies.


ferozco:
👍 Looking through the changes, this seems like only this [PR](https://github.com/Netflix/concurrency-limits/commit/d3c15f2866504abc0c9a8e8bbc0d393f0b3bd2ce) would have any impact and the defaults being applied effectively revert the behaviour change

PR #1110, opened 2019-05-23
[improvement] ConcurrencyLimiter acquire queue histogram

gatesn:
## Before this PR
It's hard to tell when we're going up against our concurrency limits. All we know is that there's a "slow" acquire.

## After this PR
Histogram of queue sizes across all requests.

## Possible downsides?
Without per-endpoint metrics it will still be quite hard to see what's happening. But the histogram should give us an indication.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.okhttp;
 
+import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ThreadWorkQueue.java
```diff
@@ -40,6 +40,10 @@ boolean isEmpty() {
         return queuedRequests.isEmpty();
     }
 
+    int size() {
+        return queuedRequests.size();
```
ellisjoe, line 44: this gives you back the number of threads, not the number of queued requests

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -208,6 +213,9 @@ public void onDropped() {}
         }
 
         synchronized void processQueue() {
+            // Too many metrics to emit per endpoint, so we record as histogram instead
+            acquireQueue.update(waitingRequests.size());
```
ellisjoe, line 217: so if we end up with a single endpoint that is fast, it'll contribute far more frequently to this histogram which is just gonna skew things here

ellisjoe: I wonder if you just want to register a gauge which sums all of the queued requests across endpoints and threads

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ThreadWorkQueue.java
iamdanfox: Good spot - looks like we'd need to recurse through the map and sum up the size of all the per-thread queues.

iamdanfox:
I'd like to close out because I hope the improved tracing spans should give you clearer information about exactly how many permits are used/max for a single endpoint.

PR #1111, opened 2019-05-23
[improvement] Bump initial concurrency limits form 10 -> 64 to match OkHttp dispatcher limits

gatesn:
## Before this PR
On startup, the server is unnecessarily rate-limited on outbound requests until it's AIMD limiter has sufficiently increased.

## After this PR
The initial concurrency limit matches what OkHttp's dispatcher would allow anyway. If this is too much for the server the multiplicative-decrease should backoff aggressively.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -107,7 +107,7 @@ final class ConcurrencyLimiters {
     Limit newLimit() {
         return new ConjureWindowedLimit(AIMDLimit.newBuilder()
                 .timeout(Long.MAX_VALUE, TimeUnit.NANOSECONDS)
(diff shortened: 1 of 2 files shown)


iamdanfox:
This should only have an impact when a service is first starting up, and during the lifetime of a service the AIMD limiter should tune to itself to the same eventual value.

Our atlas-based storage service will already give people 429s if they send a quick burst after startup, so I don't think we're at risk of overloading that. The only danger really is if an upstream service doesn't have any rate limiting.

Overall, I think net positive :+1:

iamdanfox:
Talking to @j-baker, it seems we might have merged this prematurely, as this will more likely result in us approaching the limit from above, crashing through it (likely all the way to 1), and then slowly climbing back up to the true limit.

What we didn't account for is that climbing up is actually quite slow, because the ConjureWindowedLimit can increase at most once per second.

EDIT gonna revert this now we can hopefully see exactly what's going on after adding this: https://github.com/palantir/conjure-java-runtime/pull/1112

PR #1112, opened 2019-05-23
[fix] Debug log line for current limit

iamdanfox:
## Before this PR

It's hard to figure out what concurrency limiters are actually doing.  Unfortunately adding per-endpoint metrics to DD is a bit expensive.

## After this PR
==COMMIT_MSG==
New debug log line from origin `com.palantir.conjure.java.okhttp.ConcurrencyLimiters` with message `Limit`.
==COMMIT_MSG==

## Possible downsides?

People have to know to enable debug for this origin, so you can't look at the limits retrospectively.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -125,7 +125,7 @@ final class ConcurrencyLimiters {
         if (!useLimiter) {
             return NoOpConcurrencyLimiter.INSTANCE;
         }
-        Supplier<Limiter<Void>> limiter = () -> SimpleLimiter.newBuilder().limit(newLimit()).build();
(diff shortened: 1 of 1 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -209,6 +215,10 @@ public void onDropped() {}
 
         synchronized void processQueue() {
             while (!waitingRequests.isEmpty()) {
+                log.debug("Limit",
+                        SafeArg.of("limit", limiter.getLimit()),
+                        safeArgMethod,
+                        safeArgPathTemplate);
```
dansanduleac, line 221: should we put this under a `log.isDebugEnabled()` guard so we don't create an object each time?

iamdanfox: I dunno man there are so many objects that get allocated when you make a HTTP request already, i've already pulled out the method and the path template because they're constant, I think `SafeArg.of("limit", limiter.getLimit())` isn't thaaat bad

dansanduleac (approved):

PR #1114, opened 2019-05-23
[fix] Bump tracing

dansanduleac:
## Before this PR

We depend on tracing-java 2.10.1 which (along with 2.10.0) have malformed POMs (broken by https://github.com/palantir/tracing-java/pull/107, fixed by https://github.com/palantir/tracing-java/pull/133) - the `provided` dependencies didn't have any version.
This in turn means that tools that transitively resolve maven coordinates (e.g. aether) end up failing to fully resolve all of the transitive dependencies of conjure-java-runtime coordinates.

## After this PR
==COMMIT_MSG==
Bump tracing to avoid depending on coordinate with malformed pom.
==COMMIT_MSG==


ferozco:
👍

PR #1115, opened 2019-05-24
[fix] log hosts on retry

pnepywoda:
## Before this PR
Didn't log hostnames on retry

## After this PR
Logs hostnames on retry

## Possible downsides?
Hostnames aren't considered unsafe, so shouldn't be any downside.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -207,6 +207,8 @@ public void onFailure(Call call, IOException exception) {
 
                 log.info("Retrying call after failure",
                         SafeArg.of("backoffMillis", backoff.get().toMillis()),
+                        SafeArg.of("originalHost", request().url().host()),
+                        SafeArg.of("redirectToHost", redirectTo.get().host()),
```
pkoenig10, line 211: This should also be added to the 2 other places we retry requests in this class

iamdanfox (approved):

PR #1116, opened 2019-05-24
Initialize RemoteException cause for Retrofit clients

pkoenig10:
## Before this PR
A `RemoteException` created when using a Retrofit clients does not have any cause, making it difficult to know the request that caused the exceptions. Here is an example stacktrace using a Retrofit client:
```
com.palantir.conjure.java.api.errors.RemoteException: {throwable0_message}
	at com.palantir.conjure.java.okhttp.RemoteExceptionResponseHandler.handle(RemoteExceptionResponseHandler.java:62)
	at com.palantir.conjure.java.okhttp.RemotingOkHttpCall$3.onResponse(RemotingOkHttpCall.java:250)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:216)
	at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at com.palantir.tracing.DeferredTracer.withTrace(DeferredTracer.java:115)
	at com.palantir.tracing.Tracers$TracingAwareCallable.call(Tracers.java:343)
	at com.palantir.tracing.WrappingExecutorService.lambda$wrapTask$0(WrappingExecutorService.java:67)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(Thread.java:834)

```

## After this PR
A `RemoteException` created when using a Retrofit clients will have a cause whose exception is the stacktrace of the call that enqueued the request.

### DIFF
--- a/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncException.java
+++ b/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncException.java
@@ -1,0 +1,21 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.conjure.java.client.retrofit2;
+
+final class AsyncException extends RuntimeException {
+
+    AsyncException() {}
+}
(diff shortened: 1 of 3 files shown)


iamdanfox (changes requested):
Initializing a `new AsyncException()` every single time we make a network call will slow down all retrofit services slightly, as it has to walk the stack to figure out where it came from. I'm concerned that if we add this to every single call it might actually be net-negative.

Strawman alternative to try and salvage some perf:

- only instantiate a `new AsyncException()` if we've observed a failure for this specific endpoint in the last 30 seconds,
- also keep some global rate limiter of how many times we think it's acceptable to build a whole stacktrace per minute (maybe just 60?) and then if we're on a particularly hot codepath, the majority of calls wouldn't get a full stacktrace built.

Downside is that obviously if we make 1000 calls and only one of them fails, then we're gonna have to be pretty lucky to actually get a full stacktrace.

pkoenig10:
> if we make 1000 calls and only one of them fails

In practice, I think this type of behavior is what happens the majority of the time. So trying to be smarter here will result in not having the stacktraces in many cases where we want it.

iamdanfox:
Closing out because I don't think we can take the perf hit with this approach - happy to try something else though!

PR #1119, opened 2019-06-04
Hosts are unsafe

ferozco:
@gatesn raised a valid point that #1115 could leak unsafe information since you may be given external (unsafe) URIs. 

We can either log the values as unsafe params or revert the PR entirely

@pnepywoda 	for SA


pnepywoda:
Log IP addresses?

pnepywoda:
https://github.com/palantir/conjure-java-runtime/pull/1120

ferozco:
Closing in favour of #1120

PR #1120, opened 2019-06-04
log ip instead of hostname

pnepywoda:
Hostnames can be unsafe to log. Log IP address instead


iamdanfox (changes requested):
I think this might not actually be necessary - internally hostnames are always safe (appear in request logs etc). Can we clarify in exactly what scenarios they can be unsafe?

pnepywoda:
@gatesn & @ferozco can you answer Dan?

PR #1127, opened 2019-06-11
[fix] Conjure ObjectMapper now follows spec serialising optional fields

JacekLach:
## Before this PR

`OptionalExample.empty()` serializes as `{"value"=null}`

https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#53-named-types

==COMMIT_MSG==
Conjure ObjectMappers no longer produce JSON object values of `null` when serialising absent fields on conjure objects.
==COMMIT_MSG==

## Possible downsides?
Breaks some tests where people wrote generated maps on disk, I guess.


dansanduleac:
👍

ferozco (changes requested):
Thinking about this more, I don't think we should merge this change since it will break users that return responses with null values. This typically happens when endpoints return raw data (rows in a dataset, values in a timeseries, etc) rather than structure conjure responses.

JacekLach:
Why are people using a conjure mapper to serialise non-conjure responses?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -97,6 +98,7 @@ public static ObjectMapper newCborServerObjectMapper() {
      */
     public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
         return mapper
+                .setSerializationInclusion(JsonInclude.Include.NON_NULL)
```
pkoenig10, line 101: Instead we should probably use `NON_ABSENT` and remove the `configureAbsentsAsNulls` config below. This is what the docs recommend:
https://github.com/FasterXML/jackson-modules-java8/blob/jackson-modules-java8-2.9.9/datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/Jdk8Module.java#L21-L23

pkoenig10: We should also probably just use `setDefaultPropertyInclusion(...)` instead. `setSerializationInclusion` is a just convenience method that calls `setPropertyInclusion` which is deprecated in favor of `setDefaultPropertyInclusion`.

PR #1131, opened 2019-06-18
[improvement] Jackson encoder avoids intermediate String request body

schlosna:
## Before this PR

Feign's JacksonEncoder would convert object to String, then to UTF-8 bytes.

## After this PR
==COMMIT_MSG==
Serialize Jackson request object directly to UTF-8 byte array without intermediate String representation.

Introduces local Jackson encoder until https://github.com/OpenFeign/feign/pull/989 is upstreamed.

See https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway
==COMMIT_MSG==

## Possible downsides?
We maintain this code until fix lands in upstream version we can pick up.

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/ConjureFeignJacksonEncoder.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/ConjureFeignJacksonEncoder.java
@@ -1,0 +1,50 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.client.jaxrs;
+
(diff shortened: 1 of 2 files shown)


schlosna:
@carterkozak for SA

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/ConjureFeignJacksonEncoder.java
```diff
@@ -0,0 +1,50 @@
...
+
+    ConjureFeignJacksonEncoder(ObjectMapper mapper) {
+        this.mapper = mapper;
+    }
+
+    @Override
+    public void encode(Object object, Type bodyType, RequestTemplate template) {
+        try {
+            JavaType javaType = mapper.getTypeFactory().constructType(bodyType);
+            template.body(mapper.writerFor(javaType).writeValueAsBytes(object), StandardCharsets.UTF_8);
```
iamdanfox, line 45: Eyeballing this, I wondered whether the plain `mapper.writeValueAsBytes()` might avoid allocating the objectwriter, but I guess the reflection is more costly.

The PR as it is now is consistent with what the upstream feign jackson encoder used to do, so I guess they considered this!

```java
template.body(mapper.writerFor(javaType).writeValueAsString(object));
```

iamdanfox (approved):
Thanks @schlosna!

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/ConjureFeignJacksonEncoder.java
carterkozak: This implementation bounds the writer to the type specified on our interface, rather than Object. It's unlikely that it makes a difference for our use in conjure, but it does provide additional specificity.

carterkozak:
thanks!

carterkozak:
released in https://github.com/palantir/conjure-java-runtime/releases/tag/4.29.0

PR #1137, opened 2019-07-01
[improvement] Release concurrency limiters on response

gatesn:
## Before this PR
Limiters are only released once the response body has been closed. For large requests, such as those we sometimes see from streaming blob-store-esque interfaces it's easy to run into concurrency limits even though the server is responding in good time.

## After this PR
Limits are released as soon as the response is returned, not waiting for the body to be closed.

## Possible downsides?
Instead of using concurrency limiters (endpoint-level) for client-side blocking of requests, we fall back on either the OkHttp dispatcher limits (host-level), or server-side QoS.


ferozco (commented):
I think this change may be undesirable since we would now start marking streaming response as a success once the first byte is received instead of waiting for the entire response

dansanduleac:
The whole point of the concurrency limiter is to model how many resources we are using on a server, and a streamable response definitely still holds on to a server thread until it's done writing, since the server is not using NIO + selectors to fulfil these responses, but writing them in one blocking invocation.
Unless we decide the whole premise of the concurrency limiter is broken, I'd prefer keeping the existing behaviour and not special-case streaming endpoints.

PR #1142, opened 2019-07-10
Clean up URL selector logic

dansanduleac:
## Before this PR

A lot of the UrlSelectorImpl methods manipulate multiple mutable fields, making it hard to follow what's going on. For instance, both the set of URIs (which can change non-deterministically due to the periodic re-shuffling) and the current URI index are accessed multiple times through multiple nested methods.

Additionally, it's not clear when you're dealing with base urls vs urls that are supposed to start with base URLs. As a result, a few places deal with attempting to find a baseUrl of a value that in most cases is already a baseUrl, leading to confusion as to what values might be passed in.

Finally, the current index can become stale after a reshuffling.

## After this PR
==COMMIT_MSG==
Clean up UrlSelector logic so it's easier to follow.

* `currentUrl` is not an integer anymore but a HttpUrl, and now named `lastBaseUrl` to clarify that it's meant to be a base url, and to disambiguate from URL parameters named `current` in a bunch of methods. Its purpose is to track last the base url used by the last redirect
* `redirectTo(HttpUrl, HttpUrl)` now makes more assumptions about its inputs - namely that the directBaseUrl is actually a baseUrl, so it's not necessary to call baseUrlFor on it anymore
* `baseUrls.get()` is only called on public methods and then passed in to nested methods, making it easier to reason that it doesn't change throughout a given call stack, independently of the reshuffling cache
* `indexFor` and `getNext` are now static
==COMMIT_MSG==

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
@@ -153,9 +153,10 @@ final class UrlSelectorImpl implements UrlSelector {
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
-        // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
         List<HttpUrl> httpUrls = baseUrls.get();
-        int currentIndex = getCurrentIndex(existingUrl, httpUrls);
+        // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
+        Optional<Integer> existingUrlIndex = indexFor(existingUrl, httpUrls);
+        int currentIndex = existingUrlIndex.orElseGet(() -> getCurrentIndex(httpUrls));
 
         Optional<HttpUrl> nextUrl = getNext(currentIndex, httpUrls);
         if (nextUrl.isPresent()) {
@@ -192,21 +193,6 @@ final class UrlSelectorImpl implements UrlSelector {
                     failedUrls.put(baseUrls.get().get(index), UrlAvailability.FAILED)
             );
         }
-    }
-
-    /**
-     * Best-effort to find the index of the last used base URL, by going through the following cases in order.
-     * <ul>
-     *     <li>The index of a baseUrl from {@code httpUrls} that is a path-prefix of {@code justAttemptedUrl}</li>
-     *     <li>The index of {@link #currentBaseUrl}, which is expected to exist in {@code httpUrls}</li>
-     * </ul>
-     *
-     * @param justAttemptedUrl  URL that was just attempted
-     * @param httpUrls          the current list of base URLs
-     */
-    private int getCurrentIndex(HttpUrl justAttemptedUrl, List<HttpUrl> httpUrls) {
-        Optional<Integer> existingUrlIndex = indexFor(justAttemptedUrl, httpUrls);
-        return existingUrlIndex.orElseGet(() -> getCurrentIndex(httpUrls));
     }
 
     /**
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
@@ -119,7 +119,7 @@ final class UrlSelectorImpl implements UrlSelector {
     }
 
     private Optional<HttpUrl> redirectTo(HttpUrl current, HttpUrl redirectBaseUrl) {
-        Optional<Integer> baseUrlIndex = indexFor(redirectBaseUrl);
+        Optional<Integer> baseUrlIndex = indexFor(redirectBaseUrl, baseUrls.get());
         baseUrlIndex.ifPresent(currentUrl::set);
 
         return baseUrlIndex
@@ -145,7 +145,7 @@ final class UrlSelectorImpl implements UrlSelector {
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
         // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
-        Optional<Integer> existingUrlIndex = indexFor(existingUrl);
+        Optional<Integer> existingUrlIndex = indexFor(existingUrl, baseUrls.get());
 
         int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
 
@@ -179,7 +179,7 @@ final class UrlSelectorImpl implements UrlSelector {
     @Override
     public void markAsFailed(HttpUrl failedUrl) {
         if (useFailedUrlCache) {
-            Optional<Integer> indexForFailedUrl = indexFor(failedUrl);
+            Optional<Integer> indexForFailedUrl = indexFor(failedUrl, baseUrls.get());
             indexForFailedUrl.ifPresent(index ->
                     failedUrls.put(baseUrls.get().get(index), UrlAvailability.FAILED)
             );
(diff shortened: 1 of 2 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -40,11 +41,14 @@
...
 
     private final Supplier<List<HttpUrl>> baseUrls;
-    private final AtomicInteger currentUrl;
+    private final AtomicReference<HttpUrl> currentBaseUrl;
     private final Cache<HttpUrl, UrlAvailability> failedUrls;
     private final boolean useFailedUrlCache;
 
     private UrlSelectorImpl(ImmutableList<HttpUrl> baseUrls, boolean reshuffle, Duration failedUrlCooldown) {
+        Preconditions.checkArgument(!baseUrls.isEmpty(), "Must specify at least one URL");
+        Preconditions.checkArgument(!failedUrlCooldown.isNegative(), "Cache expiration must be non-negative");
```
dansanduleac, line 50: I just moved these at the top, makes reasoning about hte rest of the constructor easier

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
 
+    /**
+     * Attempt to redirect to the given redirectUrl, which could be a longer URL than just a base path, in which
+     * case it will first be matched to a baseUrl from {@link #baseUrls}.
+     */
     @Override
-    public Optional<HttpUrl> redirectTo(HttpUrl current, String redirectBaseUrl) {
-        return redirectTo(current, HttpUrl.parse(redirectBaseUrl));
+    public Optional<HttpUrl> redirectTo(HttpUrl current, String redirectUrl) {
+        return baseUrlFor(HttpUrl.parse(redirectUrl), baseUrls.get()).flatMap(baseUrl -> redirectTo(current, baseUrl));
```
dansanduleac, line 124: Basically pulled up this line from `redirectTo(HttpUrl, HttpUrl)`:
https://github.com/palantir/conjure-java-runtime/pull/1142/files#diff-97df878facbe38492456bc963582d9e8L122
because in all other invocations, the thing being passed in is guaranteed to be a base URL contained in `baseUrls`

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
+        List<HttpUrl> httpUrls = baseUrls.get();
         // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
-        Optional<Integer> existingUrlIndex = indexFor(existingUrl);
-
-        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
+        Optional<Integer> existingUrlIndex = indexFor(existingUrl, httpUrls);
+        int currentIndex = existingUrlIndex.orElseGet(() -> getCurrentIndex(httpUrls));
```
dansanduleac: could call this something better, this is meant to represent "index of a base URL matching either the existingUrl or the last used url in a previous redirect"

dansanduleac: settled on getLastIndex and lastBaseUrl, what do you think?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
+        }
     }
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
+        List<HttpUrl> httpUrls = baseUrls.get();
         // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
-        Optional<Integer> existingUrlIndex = indexFor(existingUrl);
-
-        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
```
dansanduleac, line 150: this was not a "next index", confusingly, but the index of the base of _this_ existingUrl (falling back to the last used base url)

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
-                                .build());
-                    }
-                });
+        lastBaseUrl.set(redirectBaseUrl);
+
+        if (!isPathPrefixFor(redirectBaseUrl, current)) {
+            // The requested redirectBaseUrl has a path that is not compatible with
+            // the path of the current URL
+            return Optional.empty();
+        } else {
```
ferozco: nit: remove a level of nesting by omitting the `else`

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
+                    .port(redirectBaseUrl.port())
+                    .encodedPath(
+                            redirectBaseUrl.encodedPath()  // matching prefix
+                                    + current.encodedPath().substring(redirectBaseUrl.encodedPath().length()))
+                    .build());
+        }
     }
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
```
ferozco: s/existing/current/

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
             indexForFailedUrl.ifPresent(index ->
                     failedUrls.put(baseUrls.get().get(index), UrlAvailability.FAILED)
             );
         }
     }
 
+    /**
+     * Returns the index of {@link #lastBaseUrl}, which is expected to exist in {@code httpUrls}.
+     */
+    private int getLastIndex(List<HttpUrl> httpUrls) {
```
ferozco: getIndexOfLastBaseUrl?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
```diff
@@ -113,84 +115,102 @@ private static String switchWsToHttp(String url) {
...
 
     @Override
     public Optional<HttpUrl> redirectToNext(HttpUrl existingUrl) {
+        List<HttpUrl> httpUrls = baseUrls.get();
         // if possible, determine the index of the passed in url (so we can be sure to return a url which is different)
-        Optional<Integer> existingUrlIndex = indexFor(existingUrl);
-
-        int potentialNextIndex = existingUrlIndex.orElse(currentUrl.get());
+        Optional<Integer> existingUrlIndex = indexFor(existingUrl, httpUrls);
+        int thisIndex = existingUrlIndex.orElseGet(() -> getLastIndex(httpUrls));
```
ferozco: there has to be a better name than `thisIndex`. previouslyUsedIndex? startIndex?

ferozco:
It doesn't seem like there are any unintentional changes to behaviour caused by this change

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
dansanduleac: I thought of startIndex too, we can use that I guess! it's not necessarily previouslyUsed cos it preferrentially comes from the existingUrl rather than the lastUsedUrl

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UrlSelectorImpl.java
dansanduleac: sounds good

ferozco:
👍

PR #1147, opened 2019-07-16
Concurrency limiters should respect cancellation

j-baker:
Right now the concurrency limiters don't deal with the case where someone cancelled the future before we could set the result. In this case, the permit is leaked. Instead, let's check if the future was otherwise completed and if so, immediately move on to the next one. See internal ticket PDS-94666 for details.


robert3005 (approved):
LGTM

PR #1149, opened 2019-07-16
Conjure should not retry cancelled calls

j-baker:
Currently if you cancel a call, Conjure will cancel the underlying call,
and then retry it.

This has a number of bad outcomes - it generally means that cancelled
calls run twice. It also means that they leak their concurrency permits.

Sadness for all.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -183,6 +183,10 @@ final class RemotingOkHttpCall extends ForwardingCall {
         super.enqueue(new Callback() {
             @Override
             public void onFailure(Call call, IOException exception) {
+                if (isCanceled()) {
+                    callback.onFailure(call, exception);
+                }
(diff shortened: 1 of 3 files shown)


dansanduleac:
:+1:

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -183,6 +183,10 @@ private void enqueueInternal(Callback callback) {
         super.enqueue(new Callback() {
             @Override
             public void onFailure(Call call, IOException exception) {
+                if (isCanceled()) {
+                    callback.onFailure(call, exception);
```
carterkozak, line 187: return?

carterkozak:
thanks 👍

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
dansanduleac: sigh, good spot

PR #1152, opened 2019-07-18
Do not retry streamed requests.

jkozlowski:
* This ports logic from RetryAndFollowupInterceptor in OkHttp.

## Before this PR
Streaming requests are retried and clients need to guard against that.

## After this PR
==COMMIT_MSG==
Streaming requests are not retried.
==COMMIT_MSG==

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -32,7 +32,6 @@ import com.palantir.logsafe.exceptions.SafeIoException;
 import com.palantir.tracing.AsyncTracer;
 import java.io.IOException;
 import java.io.InterruptedIOException;
-import java.net.HttpRetryException;
 import java.net.SocketTimeoutException;
 import java.time.Duration;
 import java.util.Optional;
@@ -212,7 +211,8 @@ final class RemotingOkHttpCall extends ForwardingCall {
                     return;
                 }
 
-                log.info("Retrying call after failure",
+                log.info(
+                        "Retrying call after failure",
                         SafeArg.of("backoffMillis", backoff.get().toMillis()),
                         UnsafeArg.of("requestUrl", call.request().url().toString()),
                         UnsafeArg.of("redirectToUrl", redirectTo.get().toString()),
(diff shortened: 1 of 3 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -245,6 +247,13 @@ public void onResponse(Call call, Response response) throws IOException {
                     return;
                 }
 
+                if (isStreamingBody(call)) {
+                    callback.onFailure(
+                            call,
+                            new HttpRetryException("Cannot retry streamed HTTP body", response.code()));
```
jkozlowski: Question: I think it would be nice to attach some rendering of the body here, so we know what the response actually was. Like the test responds with 503, so host is up: would be nice to know that. Suggestions?

dansanduleac: this will never run given the check at the top: https://github.com/palantir/conjure-java-runtime/pull/1152/files#diff-4c0c5883336d4eff08c9bc33ba1059c0R188

dansanduleac: Can we move this to the top and add the exception parameter as a suppressed exception to this one?
Also can we make this a SafeIoException instead, so it's loggable?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -269,6 +278,10 @@ public void onResponse(Call call, Response response) throws IOException {
                 callback.onFailure(call, new SafeIoException("Failed to handle request, "
                         + "this is an conjure-java-runtime bug."));
             }
+
+            private boolean isStreamingBody(Call call) {
+                return call.request().body() instanceof UnrepeatableRequestBody;
```
dansanduleac: A bit worried about relying on an internal API here, which is gone from okhttp 3.14, though I'm also not sure if they intend to continue development on the 3.x branch.
Do you think we should try to use reflection here to be defensive?

jkozlowski: My guess is we're not sure we'll be upgrading any time soon and once you do this will just fail to compile. If we upgrade there is now a public API for checking this in the newest version.

Can you suggest how would you solve this with reflection?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
jkozlowski: That is in another method :) That check is in onFailure, this is onResponse. Sometimes reading diff in github bites you

dansanduleac: bah, tricked!

jkozlowski: I'll happily make this a SafeIoException tomorrow

dansanduleac:
👍

dansanduleac:
👍

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -47,6 +47,7 @@
 import okhttp3.Request;
 import okhttp3.Response;
 import okhttp3.ResponseBody;
+import okhttp3.internal.http.UnrepeatableRequestBody;
```
iamdanfox, line 50: I am not a fan of this PR because the `okhttp3.internal.http.UnrepeatableRequestBody` class was deleted in okhttp 3.14.0. https://github.com/square/okhttp/find/parent-3.14.0

We knew about this because there was one internal service that relied on this, but now conjure-java-runtime also relies on this internal class :(

jkozlowski: We're not upgrading past 3.13.1 cause the newer versions were buggy, not because internal projects were using UnrepeatableRequestBody. And internal projects were relying on UnrepeatableRequestBody to workaround this bug and also legitimately indicate to OkHttp that those requests should not be retried.

Once the version of okhttp here is upgraded, we can just switch to public API for this, as per the response above.

PR #1166, opened 2019-08-07
Try not to leak response body on cancelled calls

gmaretic:
## Before this PR
There was a race condition when cancelling calls, where we would try to cancel a call, but it would still successfully complete, resulting in leaking the response body.

## After this PR
==COMMIT_MSG==
We submit a task that closes the response body if the call successfully completes despite the cancellation.
==COMMIT_MSG==

## Possible downsides?
We submit more tasks to the executionExecutor which is a cached thread pool


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -148,6 +148,25 @@ public void onResponse(Call call, Response response) {
         }
     }
 
+    private void cancelCallAndCloseResponseBody(SettableFuture<Response> future) {
+        getDelegate().cancel();
+        executionExecutor.submit(() -> closeResponseBody(future));
```
iamdanfox: why is a try-finally not sufficient here? why are we putting this onto an executor?

iamdanfox:
Thanks for the PR! Amusingly the javadoc of `RetryAndFollowUpInterceptor#cancel` says:

> It's the caller's responsibility to close the request body and response body streams; otherwise resources may be leaked.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
carterkozak: in most cases I don't think action is necessary here, could we check if there's a response before putting work on the executor?

gmaretic: Getting the future is still blocking until the call completes, in case we failed to cancel. In the case where we were hitting this, it took several seconds for the call to complete before we were able to get the future and close the response body.
If you have a better way of doing this, I'm happy to change :)

iamdanfox: @gmaretic just revisiting this - couldn't we check the `future.isDone()` / `future.isCancelled()` methods to find out of the call is finished (in a non-blocking way) and avoid submitting stuff to the executor?

I'm thinking that this would be beneficial because the `catch (ExecutionException e)` branch seems a lot more common than the `catch (InterruptedException e)` one.

pkoenig10: Wouldn't it be simplest to just add a listener to the future using `future.transform(...)`? Then we don't need to check whether the future is done or submit anything to an executor.

iamdanfox:
Re-opening for when @gmaretic returns because I think the atlasdb team still wants this

gmaretic:
@iamdanfox The only time we encounter this is when we cancel outstanding requests that then still subsequently respond, so should only be necessary on interrupts. This is also why it needs to be asynchronous, since if our cancellation is ignored, we might have to wait for the response.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,6 +156,20 @@ public void onResponse(Call _call, Response response) {
         }
     }
 
+    private void closeResponseBody(SettableFuture<Response> future) {
+        try {
+            Response response = future.get();
+            if (response.body() != null) {
```
felixdesouza: I think to make intent clearer, you want something along the lines of:

```java
if (future.isDone()) {
  Response response = Futures.getDone(future);
  if (response.body() != null) {
    response.close();
  } 
}
```

I think you may have to deal with the `ExecutionException`, or just use `Futures.getUnchecked`.

felixdesouza: I would actually move the future setting code in here, and call the method `handleInterruption`

Also docs as to why what you're doing works, something along these lines but obviously better.

```java
private InterruptedIOException handleInterruption(SettableFuture<Response> future) {
  getDelegate().cancel();
  // (1) if the success callback gets called before we try to set the exception, we will then close the response
  // since it won't get processed

  // (2) if the success callback gets called after we try to set the exception, the 
  // callback will close the response
  if (!future.setException(new InterruptedIOException("Call cancelled via interruption")) {
    // Future is already set via the callback => Future.isDone is necessarily true
    try {
      Response response = Futures.getDone(future);
      // Future set via the callback but successfully, need to clean up.
      if (response.body() != null) {
        response.close();
      }
    } catch (ExecutionException e) {
      // Future set via the callback but exceptionally.
      // Empty clause which will be dealt with via the fall through block
    }
  }

  Thread.currentThread().interrupt();
  return new InterruptedIOException("Call cancelled via interruption");
}
```

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -131,6 +133,9 @@ public void onResponse(Call _call, Response response) {
             return future.get();
         } catch (InterruptedException e) {
             getDelegate().cancel();
+            if (!future.setException(e)) {
+                closeResponseBody(future);
+            }
             Thread.currentThread().interrupt();
             throw new InterruptedIOException("Call was interrupted during execution");
         } catch (ExecutionException e) {
```
felixdesouza, line 139: I don't believe you ever get a leak here, because this can only be set when the response fails via the callback. We *do* set the exception of the future inside the interruption handling, but we're already inside a catch at the point and we've stopped blocking on the `get`.

changelog/@unreleased/pr-1166.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: fix
+fix:
+  description: Upon cancelling remoting calls, we now submit a task that closes the
```
felixdesouza: technically not accurate since we're not submitting anything

> Upon cancelling a remoting call via interruption, we will ensure that if the call completed successfully, its request body will be closed as it won't be processed upstream.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -119,7 +119,9 @@ public void onFailure(Call _call, IOException exception) {
 
             @Override
             public void onResponse(Call _call, Response response) {
-                future.set(response);
+                if (!future.set(response)) {
+                   closeResponseBody(response);
+                }
```
pkoenig10: We already have this logic for retrofit clients, any way we can we avoid duplication?
https://github.com/palantir/conjure-java-runtime/blob/4.38.1/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactory.java#L95-L98

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,6 +151,31 @@ public void onResponse(Call _call, Response response) {
...
+    private InterruptedIOException handleInterruption(SettableFuture<Response> future) {
+        getDelegate().cancel();
+        // Regardless of the cancel above, the call may have succeeded or is going to succeed, and we need to close the
+        // response body correctly:
+        // (1) if the success callback gets called before we try to set the exception, we must close the response
+        // (2) if the success callback gets called after we set the exception, the callback will close the response
+        if (!future.setException(new InterruptedIOException("Call cancelled via interruption"))) {
+            try {
+                Response response = Futures.getDone(future);
+                closeResponseBody(response);
```
pkoenig10: I think we could make this a lot simpler by just adding a listener to the future that will close the response body. This will ensure it still gets closed even if we are interrupted while waiting.

felixdesouza: I *think* this is equivalent, going off of `ListenableFuture#addListener` javadoc which `Futures#addCallback` delegates to eventually. Even if we have to use the `directExecutor`, but that's also equivalent to this change too I believe.

IIUC this is the reasoning:
1) If we interrupt before the request completes, calling thread is no longer blocking on `Future#get`, we then `Futures#addCallback` in the catch block, and then the thread that will eventually complete the request will then close the body.

2) If we interrupt after the request completes, calling thread is no longer blocking on `Future#get`, and adding the same callback would then cause the response to be closed immediately but in the calling thread

We would also stop trying to close the response body in the normal callback in lines 122 right?

gmaretic: Makes sense, this is much cleaner.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,6 +149,26 @@ public void onResponse(Call _call, Response response) {
         }
     }
 
+    private InterruptedIOException handleInterruption(SettableFuture<Response> future) {
+        getDelegate().cancel();
+        // Regardless of the cancel above, the call may have succeeded or is going to succeed, and we need to make
+        // sure the response body is closed correctly in those cases.
+        future.addListener(() -> closeResponseBody(future), MoreExecutors.directExecutor());
```
carterkozak: Futures.addCallback can make this simpler, onSuccess will provide the Response instance to avoid a Future.get. The callback can be a private enum singleton.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -151,6 +149,26 @@ public void onResponse(Call _call, Response response) {
         }
     }
 
+    private InterruptedIOException handleInterruption(SettableFuture<Response> future) {
```
carterkozak: might consider `@CheckReturnValue` so we don't refactor and forget to throw

gmaretic: Inlined instead

iamdanfox (approved):
Nice!

dansanduleac (approved):

PR #1167, opened 2019-08-07
Increase the default parallel requests for clients

gmaretic:
## Before this PR
The default limits for the dispatcher are quite constricting for using CJR with AtlasDB.

## After this PR
==COMMIT_MSG==
Increased the default limits on number of parallel requests for CJR clients
==COMMIT_MSG==

## Possible downsides?
This might not be what we want in the general case. Alternatively, we can come up with a way to make it configurable

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -94,9 +94,9 @@ public final class OkHttpClients {
 
     static {
         dispatcher = new Dispatcher(executionExecutor);
-        dispatcher.setMaxRequests(256);
+        dispatcher.setMaxRequests(1024);
         // Must be less than maxRequests so a single slow host does not block all requests
-        dispatcher.setMaxRequestsPerHost(64);
+        dispatcher.setMaxRequestsPerHost(256);
 
         dispatcherMetricSet = new DispatcherMetricSet(dispatcher, connectionPool);
     }
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -94,7 +94,7 @@ public final class OkHttpClients {
 
     static {
         dispatcher = new Dispatcher(executionExecutor);
-        dispatcher.setMaxRequests(1024);
+        dispatcher.setMaxRequests(Integer.MAX_VALUE);
         // Must be less than maxRequests so a single slow host does not block all requests
         dispatcher.setMaxRequestsPerHost(256);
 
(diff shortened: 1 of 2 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -94,9 +94,9 @@
 
     static {
         dispatcher = new Dispatcher(executionExecutor);
-        dispatcher.setMaxRequests(256);
+        dispatcher.setMaxRequests(1024);
```
iamdanfox: I think it would be great to have a comment explaining why this is the _right_ number, i.e. what data do we need to look at to be confident this is somehow an optimal tradeoff and what are we trading off here.

(i.e. why shouldn't this be INT_MAX or 10)

iamdanfox: Talking to @j-baker it might actually make sense to rip out this entirely (and set it to INT_MAX), as we already have the concurrency limiter making sure that we don't overwhelm upstreams.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -94,9 +94,9 @@
 
     static {
         dispatcher = new Dispatcher(executionExecutor);
-        dispatcher.setMaxRequests(256);
+        dispatcher.setMaxRequests(Integer.MAX_VALUE);
```
CRogers, line 98: Probably want a comment on why it's Integer.MAX_VALUE (mention the concurrency limiters) before we merge.

jeremyk-91: Alright, makes sense. I'm taking over from @gmaretic as he's on leave - will push this today

jeremyk-91: @CRogers I've added the comment on #1174, as I don't have write permissions to the repository. If we're happy with that we should merge that into this and then this to develop. Thanks!

CRogers:
:+1:

CRogers (approved):

CRogers:
@dansanduleac Please press that approve button as I've been disqualified

iamdanfox:
Just to sanity check this, I looked at one of our biggest deployments on DD and found that no service in the past week has had more than 100 active requests at the same time.

image

Based on this graph I think the PR should have no impact on normally behaving services.

iamdanfox (approved):

JacekLach:
Just to clarify (I agree with the change made completely):

>Based on this graph I think the PR should have no impact on normally behaving services.

That's not quite true - in a situation where there's a reverse proxy between a client and the service, it's quite possible that multiple services are aliased to a single hostname. In such a situation a client could have been hitting the previous per-host max of 64 requests, without any single service actually getting 64 requests concurrently.

PR #1169, opened 2019-08-07
Retry immediately on connection refused

j-baker:
It's a bit silly to enter a backoff loop if we cannot connect to the
host (it's likely down); instead we should just fast retry and
potentially fail more quickly.


iamdanfox:
Is there a possible downside here that if you have a maximum of 4 possible retries and you experience a network blip for 10 milliseconds, that all your retries could happen within this 10ms blip resulting in an overall failed call?

PR #1171, opened 2019-08-08
Allow disabling of handling of SocketExceptions

j-baker:
We have an internal service which manages its own multiplexing between a
number of Conjure clients (this was decided as the right approach a
number of months ago). We'd like it to use a client per remote server.

Unfortunately, with Conjure, discovering the server is down requires a 4
second backoff loop, which makes using these clients very painful.

This PR enables a mode that only we should use which disables retrying
on that, but retains retrying on 503s and 429s. We have a retry loop at
a higher level.

Alternatives considered:

1. Enabling URL selection pluggability in Conjure. This would require significant refactoring in order to be confident with a pluggable API.
2. Adding a hash based approach to Conjure that could be used - this has been attempted, and would also require significant effort within Conjure in order to refactor how URL selection works - would prefer to do this post-Dialogue.
3. Could just continue with our current 'use-many-clients-each-having-all-the-servers' approach, but this has very poor randomization properties, which combines poorly with rate limiting - you still essentially pin your rate limits to the rate limit of the smallest client.
4. We could blacklist internally, but this gives a lot of unpleasant tuning properties - how long do you blacklist clients for, do you do the 'single request tries to use the server' healthcheck thing, do you need a thread pool for that, etc etc.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
@@ -74,6 +74,7 @@ final class RemotingOkHttpCall extends ForwardingCall {
     private final ConcurrencyLimiters.ConcurrencyLimiter limiter;
     private final ClientConfiguration.ServerQoS serverQoS;
     private final ClientConfiguration.RetryOnTimeout retryOnTimeout;
+    private final ClientConfiguration.RetryOnSocketException retryOnSocketException;
 
     private final int maxNumRelocations;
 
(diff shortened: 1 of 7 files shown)


iamdanfox:
Cool having chatted at length about how to make progress here, this seems like the least-worst way of making clients of James' service load-balance all their outgoing requests to multiple upstream nodes, while keeping all requests relating to a single transaction going to one node.  This helps clients avoid using up all their rate limits on one node and then having to restart transactions on another node.

Notably the 10 minute shuffling doesn't solve this problem because it will still try to direct all traffic to a single server node, which may result in the rate limit being completely exhausted on that one node and a bunch of transactions being restarted when they have to switch over to the next node.

iamdanfox (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -276,6 +279,9 @@ public void onResponse(Call call, Response response) throws IOException {
     }
 
     private boolean shouldRetry(IOException exception, Optional<Duration> backoff) {
+        if (retryOnSocketException == ClientConfiguration.RetryOnSocketException.DANGEROUS_DISABLED) {
+            return false;
```
markelliot, line 283: it seems like this would be equivalent to setting retryOnTimeout to DISABLED and setting backoff to empty, why was that configuration insufficient?

j-baker: we still want to backoff on things like 429s and 503s

PR #1183, opened 2019-08-20
Correctly handle  Query Parameters for optional primitives

ferozco:
## Before this PR
Jersey's `SingleValueExtractor` would swallow any `IllegalArgument` exceptions thrown by our ParamConverters and instead set the value of the parameter to be converted to `null`.

This cause invalid requests to resulti in NPEs throughout the applications code.

## After this PR
==COMMIT_MSG==
Correctly handle query parameters for optional primitives
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/Java8OptionalTest.java
+++ b/conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/Java8OptionalTest.java
@@ -103,13 +103,6 @@ public final class Java8OptionalTest {
     }
 
     @Test
-    public void testQueryParam_optionalIntNull() {
-        Response response = target.path("optional/int").queryParam("value", "null").request().get();
-        assertThat(response.getStatus(), is(Status.OK.getStatusCode()));
-        assertThat(response.readEntity(String.class), is("0"));
-    }
-
-    @Test
     public void testQueryParam_optionalIntInvalid() {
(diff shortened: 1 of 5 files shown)


carterkozak:
I don't think we should allow string `null`:

https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#22-query-parameters
> If a value of de-aliased type optional<T> is not present, then the key must be omitted from the query string

https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#24-headers
> Parameters of Conjure type optional<T> must be omitted entirely if the value is not present, otherwise just serialized using the PLAIN Format.

Furthermore `param: optional<string>` in `url?param=null` would be ambiguous if we applied the check, there's no way to send string "null".

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/Java8OptionalTest.java
```diff
@@ -102,6 +102,12 @@ public void testQueryParam_optionalIntEmpty() {
         assertThat(response.readEntity(String.class), is("0"));
     }
 
+    @Test
+    public void testQueryParam_optionalIntInvalid() {
+        Response response = target.path("optional/int").queryParam("value", "foo").request().get();
```
raiju, line 107: Do we want an explicit test for empty string? Would expect that one to be the most likely to be accidentally "fixed"

carterkozak:
Thanks @ferozco 👍

PR #1184, opened 2019-08-21
Fix: Convert retrofit optional Header parameters into string representation of wrapped objects

robert3005:
## Before this PR
Optional header parameters get serialized as `Optional[<value>]` (or `Optional.of(<value>)`) when using retrofit clients

## After this PR
==COMMIT_MSG==
Optional header parameters are serialized as string representation of the wrapped object omitting optional wrapper
==COMMIT_MSG==

## Possible downsides?
Someone was relying on that behaviour? Unlikely but...


iamdanfox (approved):
🤦‍♂

PR #1196, opened 2019-08-28
Don't potentially leak response bodies when cancelling Retrofit calls

j-baker:
This one's a bit of a doozy. When streaming responses with Retrofit, we
will complete the response with the result. If our call is cancelled, we
might leak it.

### DIFF
--- a/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
+++ b/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
@@ -26,6 +26,7 @@
  */
 package com.palantir.conjure.java.client.retrofit2;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.palantir.conjure.java.okhttp.IoRemoteException;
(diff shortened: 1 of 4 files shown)


conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactory.java
```diff
@@ -169,4 +178,11 @@ public void onFailure(Call<R> call, Throwable throwable) {
             return future;
         }
     }
+
+    private static void close(Response<?> response) {
+        if (response.raw().body() == null) {
+            return;
+        }
+        response.raw().body().close();
```
carterkozak, line 186: nit: flow is odd, maybe something like this?

```java
Body body = response.raw().body();
if (body != null) {
    body.close();
}
```

conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/AsyncSerializableErrorCallAdapterFactoryTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Test
+    public void testResponseNotLeakedIfCancelled_listenable() {
+        ListenableFutureBodyCallAdapter<String> adapter = new ListenableFutureBodyCallAdapter<>(String.class);
+        ListenableFuture<String> result = adapter.adapt(call);
+        verify(call).enqueue(callbackCaptor.capture());
+        Callback<String> callback = callbackCaptor.getValue();
+        result.cancel(true);
+        callback.onResponse(call, response);
+        verify(response.raw().body()).close();
+    }
```
carterkozak, line 62: thanks for adding tests

carterkozak (approved):

PR #1205, opened 2019-09-03
Completely revamp tracing spans

dansanduleac:
## Before this PR

Spans for a single request's lifetime are difficult to correlate as they use _different trace ids per attempt_, and it's not obvious when additional attempts are being made.

## After this PR
==COMMIT_MSG==
For a single JaxRsClient request, all spans are now correctly attributed to a single traceId.
==COMMIT_MSG==

both jax-rs and retrofit now trace things correctly, _including_ ListenableFuture/CompletableFuture calls:

image

_This PR uses release candidates tagged on this branch: https://github.com/palantir/tracing-java/pull/124_

## Possible downsides?

- more complexity in c-j-r
- we're touching some hot codepaths here, so any unintended performance regressions could have quite an impact.


dansanduleac:
Looking good:
image

The long times at the beginning of each attempt are the backoff times for each retry.

versions.props
```diff
@@ -11,7 +11,8 @@ com.palantir.conjure.java:conjure-lib = 4.0.0
 com.palantir.conjure.java.api:* = 2.3.0
 com.palantir.conjure.verification:* = 0.16.7
 com.palantir.safe-logging:* = 1.10.1
-com.palantir.tracing:* = 3.1.0
+com.palantir.tracing:* = 3.2.0-rc4
```
iamdanfox: before merging, we need a real release of https://github.com/palantir/tracing-java/pull/124

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/SpanTerminatingInterceptor.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private SpanTerminatingInterceptor() {}
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        DetachedSpan attemptSpan = chain.request().tag(AttemptSpan.class).attemptSpan();
+        DetachedSpan dispatcherSpan = chain.request().tag(SettableDispatcherSpan.class).dispatcherSpan();
+
+        // TODO(dfox): when can the dispatcherSpan ever be null?
+        if (dispatcherSpan == null) {
```
ferozco: If the dispatcher is null then there's probably a bug in cjr. Can we log some sort of warning

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/SettableDispatcherSpan.java
```diff
@@ -0,0 +1,32 @@
...
+
+package com.palantir.conjure.java.okhttp;
+
+import com.palantir.conjure.java.client.config.ImmutablesStyle;
+import com.palantir.tracing.DetachedSpan;
+import org.immutables.value.Value;
+
+@Value.Modifiable
+@ImmutablesStyle
+interface SettableDispatcherSpan {
```
ferozco: These could probably use some docs explaining how they're used

ferozco:
👍

PR #1236, opened 2019-09-12
Fix ThreadLocalRandom usages.

jkozlowski:
## Before this PR

ThreadLocalRandom is not used correctly. Storing the result of ThreadLocalRandom#current in a variable skips seed initialization for the thread.

This would be ok if ExponentialBackoff was only used in a single thread, but I have a feeling (from quick peruse) that callbacks can happen from anywhere? Even if that was the case, it feels like a difficult thing to maintain, so might as well be safe. The impact of this is likely minimal, likely other correct usages would initialize the seed (if they happen before the first ExponentialBackoff for the thread is created), so this is likely not occurring in practice.

## After this PR

Hopefully it is now used correctly.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ExponentialBackoff.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ExponentialBackoff.java
@@ -20,6 +20,8 @@ import com.google.common.annotations.VisibleForTesting;
 import java.time.Duration;
 import java.util.Optional;
 import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
 
 /**
  * Implements "exponential backoff with full jitter", suggesting a backoff duration chosen randomly from the interval
@@ -29,12 +31,16 @@ final class ExponentialBackoff implements BackoffStrategy {
 
     private final int maxNumRetries;
     private final Duration backoffSlotSize;
-    private final Random random;
+    private final Supplier<Random> random;
 
     private int retryNumber = 0;
 
+    ExponentialBackoff(int maxNumRetries, Duration backoffSlotSize) {
+        this(maxNumRetries, backoffSlotSize, ThreadLocalRandom::current);
+    }
+
     @VisibleForTesting
-    ExponentialBackoff(int maxNumRetries, Duration backoffSlotSize, Random random) {
+    ExponentialBackoff(int maxNumRetries, Duration backoffSlotSize, Supplier<Random> random) {
         this.maxNumRetries = maxNumRetries;
         this.backoffSlotSize = backoffSlotSize;
         this.random = random;
@@ -48,6 +54,7 @@ final class ExponentialBackoff implements BackoffStrategy {
         }
 
         int upperBound = (int) Math.pow(2, retryNumber);
-        return Optional.of(Duration.ofNanos(Math.round(backoffSlotSize.toNanos() * random.nextDouble() * upperBound)));
+        return Optional.of(Duration.ofNanos(Math.round(
+                backoffSlotSize.toNanos() * random.get().nextDouble() * upperBound)));
     }
 }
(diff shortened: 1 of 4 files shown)


iamdanfox (approved):

carterkozak:
I've put some thought into catching this sort of thing using errorprone checks, but it's quite hard to tell when usage spans across threads as you noted in the description. We may want to provide a shim around ThreadLocalRandom (below) and require it to be used instead of directly using ThreadLocalRandom.

```java
/**
 * Wrapper around {@link ThreadLocalRandom} which can be used as a drop in replacement for <code>new Random()</code>.
 * Unlike {@link ThreadLocalRandom}, an instance of {@link SafeThreadLocalRandom} may be used by multiple threads
 * because each method invocation delegates to {@link ThreadLocalRandom#current()}.
 *
 * Note that because this is a shared instance, {@link Random#setSeed(long)} is unsupported.
 * This is <b>not</b> a {@link java.security.SecureRandom} instance and must not be used for cryptography.
 */
public final class SafeThreadLocalRandom extends Random {

    private static final Random instance = new SafeThreadLocalRandom();
    private boolean initialized;

    private SafeThreadLocalRandom() {
        super(0L);
        initialized = true;
    }

    /** Gets the singleton {@link SafeThreadLocalRandom} instance. */
    public static Random get() {
        return instance;
    }

    @Override
    public void setSeed(long seed) {
        // setSeed is invoked once in the constructor
        if (initialized) {
            throw new UnsupportedOperationException("SafeThreadLocalRandom does not support setSeed");
        }
    }

    @Override
    protected int next(int bits) {
        throw new UnsupportedOperationException(
                "Internal Random#next(int) is not supported, this should only be thrown if an override is missing");
    }

    @Override
    public void nextBytes(byte[] bytes) {
        ThreadLocalRandom.current().nextBytes(bytes);
    }

    @Override
    public int nextInt() {
        return ThreadLocalRandom.current().nextInt();
    }

    @Override
    public int nextInt(int bound) {
        return ThreadLocalRandom.current().nextInt(bound);
    }

    @Override
    public long nextLong() {
        return ThreadLocalRandom.current().nextLong();
    }

    @Override
    public boolean nextBoolean() {
        return ThreadLocalRandom.current().nextBoolean();
    }

    @Override
    public float nextFloat() {
        return ThreadLocalRandom.current().nextFloat();
    }

    @Override
    public double nextDouble() {
        return ThreadLocalRandom.current().nextDouble();
    }

    @Override
    public double nextGaussian() {
        return ThreadLocalRandom.current().nextGaussian();
    }

    @Override
    public IntStream ints(long streamSize) {
        return ThreadLocalRandom.current().ints(streamSize);
    }

    @Override
    public IntStream ints() {
        return ThreadLocalRandom.current().ints();
    }

    @Override
    public IntStream ints(long streamSize, int randomNumberOrigin, int randomNumberBound) {
        return ThreadLocalRandom.current().ints(streamSize, randomNumberOrigin, randomNumberBound);
    }

    @Override
    public IntStream ints(int randomNumberOrigin, int randomNumberBound) {
        return ThreadLocalRandom.current().ints(randomNumberOrigin, randomNumberBound);
    }

    @Override
    public LongStream longs(long streamSize) {
        return ThreadLocalRandom.current().longs(streamSize);
    }

    @Override
    public LongStream longs() {
        return ThreadLocalRandom.current().longs();
    }

    @Override
    public LongStream longs(long streamSize, long randomNumberOrigin, long randomNumberBound) {
        return ThreadLocalRandom.current().longs(streamSize, randomNumberOrigin, randomNumberBound);
    }

    @Override
    public LongStream longs(long randomNumberOrigin, long randomNumberBound) {
        return ThreadLocalRandom.current().longs(randomNumberOrigin, randomNumberBound);
    }

    @Override
    public DoubleStream doubles(long streamSize) {
        return ThreadLocalRandom.current().doubles(streamSize);
    }

    @Override
    public DoubleStream doubles() {
        return ThreadLocalRandom.current().doubles();
    }

    @Override
    public DoubleStream doubles(long streamSize, double randomNumberOrigin, double randomNumberBound) {
        return ThreadLocalRandom.current().doubles(streamSize, randomNumberOrigin, randomNumberBound);
    }

    @Override
    public DoubleStream doubles(double randomNumberOrigin, double randomNumberBound) {
        return ThreadLocalRandom.current().doubles(randomNumberOrigin, randomNumberBound);
    }

    @Override
    public String toString() {
        return "SafeThreadLocalRandom{}";
    }
}
```

jkozlowski:
@carterkozak I would support that, I also fixed it here: https://github.com/palantir/atlasdb/pull/4237 and another place internally. It's a pretty advanced API

PR #1241, opened 2019-09-16
Allow external usage of trace interceptor

azarum:
## Before this PR

com.palantir.tracing.okhttp3.OkhttpTraceInterceptor was deprecated, and was public to use outside the package

## After this PR
==COMMIT_MSG==
Allow external usage of trace interceptor
==COMMIT_MSG==


dansanduleac:
Talked over slack, this is a new implementation whose behaviour is tied to how we create `RemotingOkHttpCall`s, and shouldn't be public.
The solution for now is to keep using `com.palantir.tracing.okhttp3.OkhttpTraceInterceptor` for @azarum's use case.

PR #1255, opened 2019-09-22
Release limiters when response is returned

### LINKED ISSUE #1192: CJR limits block downloading blobs
## What happened?

Outgoing requests to load blobs from data store block on acquiring permits. Permits don't increase with usage, even though there are no 429s or 503s returned from the data store.

Currently, CJR only releases limiters when the response has been fully read. This slows down the increase in permits for clients that load blobs.

Have CJR release limiters once the client starts streaming the bytes of a response.

johnhferland:
fixes https://github.com/palantir/conjure-java-runtime/issues/1192

## Before this PR
Concurrency permits are only released when the response body is closed.

## After this PR
==COMMIT_MSG==
Concurrency permits are now released when the response is received.
==COMMIT_MSG==

## Possible downsides?
- It's now technically possible to have more concurrent connections than the maximum number of permits
- This increases the rate at which limits will increase, which may increase load on servers from clients. I expected this will be particularly noticeable for reading larger objects.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimitingInterceptor.java
```diff
@@ -52,7 +46,7 @@
  * 429 and 503 response codes are used for backpressure, whilst 200 -> 399 request codes are used for determining
  * new limits and all other codes are not factored in to timings.
  * <p>
- * Concurrency permits are only released when the response body is closed.
+ * Concurrency permits are released when the response is received.
```
jkozlowski, line 49: I think this was discussed in the past and in general would be kinda against the principle of concurrency limiting: the fact that the body is being streamed usually means that the worker thread might have been released, but streaming the body still usually consumes resources (especially if it's just a blocking read that goes from some external resource and writes to the response body).

I think the reason why you are not observing any 429 in general in that situation is that the internal product that I expect you're referring too has request/sec rate limiting implemented, not concurrency limiting. I suspect releasing more of requests here would eventually lead to 503s from the product.

iamdanfox:
On your original issue (this is artifacts right?) doesn't c-j-r's limit still increase, so this should only affect a cold startup?

PR #1256, opened 2019-09-23
okhttp clients sets SO_KEEPALIVE

iamdanfox:
## Before this PR

As per the internal thread "SO_KEEPALIVE on Java sockets", c-j-r's default read timeout of 5 minutes means that after a network change, a client can sit there waiting for bytes that will never arrive for 5 minutes before retrying.

## After this PR
==COMMIT_MSG==
All outgoing connections created by our OkhttpClients will now have SO_KEEPALIVE enabled. This should make clients more responsive to a router getting terminated.
==COMMIT_MSG==

See `4.2.3.6  TCP Keep-Alives` in https://www.ietf.org/rfc/rfc1122.txt

- [x] before merging, add tests.

## Possible downsides?

- this stuff all delegates to the OS, so if the OS does crazy things we're boned.
- if some of our upstream servers don't respond correctly to these packets, _successful connections will stop working_

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/KeepAliveSslSocketFactory.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/KeepAliveSslSocketFactory.java
@@ -22,7 +22,11 @@ import javax.net.ssl.SSLSocketFactory;
 
 /**
  * Enables {@link java.net.StandardSocketOptions#SO_KEEPALIVE}. By default this will only send a packet after two
- * hours (which is long after our timeouts), but this can be lowered to a useful timeout at the OS level.
+ * hours (which is long after our timeouts), but this can be lowered to be useful in the kernel, e.g.:
+ *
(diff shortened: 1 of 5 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/KeepAliveSslSocketFactory.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    Socket wrap(Socket socket) throws SocketException {
+        socket.setKeepAlive(true);
+        return socket;
+    }
+
+    public String toString() {
+        return "ConfigurableSslSocketFactory{delegate=" + delegate + '}';
```
ashrayjain: should this be `KeepAliveSslSocketFactory`?

ferozco:
👍  LGTM besides fixing up the toString

dansanduleac (approved):

PR #1257, opened 2019-09-23
Only apply stalebot to PRs (not issues)

iamdanfox:
Aaargh unintentionally re-enabled stalebot with old config. I'll unlabel all the issues.


ferozco:
👍

PR #1262, opened 2019-09-30
Separate QoS responses in host metrics

pkoenig10:
## Before this PR
429 and 503 responses are included in the host metrics `get4xx` and `get5xx` timers, respectively. This can result in false positives in the `SERVICE_DEPENDENCY` health check.

## After this PR
429 and 503 are not included in the host metrics `get4xx` and `get5xx` timers. These responses are counted in a separate `getQos` timer.


ferozco:
👍

PR #1266, opened 2019-10-03
Add KeepAliveSslSocketFactory wrapper to the other ClientConfigurations.of()

ucabvas:
## Before this PR
In https://github.com/palantir/conjure-java-runtime/pull/1256 we added the `KeepAliveSslSocketFactory` wrapper to one of the `ClientConfigurations.of()` functions. Some downstream code still uses the other one and doesn't benefit from this functionality. We're now adding the wrapper to the second `ClientConfigurations.of()` function as well.

## After this PR
==COMMIT_MSG==
SO_KEEPALIVE enabled for all OkhttpClients created using `ClientConfigurations.of()`.
==COMMIT_MSG==

## Possible downsides?
Same as https://github.com/palantir/conjure-java-runtime/pull/1256:
* this stuff all delegates to the OS, so if the OS does crazy things we're boned.
* if some of our upstream servers don't respond correctly to these packets, successful connections will stop working

### DIFF
--- a/.changelog.yml
+++ b/.changelog.yml
@@ -1,0 +1,3 @@
+# Excavator auto-updates this file. Please contribute improvements to the central template.
+
+# This file is intentionally empty. The file's existence enables changelog-app and is empty to use the default configuration.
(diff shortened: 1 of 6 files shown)


iamdanfox:
Have opened a PR to make check pass: https://github.com/palantir/conjure-java-runtime/pull/1267

iamdanfox (approved):
Thanks @ucabvas, merging in develop should make this green

PR #1269, opened 2019-10-04
ClientConfiguration equality works again

iamdanfox:
## Before this PR

In https://github.com/palantir/conjure-java-runtime/pull/1256, I introduced a KeepAliveSocketFactory class which was intended to be entirely an implementation detail.

However, some internal consumers have been unable to pick up this change due to their mockito tests failing, because setting up a mock with `when(ClientConfigurations.of(a,b,c)).then` doesn't match later invocations because I broken equality of the ClientConfiguration class (by including the KeepAliveSocketFactory).

## After this PR
==COMMIT_MSG==
Multiple calls to ClientConfigurations.of() with the same arguments will once again produce an equal ClientConfiguration object. This fixes a regression introduce in 4.37.0
==COMMIT_MSG==

## Possible downsides?
-??

### DIFF
--- a/changelog/@unreleased/pr-1269.v2.yml
+++ b/changelog/@unreleased/pr-1269.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: Multiple calls to ClientConfigurations.of() with the same arguments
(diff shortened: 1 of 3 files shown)


ferozco:
👍

PR #1295, opened 2019-10-29
Propagate 401s

### LINKED ISSUE #1261: Propagate 401 Unauthenticated Exceptions
It'd be nice for clients to differentiate between 403 and 401. We'll want to first do https://github.com/palantir/conjure/pull/367 and https://github.com/palantir/conjure/issues/398 and then we can add propagation like there [already is for 403](https://github.com/palantir/conjure-java-runtime/blob/develop/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java#L80)

pkoenig10:
Fixes #1261

## Before this PR
401 responses are not propagated.

## After this PR
401 responses are propagated.


carterkozak:
This will require a similar change here: https://github.com/palantir/conjure-java/blob/724f4e77502020a21cc379cfe7d987bd4f93c020/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java#L120-L140

carterkozak (approved):

PR #1298, opened 2019-10-30
Implement unknown remote exception

gatesn:
## Before this PR
Non-Conjure servers (or proxies in front of Conjure servers) will usually return a response error body that cannot be parsed as a Conjure exception. In such cases, CJR currently throws a SafeIoException where the only way to retrieve the status code is to parse the exception message itself.

## After this PR
A structured exception is raised allowing the client to interrogate, log, monitor, etc. the status codes from the server.

## Possible downsides?
* Will break logic written to parse exception messages.
* ~Until https://github.com/palantir/conjure-java-runtime-api/pull/374 merges, there is a regression in the safe-logged arguments~

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UnknownRemoteExceptionResponseHandler.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UnknownRemoteExceptionResponseHandler.java
@@ -17,40 +17,37 @@
 package com.palantir.conjure.java.okhttp;
 
 import com.google.common.io.CharStreams;
-import com.palantir.logsafe.SafeArg;
-import com.palantir.logsafe.UnsafeArg;
-import com.palantir.logsafe.exceptions.SafeIoException;
+import com.palantir.conjure.java.api.errors.UnknownRemoteException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.util.Optional;
-import javax.ws.rs.core.HttpHeaders;
 import okhttp3.Response;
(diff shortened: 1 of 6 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/UnknownRemoteExceptionResponseHandler.java
```diff
@@ -17,40 +17,37 @@
...
     @Override
-    public Optional<IOException> handle(Response response) {
+    public Optional<UnknownRemoteException> handle(Response response) {
         if (response.isSuccessful() || response.code() == MoreHttpCodes.SWITCHING_PROTOCOLS) {
             return Optional.empty();
         }
 
+        String body;
         try {
-            String body = response.body() != null && response.body().byteStream() != null
```
dansanduleac, line 42: note: turns out `response.body().byteStream()` is never null

dansanduleac:
👍

iamdanfox:
As you predicted in the possible downsides section:

> Will break logic written to parse exception messages.

It seems this PR has caused a ton of upgrade WC PRs to be blocked because of code like the following:

```java
        try {
            foo();
        } catch (SafeIoException io) {
            Optional<Arg<?>> codeArg = io.getArgs().stream()
                    .filter(arg -> "code".equals(arg.getName()))
                    .findAny();

            if (!codeArg.isPresent()) {
                fail("Couldn't find status code");
            }

            assertThat(codeArg.get().getValue()).isEqualTo(307);
        }
```

I've switched this specific one with:
```java
        assertThatThrownBy(() -> {
            foo();
        }).isInstanceOfSatisfying(UnknownRemoteException.class, e -> assertThat(e.getStatus()).isEqualTo(307));
```

PR #1300, opened 2019-10-30
Accept an OkHttpClient.Builder

gatesn:
## Before this PR
OkHttpClients pre-configures OkHttp with metrics, tracing, client and server QoS, URI rotation, cipher suites etc. None of this is specific to Conjure RPC.

I've therefore found myself wanting to use `OkHttpClients` for non-Conjure RPC requests, such as downloading files. Currently when doing this I also add a DiskCache to the okhttp client using `OkHttpClient.Builder#cache`.

As it doesn't make sense for Conjure `ClientConfiguration` to support a disk cache (although this would also be a possible solution?), I wonder if an acceptable solution would be to allow clients to bring their own `OkHttpClient.Builder` that is modified by `OkHttpClients` to provide all the goodies mentioned above.

## After this PR
`OkHttpClients#create` can take a pre-configured `OkHttpClient.Builder` that is modified and wrapped to add on the existing feature set.

## Possible downsides?
Given this jar is named `okhttp` - I don't feel like there's any API leak here. Conjure Java Runtime can still switch away from OkHttp in the future without having to support this set of classes.


gatesn:
Closed for non-forked PR: https://github.com/palantir/conjure-java-runtime/pull/1300

PR #1305, opened 2019-11-01
Add failure meter with IOException tag

parker89:
## Before this PR

Currently, if an outage happens or a service is throwing a lot of 500 responses, it can be due to timeouts when hitting a service. This metric will make it easier to identify which service is causing the timeouts. 

If service A sends requests to service B, this will let:

- service A have a dashboard showing that requests to B are timing out
- service B have a dashboard listing services which are timing out on it

## After this PR
==COMMIT_MSG==
Adds a meter, `client.response.error` with tag `reason:IOException` when a request fails due to an IOException
==COMMIT_MSG==

## Possible downsides?

more metrics = more money

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/InstrumentedInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/InstrumentedInterceptor.java
@@ -16,6 +16,7 @@
 
 package com.palantir.conjure.java.okhttp;
 
+import com.codahale.metrics.Meter;
 import com.codahale.metrics.Timer;
 import com.google.common.base.Stopwatch;
 import com.palantir.tritium.metrics.registry.MetricName;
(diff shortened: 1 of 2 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/InstrumentedInterceptor.java
```diff
@@ -37,11 +38,17 @@
...
     private final String serviceName;
     private final Timer responseTimer;
+    private final Meter ioExceptionMeter;
 
     InstrumentedInterceptor(TaggedMetricRegistry registry, HostEventsSink hostEventsSink, String serviceName) {
         this.hostEventsSink = hostEventsSink;
         this.serviceName = serviceName;
         this.responseTimer = registry.timer(name());
+        this.ioExceptionMeter = registry.meter(MetricName.builder()
+                .safeName("client.response.failure")
```
iamdanfox: We're estimating this will cost $5875/month right?

iamdanfox: Given that we already have `server.response.error` would we do a little `s/failure/error/` to be consistent

carterkozak: Discussed with @dansanduleac the other day, but we may want to consider coupling this with a conjure-java-runtime-api change for client exceptions which didn't produce a response status. This way we can differentiate socket timeout, connection timeout, no route to host, etc. Ideally we would spec out the client runtime exceptions we throw from clients so that retrofit, dialogue, and feign produce the same failures, and migration between them has fewer subtle edge cases.

parker89: My estimate is $8k/mo

parker89: Agree it would be good to differentiate between types of IOExceptions down the line. I don't see that as a blocker for merging this PR though.

carterkozak: I don’t think this pr makes it more difficult to implement that in the future :-)

parker89:
bump @iamdanfox

iamdanfox (approved):
Approving because we found a bunch of other unused metrics (in foundry-spark) that we deleted which should offset this cost increase

PR #1321, opened 2019-11-14
Improvement: Support decoding response values into Optional<ResponseBody> in retrofit clients

robert3005:
## Before this PR
Endpoints returning `Optional<ResponseBody>` in retrofit interfaces would fail with jackson response error even though their encoding is application/octet-stream 

## After this PR
==COMMIT_MSG==
Decode `Optional<ResponseBody>` return value types as optionals of binary stream instead of mistakenly treating them as json responses
==COMMIT_MSG==

## Possible downsides?
How do we want to handle empty 200 responses? This is a little bit different to jackson but haven't checked what spec says about this.

### DIFF
--- a/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/OptionalResponseBodyConverterFactory.java
+++ b/conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/OptionalResponseBodyConverterFactory.java
@@ -1,0 +1,58 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.client.retrofit2;
+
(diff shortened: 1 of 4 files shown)


robert3005:
Runtime change for https://github.com/palantir/conjure-java/pull/643

conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/OptionalResponseBodyConverterFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    @Nullable
+    @Override
+    public Converter<ResponseBody, ?> responseBodyConverter(
+            Type type,
+            Annotation[] _annotations,
+            Retrofit _retrofit) {
+        if (type instanceof ParameterizedType) {
+            Type innerType = getParameterUpperBound(0, (ParameterizedType) type);
+            Class<?> rwaInnerClass = getRawType(innerType);
+            if (ResponseBody.class.isAssignableFrom(rwaInnerClass)) {
```
dansanduleac: tpyo

conjure-java-retrofit2-client/src/main/java/com/palantir/conjure/java/client/retrofit2/OptionalResponseBodyConverterFactory.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return null;
+    }
+
+    private enum OptionalResponseBodyConverter implements Converter<ResponseBody, Optional<ResponseBody>> {
+        INSTANCE;
+
+        @Override
+        public Optional<ResponseBody> convert(ResponseBody value) {
+            return Optional.ofNullable(value);
```
dansanduleac: should change to `of` since when we get 204, this will never be called, and instead we expect `CoerceNullValuesCallAdapterFactory` to return an `Optional.empty()`

dansanduleac:
👍

robert3005:
👍

dansanduleac:
🚅

Issue #1330, opened 2019-11-20
Undertow endpoints cannot deserialize optional header params sent by Feign clients


pkoenig10:
## What happened?
A service is exposing a Conjure Undertow endpoint with an `Optional<AuthHeader>` header parameter. A Conjure Feign client calls this endpoint with an `Optional.empty()` value and receives a 400 because this parameter fails to deserialize.

## What did you want to happen?
The request from the client should have succeeded.

## Details

This is related to https://github.com/palantir/conjure-java-runtime/issues/790.

This request works against a Jersey service because Jersey will [return null if an exception is thrown while deserializing an empty string](https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L81-L90).

pkoenig10:
I can provide a link to an internal CI build that demonstrates this behavior.

carterkozak:
Same problem as https://github.com/palantir/conjure-java-runtime/issues/1331#issuecomment-556465421

> The conjure spec requires optional headers to be absent from the request to deserialize into an empty optional, otherwise we would not be able to differentiate between a present optional of an empty string, and and absent optional.

pkoenig10:
Right. Ideally we would fix #790. Although it doesn't seems like we are planning to do that.

Also, without a server-side fix, servers with optional parameters cannot switch to using Undertow endpoints because they cannot guarantee that they will not break existing clients.

The only reason optional parameters work at all today is because _both_ clients and servers do not adhere to the spec.

a-k-g:
Just ran into this in another internal product.

Issue #790, opened 2018-08-09
Optional headers with an empty value are converted to the empty string


DoronShapiro:
## What happened?

I have an endpoint method with an optional argument annotated by `@HeaderParam`:
```java
@GET
@Path("/doStuff")
String doStuff(@HeaderParam("value") Optional<String> value);
```
I passed it the value `Optional.empty()` via a `JaxRsClient`:
```java
MyService service = JaxRsClient.create(...
service.doStuff(Optional.empty())
```

However, the implementation of `doStuff` receives the value `Optional.of("")`, rather than `Optional.empty()`.

There is a failing test case to show this at https://github.com/palantir/http-remoting/compare/develop...DoronShapiro:optional-header-tests.

This behavior happens because the implementation of [`Java8OptionalAwareContract`](https://github.com/palantir/http-remoting/blob/develop/jaxrs-clients/src/main/java/com/palantir/remoting3/jaxrs/feignimpl/Java8OptionalAwareContract.java#L81) serializes empty headers as an empty string, but when they are deserialized by the [`Java8OptionalParamConverterProvider`](https://github.com/palantir/http-remoting/blob/develop/jersey-servers/src/main/java/com/palantir/remoting3/servers/jersey/Java8OptionalParamConverterProvider.java#L52), it is assumed that the value will be null if empty (and that the empty string represents a non-empty value).

## What did you want to happen?

The server should receive a value of `Optional.empty()` for the header argument.

One solution is for the client to not include the header at all if its value is `Optional.empty()`.

iamdanfox:
I think this is still a valid bug - keep open!

uschi2000:
keep open

pkoenig10:
This is pretty easy to fix, but requires upgrading Feign to get https://github.com/OpenFeign/feign/pull/724.

pkoenig10:
Looking closer that PR doesn't actually fix the issue. The header value in the `RequestTemplate` is not `null` but is the template string:
https://github.com/OpenFeign/feign/blob/ed2cef04658a28b5ca8c50645032df0d5c7fd7b7/core/src/main/java/feign/RequestTemplate.java#L150-L154

The tests in that PR pass because of a bug in `RecordedRequestAssert#hasNoHeaderNamed`:
https://github.com/OpenFeign/feign/pull/835

The real fix is in https://github.com/OpenFeign/feign/pull/778, which completely overhauled the `RequestTemplate`.

pkoenig10:
[...]

PR #1332, opened 2019-11-20
Improve handling of empty optional params

### LINKED ISSUE #1331: Servers cannot deserialize empty primitive optional values sent by Feign clients
## What happened?
A service is exposing an endpoint with an `OptionalLong` header parameter. A Conjure Feign client calls this endpoint with an `OptionalLong.empty()` value and receives a 400 because this parameter fails to deserialize with `NumberFormatException`.

The request from the client should have succeeded.

## Details

This is related to https://github.com/palantir/conjure-java-runtime/issues/790.

Requests with `Optional<T>` parameters work because Jersey will [return null if an exception is thrown while deserializing an empty string](https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L81-L90).

pkoenig10:
Fixes #1331.

## Before this PR
When deserializing `OptionalInt`, `OptionalLong`, or `OptionalDouble` values used as header or query parameters, empty string values fail to deserialize with `NumberFormatException`.

## After this PR
When deserializing `OptionalInt`, `OptionalLong`, or `OptionalDouble` values used as header or query parameters, empty string values deserialize as empty optional values.

This is necessary to properly support optional header/query params for Conjure clients due to https://github.com/palantir/conjure-java-runtime/issues/790.

This also make the handling of primitive optional consistent with Jersey's handling of `Optional<T>`:
https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverters.java#L81-L90

## Possible downsides?
- We only need this behavior for header params, but it will also apply to query params.
- This behavior violates the Conjure spec and is inconsistent with the behavior of Undertow endpoints. But Undertow endpoints have this problem for both primitive optionals and `Optional<T>`. See #1330.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/Java8OptionalDoubleParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/Java8OptionalDoubleParamConverterProvider.java
@@ -43,7 +43,7 @@ public final class Java8OptionalDoubleParamConverterProvider implements ParamCon
     public static final class OptionalDoubleParamConverter implements ParamConverter<OptionalDouble> {
         @Override
         public OptionalDouble fromString(final String value) {
-            if (value == null) {
+            if (value == null || value.isEmpty()) {
                 return OptionalDouble.empty();
             }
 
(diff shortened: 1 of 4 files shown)


pkoenig10:
#1337 is the better fix.

PR #1334, opened 2019-11-21
Fix: Cleanup jackson version overrides

robert3005:
## Before this PR
We had a bunch of specific jackson versions

## After this PR
==COMMIT_MSG==
Consistently specify jackson version as 2.10.1
==COMMIT_MSG==

## Possible downsides?
we already picked up 2.10.1 and this is just a cleanup that could prevent issues in future


versions.lock
```diff
@@ -1,12 +1,12 @@
...
+com.fasterxml.jackson.core:jackson-annotations:2.10.1 (9 constraints: 5da6fb11)
+com.fasterxml.jackson.core:jackson-core:2.10.1 (14 constraints: e4293053)
+com.fasterxml.jackson.core:jackson-databind:2.10.1 (19 constraints: a061ef90)
+com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:2.10.1 (2 constraints: 941c47c9)
 com.fasterxml.jackson.datatype:jackson-datatype-guava:2.10.1 (3 constraints: 6a20e231)
-com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.9.9 (3 constraints: 4a200b23)
-com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.9.9 (2 constraints: a0134956)
+com.fasterxml.jackson.datatype:jackson-datatype-jdk8:2.10.1 (3 constraints: 6a20e231)
+com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.10.1 (2 constraints: c0138060)
 com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:2.10.1 (2 constraints: d12e2568)
```
carterkozak, line 10: heh

carterkozak (approved):

dansanduleac (approved):

versions.props
```diff
@@ -1,17 +1,6 @@
 ch.qos.logback:logback-classic = 1.1.7
 # catch-all jackson version
 com.fasterxml.jackson.*:jackson-* = 2.10.1
-# individual components are sometimes updated out of band to fix vulnerabilities
```
dansanduleac, line 4: For future reference, a star constraint like the above will handle these hotfix releases just fine as well.
As in, it will try to find the given version for all components, but will fall back to the highest constrained version (of a bottom lib like `jackson-databind`) if the requested version doesn't exist.

In fact jackson itself features in gradle's integration tests for this feature: 
https://github.com/gradle/gradle/blob/5d5c27d35e276d674da9a28a4af972d79589b78d/subprojects/dependency-management/src/integTest/groovy/org/gradle/integtests/resolve/alignment/AlignmentIntegrationTest.groovy#L241-L292

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -104,6 +104,7 @@ public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
                 .registerModule(new JavaTimeModule())
                 .registerModule(new LenientLongModule())
                 .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                .disable(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS)
```
robert3005, line 107: For those curious https://github.com/FasterXML/jackson-modules-java8/pull/75

PR #1336, opened 2019-11-21
Add metric for internal server error not caused by remote exceptions

parker89:
==COMMIT_MSG==
Add `conjure.server.exception` metric for internal server errors not caused by remote exceptions
==COMMIT_MSG==

A conjure server will return a HTTP 500 for three reasons:
- The server threw an exception while processing the request (eg RuntimeException)
- The server received a 500 from another server and the RemoteExceptionMapper caused it to propagate the 500
- The server received a 400 from another server and the RemoteExceptionMapper caused it to return a 500 since it wasn't properly handled by the service

Currently all of those events are captured in the metric for 500 responses. This PR will add a metric to only surface the first scenario (ie 500s NOT caused by RemoteExceptions). This will help owners of a service determine if 500s are due to an issue with their service or issues with a dependent service. 

For ex:
```
caller/browser -> [server A] -> [server B]
```

I would like this metric to increment when Server A throws a 500 because it has a bug, but NOT when Server A throws a 500 because it received a 500 from Server B. 

As far as the implementation, I'm open to suggestions here. I thought adding this to the ExceptionMapper classes would be most straightforward. However that required changing the `ConjureJerseyFeature` to be class instead of an enum, and adding a constructor that takes in a `TaggedMetricRegistry`. Unfortunately this breaks the public API (https://github.com/palantir/conjure-java-runtime#conjure-java-jersey-server), so if there's a less intrusive way of adding this metric, I'm happy to make changes.


iamdanfox:
@parker89 the use-case sounds very legit - intuitively I'd expect the number of truly _internal_ server errors (i.e. all transitive calls succeeded but my server still barfed) to be very low, and I'd expect it to be a very good indicator that the dev team should probably be notified.

In terms of the implementation, given we we've got your new `client.response.error` instrumentation on all outgoing network calls, and we've got the overall WC metric of `server.response.error`, couldn't we subtract one from the other to get the interesting bit in the middle?

ferozco:
Could we also use [metric-schema](https://github.com/palantir/metric-schema/) to define the new metrics? You can see an example [here](https://github.com/palantir/conjure-java-runtime/pull/1341)

parker89:
@iamdanfox agree the number here should be quite low, which should allow high-confidence paging and blue/green rollback.  Most 500s are probably caused by not handling 400s properly, followed by 500s caused by propagating 500s, and then followed by the actual initiating 500 (the only thing this metric will surface).

`client.response.error` is pre-retry (ie each failure will increment this by 4, and many non-failures will also increment this metric), so it's hard to take the difference of these. It's also hard to subtract metrics sometimes, particularly around node restarts, which is why I had to add the `server.response.500` metric instead of just subtracting `server.response.dropped` from `server.response.5xx`. 

@ferozco I'll update this to use the metric-schema.


Do either of you have an opinion on the breaking change to how users will have to instantiate a `ConjureJerseyFeature`? And how do you feel about the metric name?

parker89:
Updated the PR to use metric-schema

parker89:
@iamdanfox I updated the PR to not require a breaking API/ABI change.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -16,30 +16,41 @@
...
 import javax.ws.rs.core.Feature;
 import javax.ws.rs.core.FeatureContext;
 
 public enum ConjureJerseyFeature implements Feature {
     INSTANCE;
 
+    private final Meter internalExceptionMeter;
+
+    ConjureJerseyFeature() {
+        TaggedMetricRegistry registry = DefaultTaggedMetricRegistry.getDefault();
```
iamdanfox: I think this guy is actually deprecated... would you mind flipping to the SharedTaggedMetricRgistry thing? https://github.com/palantir/tritium/blob/18e2697fd596e717129f455abbfccfe34780ad1d/tritium-registry/src/main/java/com/palantir/tritium/metrics/registry/DefaultTaggedMetricRegistry.java#L34

carterkozak: Are we considering that 'less' deprecated? https://github.com/palantir/tritium/pull/519

conjure-java-jersey-server/src/main/metrics/conjure-server-metrics.yml
```diff
@@ -0,0 +1,12 @@
+options:
+  javaPackage: com.palantir.conjure.java.server.jersey
+  javaVisibility: packagePrivate
+namespaces:
+  conjure.server:
+    shortName: ConjureServer
+    docs: Conjure server metrics
+    metrics:
+      exception:
```
iamdanfox: So I think it would be good to align this will all the other existing server-side metrics we have in WC, e.g. `server.response.500`, `server.response.error`, `server.response.all`.

- I'd like to include the `.all` suffix to emphasise that this isn't tagged by the endpoint, it's the entire server.
- I'd like to include a `server.response` prefix to emphasise that these are errors encountered while trying to produce a response (as distinct from background tasks or config deserialization etc).

thoughts on calling this something like `server.response.internalerror.all`?

iamdanfox:
Btw I think we gotta make sure your metrics also reflect exceptions that come out of [conjure-java](https://github.com/palantir/conjure-java) generated undertow handlers, as these don't go through the normal Jersey codepaths here...

conjure-java-jersey-server/src/main/metrics/conjure-server-metrics.yml
parker89: Good call, I think that name makes sense.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
parker89: I used `DefaultTaggedMetricRegistry.getDefault()` since it's used that way here: https://github.com/palantir/conjure-java-runtime/blob/develop/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java#L91

Is just instantiating a `new DefaultTaggedMetricRegistry()` the right thing to do? It's done that way here: https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DispatcherMetricSet.java#L34

parker89:
@iamdanfox I updated the PR to address your two comments. If this looks good to you, I'll also figure out how to add this metric to the undertow handler.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
iamdanfox: I think the thing you've done here is OK, because ultimately the shared one _will_ get spat out into a `var/log/metric.log` file by sls-logging.  If you'd just created a brand `new DefaultTaggedMetricRegistry()` without hooking it up then the metrics wouldn't get logged.

iamdanfox: If we were writing this stuff from scratch, then wiring in an instance as a constructor param would be preferrable, but I think avoiding the java break is more important in this repo.

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/JsonExceptionMapperTest.java
```diff
@@ -18,21 +18,30 @@
...
-    private final JsonExceptionMapper<RuntimeException> mapper = new JsonExceptionMapper<RuntimeException>() {
-        @Override
-        ErrorType getErrorType(RuntimeException _exception) {
-            return ErrorType.INVALID_ARGUMENT;
-        }
-    };
+    private final TaggedMetricRegistry taggedMetrics = new DefaultTaggedMetricRegistry();
+    private final Meter internalExceptionMeter = taggedMetrics.meter(MetricName.builder()
+            .safeName("conjure.server.exception")
+            .build());
```
iamdanfox: maybe just make an anonymous meter here instead e.g. `new Meter()` (as this name is now out of sync), then further down you can actually assert that it was incremented when you spect it to!

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -16,30 +16,41 @@
...
 import javax.ws.rs.core.Feature;
 import javax.ws.rs.core.FeatureContext;
 
 public enum ConjureJerseyFeature implements Feature {
     INSTANCE;
 
+    private final Meter internalErrorMeter;
+
+    ConjureJerseyFeature() {
+        TaggedMetricRegistry registry = new DefaultTaggedMetricRegistry();
```
iamdanfox: oh hang on this isn't gonna have the outcome you want - metrics only end up on disk in order to appear in DataDog.  WC makes up a new metric registry and then hooks up sls-logging to make it dump them every 30 seconds. It _also_ hooks up a reporter to the `SharedTaggedMetricRegistry.getSingleton()`.

In this case, nothing is hooked up to this instance, so they'll never appear on disk.

I think we _do_ need to use either `DefaultTaggedMetricRegistry.getDefault()` or `SharedTaggedMetricRegistry.getSingleton()` (would prefer the latter as the former was deprecated first).

conjure-java-jersey-server/src/main/metrics/conjure-server-metrics.yml
```diff
@@ -0,0 +1,12 @@
+options:
+  javaPackage: com.palantir.conjure.java.server.jersey
+  javaVisibility: packagePrivate
+namespaces:
+  server.response:
+    shortName: ConjureServer
```
iamdanfox: suggest JerseyServer as the shorthame?

conjure-java-jersey-server/src/main/metrics/conjure-server-metrics.yml
```diff
@@ -0,0 +1,12 @@
...
+    docs: Conjure server metrics
```
iamdanfox: Metrics produced by Jersey servers using the `ConjureJavaJerseyFeature.INSTANCE`?

parker89: Changed to your suggested wording.

conjure-java-jersey-server/src/main/metrics/conjure-server-metrics.yml
parker89: Done

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
parker89: Makes sense. Changed it to use the latter.

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/JsonExceptionMapperTest.java
parker89: Done, and added another test below

parker89:
@iamdanfox Thanks for the comments. I updated everything as you suggested.

parker89:
Conjure-java equivalent here: https://github.com/palantir/conjure-java/pull/674/

parker89:
I also had to bump up tritium to 0.12.5 to get the new SharedTaggedMetricRegistries class.

iamdanfox (approved):
Ok let's give it a shot.  This is only one extra meter per server, so shouldn't be much extra DD$

iamdanfox:
@callumr I've lost my approval perms - would you mind?

CRogers:
:+1:

PR #1340, opened 2019-11-22
WIP: try upgrading okhttp

iamdanfox:
## Before this PR

We've been pinned on okhttp 3.13.1 for a long time, but @jkozlowski is investigating some bugs related to http2.

## After this PR
This PR is gonna be a baby-step release-by-release upgrade process to get onto latest.
==COMMIT_MSG==

==COMMIT_MSG==


versions.props
```diff
@@ -28,12 +28,10 @@ org.jmock:jmock = 2.12.0
 org.mockito:mockito-core = 3.1.0
 org.slf4j:slf4j-api = 1.7.12
 io.leangen.geantyref:geantyref = 1.3.4
+com.squareup.okhttp3:* = 3.14.0
```
jkozlowski, line 31: I think this particular version was known to cause issues? Curious how do we make sure we can go through the step-by-step upgrade, without affecting users? At the very least should probably try 3.14.4

PR #1361, opened 2019-12-06
Test retrofit ListenableFuture tracing

gatesn:
TraceIDs don't seem to propagate from requests to ListenableFuture callbacks

### DIFF
--- a/conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
+++ b/conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
@@ -35,6 +35,7 @@ import com.palantir.conjure.java.okhttp.HostMetricsRegistry;
 import com.palantir.conjure.java.serialization.ObjectMappers;
 import com.palantir.logsafe.exceptions.SafeNullPointerException;
 import com.palantir.logsafe.testing.Assertions;
+import com.palantir.tracing.Observability;
 import com.palantir.tracing.Tracer;
 import com.palantir.tracing.Tracers;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


gatesn:
We currently pull in Retrofit 2.4.0, which suggests we're using the OkHttp dispatcher executor by default: https://github.com/square/retrofit/blob/parent-2.4.0/retrofit/src/main/java/retrofit2/DefaultCallAdapterFactory.java#L23.

ferozco:
To me this is the expected behaviour, if you wanted to transfer the trace across threads you'd have to explicitly create a detached span or use your own executor

conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
```diff
@@ -245,18 +253,52 @@ public void testCborRequests() throws IOException, InterruptedException {
...
-    private void makeFutureRequest(Supplier<Future<String>> futureSupplier) {
+    private void makeListenableFutureRequest(Supplier<ListenableFuture<String>> futureSupplier) {
+        Tracer.initTrace(Observability.SAMPLE, Tracers.randomId());
+        String traceId = Tracer.getTraceId();
         String value = "value";
+
+        ListenableFuture<String> future = futureSupplier.get();
+
+        // Ensure that the trace ID is propagated
+        CompletableFuture<String> callbackTraceId = new CompletableFuture<>();
```
carterkozak, line 272: pedantic nit: Mind using a listenable future here for simplicity?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -70,18 +70,21 @@
...
      *         JVM shutdown to occur cleanly without waiting a full minute after the last request completes.
      *     </li>
+     *     <li>
+     *         Tracing information is propagated from the caller to the OkHtttp callback.
+     *     </li>
      * </ol>
      */
     private static final ExecutorService executionExecutor =
-            Executors.newCachedThreadPool(executionThreads);
+            Tracers.wrap("OkHttp: callback", Executors.newCachedThreadPool(executionThreads));
```
carterkozak, line 87: nice

carterkozak:
I think we should provide out-of-the-box tracing for async requests, otherwise you lose data converting from feign to retrofit.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
carterkozak: Actually, I'm not sure this works correctly. If memory serves, multiple operations may be run from a queue of pending requests here, but we'd end up changing the traceId.

We should add test coverage for queued requests individually retaining correct traceIds.

jkozlowski:
@ferozco I would expect this to interop correctly, at least mark the request correctly. Callbacks on that future with MoreExecutors.direct are a slightly different matter I guess

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
gatesn: I'm not sure I understand the case you're talking about. Are you say non-callback operations run from the same executor? Or that when requests queue, the traceIds might be off?

carterkozak: Two issues:
1. Operations on the dispatcher pool may drain a work queue in their execution, applying thread state more broadly than we expect. Regardless of how it works today, this type of behavior can be added at any time and may cause confusion in our tracing output.
2. When requests are queued, they're not queued on this executor, so the traceIds will not be what we expect.

jkozlowski: I think this may be similar to some work when trying to instrument a piece of code that uses netty? We also did not want to make netty aware of our Tracing infra. So we created a small executor for each traceId. In here, the RemotingOkhttpCall already captures the traces correctly, so we need to restore them appropriately at this layer? Alternatively, I see AsyncSerializableErrorCallAdapterFactory? If we set and reset the trace in #onResponse and #onFailure, then all the callbacks registered with MoreExecutors.direct should work?

gatesn:
Think we should attend to this in the new year

PR #1372, opened 2019-12-12
Jersey 2.25.1 -> 2.29, incl. many dependency changes

iamdanfox:
## Before this PR

As requested by @natgabb, who needs to use 2.26 for SSE things.

## After this PR
==COMMIT_MSG==
Jersey 2.25.1 -> 2.29
https://eclipse-ee4j.github.io/jersey.github.io/release-notes/2.26.html
==COMMIT_MSG==

> Unfortunately, there was a need to make backwards incompatible changes in 2.26. Concretely jersey-proprietary reactive client API is completely gone and cannot be supported any longer - it conflicts with what was introduced in JAX-RS 2.1 (that's the price for Jersey being "spec playground..").

## Possible downsides?
- jersey has been notorious for shipping API breaks - I think these upgrades have caused NoSuchMethodErrors in the past
- it may also surprise people that the javax.ws.rs-api jar will disappear, even if all the _classes_ should still be present (and come from the jakarta jar)


versions.props
```diff
@@ -14,14 +14,14 @@ com.palantir.conjure.verification:* = 0.18.2
...
-io.dropwizard:dropwizard-* = 1.3.7
+io.dropwizard:dropwizard-* = 2.0.0
 io.reactivex.rxjava2:rxjava = 2.2.15
 javax.ws.rs:javax.ws.rs-api = 2.1
 junit:junit = 4.12
 org.apache.commons:commons-lang3 = 3.9
 org.assertj:assertj-core = 3.14.0
 org.bouncycastle:bcpkix-jdk15on = 1.64
-org.glassfish.jersey.*:* = 2.25.1
+org.glassfish.jersey.*:* = 2.26
```
robert3005: 2.29?

changelog/@unreleased/pr-1372.v2.yml
```diff
@@ -0,0 +1,17 @@
...
+    allprojects {
+      dependencies {
+        modules {
+          module('javax.ws.rs:javax.ws.rs-api') {
+              replacedBy('jakarta.ws.rs:jakarta.ws.rs-api', 'Oracle have trademarks forced Eclipse to rename https://waynebeaton.wordpress.com/2019/04/04/renaming-java-ee-specifications-for-jakarta-ee/')
+          }
+        }
+      }
+    }
+    ```
```
iamdanfox, line 15: This is another snippet that I would really prefer not to proliferate into thousands of gradle projects internally. Most people don't actually have `com.palantir.baseline-class-uniqueness` running, so they probably won't notice, but it's a little bit spicy anyway.

Might be worth doing carter's lockfile suggestion

iamdanfox:
@pkoenig10 FYSA - I think once this merges, MP will be unable to pick up conjure-java-jersey-server releases until you have also upgraded dropwizard 1.3.7 -> 2.0.0.

CRogers (approved):

CRogers (approved):

PR #1374, opened 2019-12-12
Add param converters for authentication credentials

### LINKED ISSUE #1373: Invalid credentials should result in 401
## What happened?
When a client makes a request with invalid credentials, they receive a 400 response. This is especially problematic because conjure clients do not propagate 400s, meaning these request ultimately manifest as 500s in the calling service.

Invalid credentials should result in a 401 response.

This was the outcome of the discussion in https://github.com/palantir/auth-tokens/pull/228. This was implemented in conjure-undertow in https://github.com/palantir/conjure-java/pull/656.

pkoenig10:
## Before this PR
Malformed authentication credentials will result in a 400. This is unfortunate because it will propagate as a 500.

## After this PR
Malformed authentication credentials will result in a 401.

This change is the equivalent to part of https://github.com/palantir/conjure-java/pull/656 for Jersey servers.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
@@ -47,7 +47,7 @@ public final class AuthHeaderParamConverterProvider implements ParamConverterPro
         @Override
         public AuthHeader fromString(final String value) {
             if (value == null) {
-                throw UnauthorizedException.missingCredentials();
+                return null;
             }
             try {
                 return AuthHeader.valueOf(value);
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
@@ -13,16 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-/*
- * Adapted from https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jersey/src/main/java/io/dropwizard/jersey/optional/OptionalIntParamConverterProvider.java
- */
 
 package com.palantir.conjure.java.server.jersey;
 
 import com.palantir.logsafe.Preconditions;
+import com.palantir.tokens.auth.AuthHeader;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
-import java.util.OptionalInt;
 import javax.ws.rs.ext.ParamConverter;
 import javax.ws.rs.ext.ParamConverterProvider;
 import javax.ws.rs.ext.Provider;
@@ -30,8 +27,8 @@ import org.glassfish.jersey.internal.inject.Custom;
 
 @Custom
 @Provider
-public final class Java8OptionalIntParamConverterProvider implements ParamConverterProvider {
-    private final OptionalIntParamConverter paramConverter = new OptionalIntParamConverter();
+public final class AuthHeaderParamConverterProvider implements ParamConverterProvider {
+    private final AuthHeaderParamConverter paramConverter = new AuthHeaderParamConverter();
 
     @Override
     @SuppressWarnings("unchecked")
@@ -39,27 +36,26 @@ public final class Java8OptionalIntParamConverterProvider implements ParamConver
             final Class<T> rawType,
             final Type _genericType,
             final Annotation[] _annotations) {
-        return OptionalInt.class.equals(rawType) ? (ParamConverter<T>) paramConverter : null;
+        return AuthHeader.class.equals(rawType) ? (ParamConverter<T>) paramConverter : null;
     }
 
-    public static final class OptionalIntParamConverter implements ParamConverter<OptionalInt> {
+    public static final class AuthHeaderParamConverter implements ParamConverter<AuthHeader> {
         @Override
-        public OptionalInt fromString(final String value) {
+        public AuthHeader fromString(final String value) {
             if (value == null) {
-                return OptionalInt.empty();
+                throw UnauthorizedException.missingCredentials();
             }
-
             try {
-                return OptionalInt.of(Integer.parseInt(value));
-            } catch (NumberFormatException e) {
-                throw new IllegalStateException(e);
+                return AuthHeader.valueOf(value);
+            } catch (RuntimeException e) {
+                throw UnauthorizedException.malformedCredentials(e);
             }
         }
 
         @Override
-        public String toString(final OptionalInt value) {
+        public String toString(final AuthHeader value) {
             Preconditions.checkArgument(value != null);
-            return value.isPresent() ? Integer.toString(value.getAsInt()) : "";
+            return value.toString();
         }
     }
 }
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
@@ -20,6 +20,8 @@ import com.palantir.logsafe.Preconditions;
 import com.palantir.tokens.auth.AuthHeader;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.ext.ParamConverter;
 import javax.ws.rs.ext.ParamConverterProvider;
 import javax.ws.rs.ext.Provider;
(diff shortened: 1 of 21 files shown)


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
```diff
@@ -0,0 +1,70 @@
...
+                ErrorType.Code.UNAUTHORIZED,
+                "Conjure:MissingCredentials");
+        private static final ErrorType MALFORMED_CREDENTIAL_ERROR_TYPE = ErrorType.create(
+                ErrorType.Code.UNAUTHORIZED,
+                "Conjure:MalformedCredentials");
+
+        @Override
+        public AuthHeader fromString(final String value) {
+            if (value == null) {
+                throw new UnauthorizedException(MISSING_CREDENTIAL_ERROR_TYPE);
```
pkoenig10: This needs to be a `WebApplicationException`. Otherwise it will get wrapped in a `ExtractorException` which will get mapped to a 400:
https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/AbstractParamValueExtractor.java#L137

The jaxrs `NotAuthorizedException` doesn't provide constructors that can be used in this scenario so we need to create our own. I chose the name `UnauthorizedException` to disambiguate.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/GuavaOptionalParamConverterProvider.java
```diff
@@ -24,10 +24,12 @@
 import javax.ws.rs.ext.ParamConverterProvider;
 import javax.ws.rs.ext.Provider;
 import org.glassfish.hk2.api.ServiceLocator;
+import org.glassfish.jersey.internal.inject.Custom;
 import org.glassfish.jersey.internal.inject.Providers;
 import org.glassfish.jersey.internal.util.ReflectionHelper;
 import org.glassfish.jersey.internal.util.collection.ClassTypePair;
 
+@Custom
```
pkoenig10, line 33: This annotation ensures that our custom param converters are considered first. 
Otherwise the order in which the param converter providers are considered is arbitrary:
https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/internal/inject/ParamConverterFactory.java#L58-L101

For example, here is an order that was used when I was debugging tests locally:
param-converter-providers

This is especially important for the `AuthHeaderParamConverterProvider` and `BearerTokenParamConverterProvider` to ensure that we don't fallback to the built-in `AggregatedProvider` which contains the `TypeValueOf` param converter (which is used currently for `AuthHeader` and `BearerToken`).

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/BearerTokenParamConverterProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
+import javax.ws.rs.ext.Provider;
+import org.glassfish.jersey.internal.inject.Custom;
+
+@Custom
+@Provider
+public final class BearerTokenParamConverterProvider implements ParamConverterProvider {
```
carterkozak, line 34: I'm not confident this is correct, while cookie auth uses the BearerToken type, we can define additional parameters of type BearerToken that aren't the conjure-defined auth token.

pkoenig10: I mention this in the PR description:
> These param converters will apply to all `AuthHeader` and `BearerToken` parameters. However it is very rare for these types to be used outside of the endpoint auth. And most of those cases are to perform some sort of delegate authentication, so a 401 is not inappropriate.

I don't think I've ever seen a `BearerToken` used as an argument outside of cookie auth. And the only times I've seen a `AuthHeader` used as argument outside of header auth is for delegating authorization, in which case you could argue that a 401 is not inappropriate.

I don't think there is another way to achieve the desired behavior. If you have other suggestions, I'm happy to hear them. But I think the benefit of responding with 401 for invalid auth credentials outweighs the minor downside of this applying to other parameters of the same type.

pkoenig10: I've updated this PR to only apply to Conjure auth parameters. This means:
- `AuthHeader` parameter must be annotated with `@HeaderParam` and have a value of `Authorization`
- `BearerToken` parameter must be annotated with `@CookieParam` and have any value

carterkozak: That sounds reasonable, I don't think we need the additional specificity for AuthHeader because conjure only uses that type for auth components, bearer tokens use the `BearerToken` type.

pkoenig10: I would prefer to not respond with 401 for parameters we know aren't the Conjure auth parameter. It's possible to have other `AuthHeader` header params by using external imports. We actually do this for a number of endpoints and use a second `AuthHeader` header param to represent a delegating user. It makes more sense to return 400 for these parameters since they are effectively arguments that just happen to be passed as headers.

Because Conjure doesn't allow arbitrary cookie parameters, this implementation will exactly match all auth parameters and nothing else.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+                ? (ParamConverter<T>) paramConverter
+                : null;
+    }
+
+    public static final class AuthHeaderParamConverter implements ParamConverter<AuthHeader> {
+        @Override
+        public AuthHeader fromString(final String value) {
+            if (value == null) {
+                throw UnauthorizedException.missingCredentials();
+            }
```
carterkozak, line 54: I think this is a break, currently conjure-jersey allows null auth token values. It shouldn't, which is why we fixed it in conjure-undertow, however there are a few projects which rely (or at one point relied) on tokens being nullable.

I am generally supportive of reducing friction migrating between frameworks, but we need to be careful about rollout.

pkoenig10:
@carterkozak are there any other changes we should make in this PR to try to gracefully handle applications that rely on tokens being nullable?

carterkozak:
Let's add test coverage to make sure this change doesn't modify null token behavior, this way we can decouple the risky piece from the type of exception thrown from our infrastructure. We can hold off on null handling until the office opens up and we can discuss more broadly :-)

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/BearerTokenParamConverterProvider.java
```diff
@@ -0,0 +1,74 @@
...
+        @Override
+        public String toString(final BearerToken value) {
+            Preconditions.checkArgument(value != null);
+            return value.toString();
+        }
+    }
+
+    private static boolean hasAuthAnnotation(Annotation[] annotations) {
+        for (Annotation annotation : annotations) {
+            if (annotation.annotationType() == CookieParam.class) {
```
carterkozak: nit:
```suggestion
            if (CookieParam.class.equals(annotation.annotationType())) {
```

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+        @Override
+        public String toString(final AuthHeader value) {
+            Preconditions.checkArgument(value != null);
+            return value.toString();
+        }
+    }
+
+    private static boolean hasAuthAnnotation(Annotation[] annotations) {
+        for (Annotation annotation : annotations) {
+            if (annotation.annotationType() == HeaderParam.class) {
```
carterkozak: nit:
```suggestion
            if (HeaderParam.class.equals(annotation.annotationType())) {
```

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
```diff
@@ -0,0 +1,78 @@
...
+                String value = ((HeaderParam) annotation).value();
+                if (value.equals(HttpHeaders.AUTHORIZATION)) {
```
carterkozak: nit, safer around nulls:
```suggestion
                if (HttpHeaders.AUTHORIZATION.equals(value)) {
```

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/GuavaOptionalParamConverterProvider.java
carterkozak: Something along these lines would make an excellent comment in the code :-)

carterkozak (commented):
Generally looks good to me, thanks! Just a couple small nits

carterkozak:
Thanks! 👍

PR #1390, opened 2020-01-07
[WIP] Verify SNI information is passed from client

ellisjoe:
## Before this PR
Unknown whether our default setup sends [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication) information as part of TLS handshakes.

## After this PR
==COMMIT_MSG==
Verify SNI information is passed from client
==COMMIT_MSG==

This is currently just a test that shows our default setups does _not_ send SNI information. You can run curl against the server to validate that it's working as expected. Internally we saw that using the Conscrypt security provider did indeed send SNI, but the default JDK security provider does not.


iamdanfox:
What's the motivation behind this btw? Necessary to appease kubernetes? My guess is that this would involve adding something to https://github.com/square/okhttp/blob/d28d2cec21641b61f3d34e05dd52f43a717c2d32/okhttp/src/main/java/okhttp3/internal/platform/Jdk9Platform.java#L38-L53

ellisjoe:
On a separate internal test I verified that SNI info is passed by default. The issue with the test here was that I was using localhost and the JDK requires the hostname to contain a `.` in order to include SNI info in the TLS handshake: https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/security/ssl/Utilities.java#L96

This was only opened to share the test with some other folks, so just going to close it out.

PR #1395, opened 2020-01-08
Add test demonstrating the deserialization behavior of JSON string body

pkoenig10:
## Before this PR
There was some confusion about the unexpected behavior when deserializing a JSON string body on an endpoints that consumes `application/json`. There is not currently a test that clearly demonstrates that this is the known behavior.

## After this PR
There is a a test that clearly demonstrates that this is known behavior.


CRogers:
:+1:

PR #1399, opened 2020-01-09
Resolve OkHttpClientsRealServerTest test flakes

carterkozak:
Provide a deterministic backoff function to ensure additional
requests aren't sent.

## Before this PR
Test flakes

## After this PR
==COMMIT_MSG==
Resolve OkHttpClientsRealServerTest test flakes
==COMMIT_MSG==

## Possible downsides?
Test might flake in other ways, if we get a bad GC this could fail.

### DIFF
--- a/okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/ExponentialBackoffTest.java
+++ b/okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/ExponentialBackoffTest.java
@@ -17,11 +17,9 @@
 package com.palantir.conjure.java.okhttp;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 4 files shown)


okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/ExponentialBackoffTest.java
```diff
@@ -33,15 +33,15 @@
     @Test
     public void testNoRetry() {
         Random random = mock(Random.class);
```
ferozco: seems a little weird to mock `Random` now that we're just passing it a supplier

ferozco (approved):

okhttp-clients/src/test/java/com/palantir/conjure/java/okhttp/ExponentialBackoffTest.java
carterkozak: That's fair, updated to use a simple AtomicDouble.

PR #1401, opened 2020-01-10
Improvement: Add method for constructing KeyManager out of SslConfiguration

robert3005:
## Before this PR
It's hard to augment sslconfiguration with additional certificates. While you can get access to trust manager you can't access keymanager (nor produce keymanager/keystore out of sslconfiguration) which makes altering ssl settings of the ssl sockets impossible 

## After this PR
==COMMIT_MSG==
Expose method for converting SslConfiguration into a KeyManager. Allows augmenting sslsockets used by conjure java runtime with additional certificates. 
==COMMIT_MSG==

## Possible downsides?
We're adding one  more method to the api, however, it's pretty low level and has very specialized usecase thus limiting possibility for abuse.

### DIFF
--- a/changelog/@unreleased/pr-1401.v2.yml
+++ b/changelog/@unreleased/pr-1401.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Expose method for converting SslConfiguration into a KeyManager. Allows
+    augmenting sslsockets used by conjure java runtime with additional certificates.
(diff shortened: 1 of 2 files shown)


dansanduleac (approved):

dansanduleac:
I'd remove the changelog though as I don't think this is useful to announce to regular users of conjure-java-runtime.

PR #1406, opened 2020-01-10
Improvement: Always include private key parser and provide PKCS#8 support

### LINKED ISSUE #1402: Support PKCS#8 private keys
Right now we only support pkcs#1 (asn.1 format) for private keys. Most tools these days will by default generate pkcs#8 which is the future

robert3005:
## Before this PR
Users had to explicitly include private key parser. The parser itself is either bouncycastle based or uses package private classes from jdk.

## After this PR
==COMMIT_MSG==
Provide standardised private key parser that's always included.
==COMMIT_MSG==

fixes #1402 

## Possible downsides?
Instead of depending on libraries we roll the parser ourselves but the complexity of it is very low and libraries providing this functionality are big, i.e. bouncycastle


keystores/src/test/java/com/palantir/conjure/java/config/ssl/TestConstants.java
```diff
@@ -118,10 +117,83 @@
...
+            + "BhAOxqCc3L7cY+PJljamBk78CgYEA1Ju6tligSsrbuwOJ7+FAAMYZ8BZwTwtGjNOqjv8TZMQnRgquiWWqd9M/QZtw5WubJ4GEUOoTs1j"
+            + "hS9e3MOWJtgB8VS4eDzaFx3f3dTPK7Vj0gE2x3M9u2wgSXz92cR+19U8IMvNR8j1coRCZFpRI0f/QDpyixGI/fsMCdMhJmfUCgYAlKUY"
+            + "LYQsgskuha/67jkEEYEslHPu9JGvOhQs33/H1gg85BP8oJSytA7Cb57P7+2fVtrvlIHa5tcywD+k99ziMRuvvrreFjaZaHfhXKecDjgN"
+            + "TvpqPrYt1jeE7dz6A6nSP5bCiaL144l0/FpMzSyBhX5pCz3Piax8dDxYThtSn0wKBgDpKN/amdcwLshKj9b1+1D1iEtpwTHpfbQhPQAP"
+            + "Ui0ylOt9JLnEdHo+bq0whgeJkjjVWnSBSDPSu+ZYBnogZZpgvqebwre0d1a/KhF3txUUtjlte8dTISmxi2iajD6YUO571/Bl7/Iudpzf"
+            + "BhTdbkFuC91gKOdX92cq3TIUYJJXBAoGAFKtc+vKYthDb6APyubyzHTkFDtZariotkVZBIFyaFW/2ERwduVMNySL3l3CKvo+ajrGwStK"
+            + "BtcpAxhR05Wi61UbVpXexYvV4GtH3ePzHexD5OAYPkj8k55RmEsJFiZhChZAD2GXbhxVyhj/zkoc834/EoO1sbA6Wdz0PLDXhc3Y=";
+    static final byte[] PKCS8_PRIVATE_KEY_DER = BaseEncoding.base64().decode(PKCS8_KEY_STRING);
+
+    static final String PKCS8_PRIVATE_KEY_TAGGED_STRING =
```
dansanduleac: why not _PEM?

dansanduleac:
So that your test setup is reproducible and understandable later, I'd also leave some comments in the TestConstants file explaining how you generated the PKCS#1 and PKCS#8 keys with openssl, and then how you extracted each of the components like the modulus via openssl from that key.

iamdanfox:
scripts > comments?

robert3005:
I think you can figure out how to generate those values using google.com or openssl docs. Not sure what you want in particular.

robert3005:
These values don't need to be ever recreated so if anyone is changing them they likely should understand what they're changing

dansanduleac:
👍

PR #1411, opened 2020-01-14
Improvement: Don't require password for keystores constructed from pems

### LINKED ISSUE #1405: Don't require password if private key is not encrypted
## What happened?

You always have to provide a password for keystore private key even if the original file is not encrypted

Password should be optional and provided by the user if necessary.

robert3005:
## Before this PR
We assumed that keystores came with passwords and required keystores constructred from pems to have passwords

## After this PR
==COMMIT_MSG==
Don't assume password is present on keystore and don't password encrypt constructed keystores
==COMMIT_MSG==

## Possible downsides?
This is strictly an improvement over existing behaviour where we always required user to provide a password even if the original material didn't have any.

fixes #1405

### DIFF
--- a/changelog/@unreleased/pr-1411.v2.yml
+++ b/changelog/@unreleased/pr-1411.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Don't assume password is present on keystore and don't password encrypt
+    constructed keystores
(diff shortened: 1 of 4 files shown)


CRogers (approved):
LGTM

PR #1412, opened 2020-01-14
RemotingOkHttpCall logs backoff state prior to retry

carterkozak:
## Before this PR
Not much helpful safe-logged data on retry.

## After this PR
==COMMIT_MSG==
RemotingOkHttpCall safe-logs backoff state prior to retry including the slot size, retry number, and max retries.
==COMMIT_MSG==

## Possible downsides?
string allocation, however in this case we're already in a suboptimal path, a string is the least of our worries.


iamdanfox (approved):
Good idea, thanks!

PR #1413, opened 2020-01-15
Concurrency limiter listeners are protected against extra calls

carterkozak:
## Before this PR
Previously, internal state could be modified by interacting with
the same listener multiple times.

## After this PR
==COMMIT_MSG==
Concurrency limiter listeners are protected against extra calls
==COMMIT_MSG==

## Possible downsides?
It's possible this will make existing leaks worse because extra releases can not counteract leaked resources.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -333,7 +333,49 @@ final class ConcurrencyLimiters {
                 }
             };
             leakDetector.register(result, allocationStackTrace);
-            return result;
+            return new AtMostOneInteractionListener(result);
+        }
+    }
+
+    /** Prevent multiple interactions with the same listener from modifying state in unexpected ways. */
+    @VisibleForTesting
+    static final class AtMostOneInteractionListener implements Limiter.Listener {
+
+        private final Limiter.Listener delegate;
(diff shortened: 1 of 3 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -333,7 +333,49 @@ public void onDropped() {
...
+
+        private final Limiter.Listener delegate;
+        private boolean completed;
+
+        AtMostOneInteractionListener(Limiter.Listener delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public void onSuccess() {
```
jkozlowski, line 352: Do you need some synchronization here?

carterkozak: No, limiters are not shared between threads so we don't need atomic/volatile variables. There's a happens-before relationship.

CRogers:
:+1: Looks reasonable to me, although my experience with the limiters part of the codebase is limited.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
jkozlowski: Yeah, I had a look and it seemed that way, but better ask.

PR #1416, opened 2020-01-15
Improvement: Move KeepAliveSocketFactory

iamdanfox:
## Before this PR

@robert3005 is trying to rid the world of jks, but this involves constructing an SSL socket factory himself instead of using the one provided in `ClientConfigurations.of`. Unfortunately, this meant that the KeepAliveSocketFactory would not be accessible, leading to copypasta.

## After this PR
==COMMIT_MSG==
Keep-alives are now enabled at okhttpclient construction time, rather than at configuration time.  This means they can't be opted-out of.
==COMMIT_MSG==

## Possible downsides?
- you can't opt out of these anymore
- anyone who was grabbing the sslsocketfactory out of the ClientConfiguration and expecting it to have keep-alives will find they no longer have keep-alives set (e.g. dialogue???)

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ForwardingSslSocketFactory.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ForwardingSslSocketFactory.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ForwardingSslSocketFactory.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
robert3005: keep the old date

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/KeepAliveSslSocketFactory.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
```
robert3005: same as above

robert3005 (approved):

carterkozak:
I think this is a regression, I'm going to write some tests for this when I have a moment.
https://github.com/palantir/conjure-java-runtime/pull/1450#issuecomment-579780306

PR #1418, opened 2020-01-16
Improvement: Support multiple certificates per certificate file

robert3005:
## Before this PR
We assumed certificate files had one certificate while in practice our infra and other tooling produces multiple per file

## After this PR
==COMMIT_MSG==
Support certificate files with multiple certificates per file
==COMMIT_MSG==

## Possible downsides?
We change the aliases in the produced files, could be fixed for single certificate files

### DIFF
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
@@ -92,12 +92,7 @@ final class KeyStores {
 
         for (File currFile : getFilesForPath(path)) {
             try (InputStream in = Files.newInputStream(currFile.toPath())) {
-                List<Certificate> certificates = readX509Certificates(in);
-                int certIndex = 0;
-                for (Certificate cert : certificates) {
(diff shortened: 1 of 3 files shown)


keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
```diff
@@ -92,7 +92,12 @@ static KeyStore createTrustStoreFromCertificates(Path path) {
 
         for (File currFile : getFilesForPath(path)) {
             try (InputStream in = Files.newInputStream(currFile.toPath())) {
-                keyStore.setCertificateEntry(currFile.getName(), readX509Certificate(in));
+                List<Certificate> certificates = readX509Certificates(in);
+                int certIndex = 0;
+                for (Certificate cert : certificates) {
+                    keyStore.setCertificateEntry(currFile.getName() + certIndex, cert);
```
dansanduleac: can we add a dash before the certIndex, and pull this out to a function?

dansanduleac:
👍 approved waiting on requested change

PR #1420, opened 2020-01-16
Tag internal error metric with cause

parker89:
This PR tags the internal error metric with either:

- "serviceInternal" for service exceptions thrown by this service
- "internal" for generic exceptions thrown by this service
- "rpc" for errors communicating with a remote service (in particular, these are Feign RetryableExceptions wrapping SafeIOExceptions) 

The primary goal of this PR is to split out logic errors caused by a service from RPC caused errors. For example we'd like to page or rollback if a service starts throwing nullPointerExceptions, but not if it's throwing RetryableExceptions due to a dependency being down. 

Follow up to this PR: https://github.com/palantir/conjure-java-runtime/pull/1336

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/IllegalArgumentExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/IllegalArgumentExceptionMapper.java
@@ -33,6 +33,6 @@ final class IllegalArgumentExceptionMapper extends JsonExceptionMapper<IllegalAr
 
     @Override
     ErrorCause getCause() {
-        return ErrorCause.OTHER;
+        return ErrorCause.UNKNOWN;
     }
 }
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/IllegalArgumentExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/IllegalArgumentExceptionMapper.java
@@ -32,7 +32,7 @@ final class IllegalArgumentExceptionMapper extends JsonExceptionMapper<IllegalAr
     }
 
     @Override
-    InternalErrorCause getCause() {
-        return null;
+    ErrorCause getCause() {
+        return ErrorCause.OTHER;
     }
 }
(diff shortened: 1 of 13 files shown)


parker89:
Added a getter method so the JerseyServerMetrics are accessible from the server, so these metrics can be used to build a health check.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/IllegalArgumentExceptionMapper.java
```diff
@@ -16,19 +16,23 @@
...
     }
 
     @Override
     ErrorType getErrorType(IllegalArgumentException _exception) {
         return ErrorType.INVALID_ARGUMENT;
     }
+
+    @Override
+    InternalErrorCause getCause() {
+        return null;
```
ferozco: seems pretty sketchy to return null here since we assume non-null pretty widely in the code base. Perhaps add another variant to the enum?

parker89: Fair point. I don't expect this to be called though. Could I throw an `UnsupportedOperationException` here instead ?

ferozco: I think i'd prefer you return something concrete so that a code change in the abstract class doesn't inadvertently break this implementation

parker89: Ok, updated the class to be just `ErrorCause` and added an `OTHER` enum value, which we won't expect to be used.

carterkozak: fly-by nit: perhaps `UNKNOWN` for uniformity with conjure?

parker89: Changed to `UNKNOWN`

ferozco (approved):
Thanks @parker89!

PR #1421, opened 2020-01-16
LeakDetector creates stack traces dynamically for 5 minutes after a leak is detected

carterkozak:
## Before this PR
Debugging concurrency limiter leaks required iteration, for configuration changes.

## After this PR
==COMMIT_MSG==
LeakDetector creates stack traces dynamically for 5 minutes after a leak is detected
==COMMIT_MSG==

## Possible downsides?
Creating exceptions is expensive, but not nearly as expensive as waiting for leaked resources.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
@@ -58,6 +58,7 @@ final class ConcurrencyLimiters {
     private final Class<?> serviceClass;
     private final ScheduledExecutorService scheduledExecutorService;
     private final boolean useLimiter;
+    private final Meter concurrencyLimiterLeakMeter;
 
     @VisibleForTesting
     ConcurrencyLimiters(
@@ -74,6 +75,8 @@ final class ConcurrencyLimiters {
         this.serviceClass = serviceClass;
         this.scheduledExecutorService = scheduledExecutorService;
         this.useLimiter = useLimiter;
+        this.concurrencyLimiterLeakMeter =
+                OkhttpMetrics.of(taggedMetricRegistry).leakDetection("ConcurrencyLimiters");
     }
 
     ConcurrencyLimiters(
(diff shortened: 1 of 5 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/LeakDetector.java
```diff
@@ -108,4 +101,15 @@ private void logLeak(Optional<RuntimeException> stackTrace) {
             this.stackTrace = stackTrace;
         }
     }
+
+    static boolean isTraceLoggingEnabled() {
+        return log.isTraceEnabled();
+    }
+
+    interface Reporter {
```
iamdanfox, line 109: I was a a little confused by why this Reporter interface is actually necessary - this whole LeakDetector thing is only ever used in the ConcurrencyLimiters class, so defining an inline anonymous implementation there feels a little weird.

What about moving the implementation into this file, or even squashing some of the logic into this actual class?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/LeakDetector.java
```diff
@@ -108,4 +101,15 @@ private void logLeak(Optional<RuntimeException> stackTrace) {
...
+
+        void onLeakDetected(Optional<RuntimeException> stacktrace);
+
+        boolean createStackTrace();
```
iamdanfox, line 113: this name was a little weird to me - kinda sounded like it might be actually doing the creation under the hood, but actually it's just a pure getter used in if condition. `shouldCreateStackTrace` or `isStackTraceCreationEnabled`?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/ConcurrencyLimiters.java
```diff
@@ -209,12 +212,32 @@ public void onDropped() {}
...
-        private final LeakDetector<Limiter.Listener> leakDetector = new LeakDetector<>(Limiter.Listener.class);
+        private final LeakDetector<Limiter.Listener> leakDetector;
 
         DefaultConcurrencyLimiter(Key limiterKey, Supplier<SimpleLimiter<Void>> limiterFactory) {
             this.limiterKey = limiterKey;
             this.limiterFactory = limiterFactory;
             this.limiter = limiterFactory.get();
+            this.leakDetector = new LeakDetector<>(
+                    Limiter.Listener.class,
+                    // Create stack traces for five minutes after a leak is detected
```
iamdanfox, line 223: also worth updating the log lines in the LeakDetector class as this is now unnecessary?

> To get a stack trace for the call where the acquire happened, set log level to TRACE.

iamdanfox (approved):
+1 to the idea!

PR #1431, opened 2020-01-21
Apply concurrency limiting only to network requests

gatesn:
## Before this PR
We do a lot of work to ensure most of our long-lived requests hit the OkHttp disk cache. However I suspect we're still being rate-limited client-side even with a cache hit.

## After this PR
Concurrency limits only apply to the network request, not the full request.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -216,7 +216,7 @@ public final class OkHttpClients {
 
         // Intercept calls to augment request meta data
         if (enableClientQoS) {
-            client.addInterceptor(new ConcurrencyLimitingInterceptor());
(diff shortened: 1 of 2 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -216,7 +216,7 @@ private static RemotingOkHttpClient createInternal(
 
         // Intercept calls to augment request meta data
         if (enableClientQoS) {
-            client.addInterceptor(new ConcurrencyLimitingInterceptor());
+            client.addNetworkInterceptor(new ConcurrencyLimitingInterceptor());
```
carterkozak, line 219: Looks reasonable to me, thanks

carterkozak:
👍

carterkozak:
Oh hmm, I think we should revert this, sorry. The interceptor only releases limiters, we don't want to _not_ release limiters for local resources.

PR #1432, opened 2020-01-21
Revert "Apply concurrency limiting only to network requests (#1431)"

carterkozak:
This reverts commit c445d7ec05f92020de8bcd3dd4dfe9b28bfdda61.

Patch would prevent limiters for cached resources from being released.

==COMMIT_MSG==
Revert concurrency limiter interceptor order change which could prevent cached resources from releasing concurrency limiters.
==COMMIT_MSG==


dansanduleac:
👍

PR #1434, opened 2020-01-22
Make JerseyServerMetrics public

parker89:
Make JerseyServerMetrics public for use in server code

### DIFF
--- a/conjure-java-jersey-server/src/main/metrics/jersey-server-metrics.yml
+++ b/conjure-java-jersey-server/src/main/metrics/jersey-server-metrics.yml
@@ -1,6 +1,6 @@
 options:
   javaPackage: com.palantir.conjure.java.server.jersey
-  javaVisibility: packagePrivate
+  javaVisibility: public
 namespaces:
   server.response:
     shortName: JerseyServer


dansanduleac (approved):

carterkozak:
Wait I don't understand why this should be public, this will cause a revapi break any time we update metrics

ferozco:
I don't think this needs to be public since the also move metric schema codegen into our API. If you need to consume the metric externally, can you not construct the metric name yourself?

parker89:
I could definitely construct the metric name on my own, but I assumed part of the point of the new metric-schema library was to make using metrics more structured? If I still use the metric name in the server code, isn't it basically considering it public? Except without compile time warnings if the metric changes? Similarly, we could leave it private here and I could just use reflection to access it, which would achieve a similar result. 

There are some other public metrics here: https://github.com/palantir/conjure-java-runtime#metrics

Is the concern adding any public metrics to the API? Or just this one in particular?

carterkozak:
The utility function used to create metrics should not be part of public api, by design it is meant not to be backwards compatible: https://github.com/palantir/metric-schema/blob/develop/docs/design.md#staged-builders

Even using the tritium/codahale metric names in code is risky because we assume we can add new tags to provide more precision over time. This works great if we're querying something like datadog or prometheus, however direct lookups based on a name and set of tags will stop working.

In general I don't think it's a good idea for an application to be both a producer and consumer of its own metrics. Perhaps we want some sort of subscriber which may be connected to both metrics and a health registry?

parker89:
What's wrong with an application consuming its own metrics?  I was primarily following the example set here:
https://github.palantir.build/foundry/witchcraft/blob/develop/witchcraft-core/src/main/java/com/palantir/witchcraft/health/HostMetricsServiceStatus.java#L38
Where witchcraft uses the `HostMetrics` from CJR: https://github.com/palantir/conjure-java-runtime/blob/develop/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DefaultHostMetrics.java

carterkozak:
DefaultHostMetrics creates codahale metric objects internally, but they're for convenience and aren't connected to a registry.
I don't think the HostMetricsServiceStatus example is a good idea, we should try not to spread badness.

PR #1438, opened 2020-01-24
Provide client-side warnings on deprecated endpoint invocations

markelliot:
## Before this PR
Clients invoking deprecated endpoints have no local signaling about those deprecations

## After this PR
Clients invoking deprecated endpoints will log a warning indicating which service class and, when available, the details of the server indicating access to a deprecated endpoint.

==COMMIT_MSG==
Provide client-side warnings on deprecated endpoint invocations
==COMMIT_MSG==

## Possible downsides?
More logging traffic, possible/nominal performance hit for checking one more header.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
@@ -23,6 +23,13 @@ import okhttp3.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * An interceptor that logs warnings when the response from a server contains the "deprecation" header. Logs include
+ * the content of the "server" header when it is provided, and always include the provided {@code serviceClassName} so
+ * that consumers applying this interceptor can locate the cause of the deprecation.
+ *
+ * <p>Note: endpoint information is not included because endpoint-level details are not available at this level.
+ */
 final class DeprecationWarningInterceptor implements Interceptor {
     private static final Logger log = LoggerFactory.getLogger(DeprecationWarningInterceptor.class);
     private final String serviceClassName;
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
@@ -49,7 +49,7 @@ final class DeprecationWarningInterceptor implements Interceptor {
 
         String deprecationHeader = response.header("deprecation");
         if (deprecationHeader != null) {
-            clientMetrics.deprecations(serviceClassName);
+            clientMetrics.deprecations(serviceClassName).mark();
 
             if (loggingRateLimiter.tryAcquire(1)) {
                 log.warn(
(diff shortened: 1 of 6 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private DeprecationWarningInterceptor(String serviceClassName) {
+        this.serviceClassName = serviceClassName;
+    }
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Response response = chain.proceed(chain.request());
+
+        if (response.isSuccessful()) {
```
carterkozak: Do we want to limit ourselves to successful responses? Failures may become more likely as changes to deprecate endpoints can lack sufficient test coverage with cross-version compatibility in mind.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+            String deprecationHeader = response.header("deprecation");
+            if (deprecationHeader != null) {
+                log.warn("Using a deprecated endpoint when connecting to service",
+                        SafeArg.of("serviceClass", serviceClassName),
+                        SafeArg.of("service", response.header("server", "no server header provided")));
```
carterkozak, line 58: This warning will be logged on every request, thoughts on a rate limiter?
We should mark a metric each time.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
markelliot: fair

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/DeprecationWarningInterceptor.java
markelliot: put in a once-per-minute rate limiter and added metrics

ferozco (commented):
code change seems fine, but I'm a bit concerned about the unnecessary noise in some cases. For example,a service could be unable to move off of a deprecated endpoint since making the change would require them to rely on a new version of their dependency which is not widely deployed. 

With our existing logging infrastructure its pretty easy for a service owner to see who is calling their deprecated endpoints and follow up to migrate them. With this change we push some of the responsibility for fixing deprecation to the caller although It seems desirable to put the pebble in the shoe of the person making the deprecations rather than the person using what was previously totally valid API.

ferozco:
Perhaps another way to handle this would be to teach an excavator how to upgrade API bindings in a way that doesn't block deployment so that users can get depredations at compile time

markelliot:
I think the compile-time tooling would be a nice addition but not a substitution because:
 1. We have clients that use Conjure that do not run code managed in repos we excavate
 2. Making code up-to-date is not the same as making releases up-to-date, and there's likely to be at least some lag in the latter, which provides motivation for ensuring that signal makes it to consumers of this library

markelliot:
Besides that, even with our logging infrastructure, we've had many cases lately where the clients were failing to provide good enough user-agent strings for us to trace the source -- localizing the signal to the clients is meant to help _in addition_ to the server-side signal, which I think we should keep.

ferozco:
Ok sure. I'm mostly concerned about the noise that all of the logs will produce. Could we start with just the metric?

markelliot:
In theory we should have few deprecated endpoint calls, if any, and where we do have them a throttled log indicating you’re making them doesn’t seem that noisy?

Can remove the logging if it’s a deal breaker though.

ferozco:
👍 its fine, if people complain we can change it

ferozco:
Thanks mark!

PR #1443, opened 2020-01-25
Update Jersey auth converters to respect nullable annotations

carterkozak:
## Before this PR
This resolves a regression when new handling was added to produce
401 response codes on missing and malformed credentials. Existing
consumers who relied on explicitly marked `@Nullable` paramters
will no longer regress.

Unfortunately this took longer than expected to detect as some projects are still working on upgrading to newer Jersey :-)

## After this PR
==COMMIT_MSG==
Update Jersey auth converters to respect nullable annotations
==COMMIT_MSG==

## Possible downsides?
It's poor practice to rely on `null` in apis, optional types should be preferred. If this were not a regression, we would not allow the change.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
@@ -52,6 +52,16 @@ public final class AuthHeaderParamConverterProvider implements ParamConverterPro
 
         private final boolean nullable;
 
+        /**
+         * This class should not be used directly.
+         *
+         * @deprecated Use AuthHeaderParamConverterProvider
+         */
+        @Deprecated
+        public AuthHeaderParamConverter() {
+            this(false);
+        }
+
         private AuthHeaderParamConverter(boolean nullable) {
             this.nullable = nullable;
         }
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
@@ -34,8 +34,8 @@ import org.glassfish.jersey.internal.inject.Custom;
 @Custom
 @Provider
 public final class AuthHeaderParamConverterProvider implements ParamConverterProvider {
-    private final AuthHeaderParamConverter nonNullParamConverter = new AuthHeaderParamConverter(false);
-    private final AuthHeaderParamConverter nullableParamConverter = new AuthHeaderParamConverter(true);
+    private static final AuthHeaderParamConverter nonNullParamConverter = new AuthHeaderParamConverter(false);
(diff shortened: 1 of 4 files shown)


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
```diff
@@ -33,21 +34,44 @@
 @Custom
 @Provider
 public final class AuthHeaderParamConverterProvider implements ParamConverterProvider {
-    private final AuthHeaderParamConverter paramConverter = new AuthHeaderParamConverter();
+    private final AuthHeaderParamConverter nonNullParamConverter = new AuthHeaderParamConverter(false);
```
ferozco: should these be static

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/BearerTokenParamConverterProvider.java
```diff
@@ -32,21 +33,44 @@
 @Custom
 @Provider
 public final class BearerTokenParamConverterProvider implements ParamConverterProvider {
-    private final BearerTokenParamConverter paramConverter = new BearerTokenParamConverter();
+    private final BearerTokenParamConverter nonNullParamConverter = new BearerTokenParamConverter(false);
```
ferozco: same here

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/AuthTest.java
```diff
@@ -195,6 +234,11 @@ public String getAuthHeader(AuthHeader value) {
             return value.toString();
         }
 
+        @Override
+        public String getAuthHeaderNullable(@Nullable AuthHeader value) {
+            return value == null ? "" : value.toString();
```
ferozco: could we use a descriptive value instead of empty string to denote the header was null?

ferozco (commented):
Thanks carter! Not sure if its worth it, but it might be worth trying to de-dupe the logic across bearerToken and authHeader converter providers

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/AuthHeaderParamConverterProvider.java
carterkozak: oof, good catch

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/AuthTest.java
carterkozak: Good call

carterkozak:
> Not sure if its worth it, but it might be worth trying to de-dupe the logic across bearerToken and authHeader converter providers

I don't think there's much duplicated complexity so I don't think it's worth combining the two at this point (rule of three?) but I could be convinced otherwise.

ferozco:
👍

carterkozak:
Thanks for the review!

PR #1445, opened 2020-01-28
Http2 requires gcm ciphers to be enabled

carterkozak:
## Before this PR

Previously http/2 could be negotiated with servers that failed
to fully implement the http/2 specification as defined in
https://tools.ietf.org/html/rfc7540#section-9.2.2

## After this PR
==COMMIT_MSG==
Http2 requires gcm ciphers to be enabled. Previously http/2 could be negotiated with servers that failed to fully implement the http/2 specification as defined in https://tools.ietf.org/html/rfc7540#section-9.2.2
==COMMIT_MSG==

## Possible downsides?
In some scenarios it's possible clients may drop back from http/2 to http/1.1, particularly those connecting to an nginx instance rather than another java service.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -44,6 +44,7 @@ import okhttp3.ConnectionSpec;
 import okhttp3.Credentials;
 import okhttp3.Dispatcher;
 import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
 import okhttp3.TlsVersion;
 import okhttp3.internal.Util;
 import org.slf4j.Logger;
@@ -244,6 +245,12 @@ public final class OkHttpClients {
 
         // cipher setup
         client.connectionSpecs(createConnectionSpecs(config.enableGcmCipherSuites()));
+        // gcm ciphers are required for http/2 per https://tools.ietf.org/html/rfc7540#section-9.2.2
+        // some servers fail to implement this piece of the specification, which can violate our
+        // assumptions.
+        if (!config.enableGcmCipherSuites()) {
+            client.protocols(ImmutableList.of(Protocol.HTTP_1_1));
+        }
 
         // increase default connection pool from 5 @ 5 minutes to 100 @ 10 minutes
         client.connectionPool(connectionPool);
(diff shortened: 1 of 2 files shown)


carterkozak:
This is difficult to test because most services we deploy implement the spec properly...

schlosna (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -244,6 +245,12 @@ private static RemotingOkHttpClient createInternal(
 
         // cipher setup
         client.connectionSpecs(createConnectionSpecs(config.enableGcmCipherSuites()));
+        // gcm ciphers are required for http/2 per https://tools.ietf.org/html/rfc7540#section-9.2.2
```
iamdanfox, line 248: I thought that RFC was more prescribing the behaviour of servers - it seems like they don't define the `deployment` term in https://tools.ietf.org/html/rfc7540#page-6, but do use `endpoint` to mean `Either the client or server of the connection.`

I think I'd almost prefer to have the hard off switch for HTTP2, and just use put a log.warn here instead?

carterkozak: The rfc is describing servers, but not all servers implement it properly. This leads to confusion when we assume clients with gcm ciphers disabled can't use http/2.

Happy to put up an implementation with an http/2 flag as well, but I'd prefer if we implemented both.

carterkozak: api portion of the http2 flag: https://github.com/palantir/conjure-java-runtime-api/pull/443

iamdanfox (approved):
OK sounds good - should help pds-109228

iamdanfox:
@felixdesouza @jeremyk-91 just a though - I know you guys are currently relying on http2 for service -> timelock communication. Does atlasdb enable GCM cipher suites for the c-j-r clients it creates?

If not I think this will force you guys down to http/1

carterkozak:
WC server already implements the protocol correctly so this will not impact timelock. This change is only relevant for servers which do not implement the expected gcm cipher check.

felixdesouza:
Yep we do: https://github.com/palantir/atlasdb/blob/812f83e318961318aa24ad0f9a29dc421d8bbe44/atlasdb-conjure/src/main/java/com/palantir/atlasdb/http/v2/ClientOptions.java#L97

PR #1446, opened 2020-01-28
Instrument OkHttp client handshakes

carterkozak:
## Before this PR
No client-side handshake instrumentation

## After this PR
==COMMIT_MSG==
Instrument OkHttp client handshakes
==COMMIT_MSG==

## Possible downsides?
metrics aren't free


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -209,7 +211,11 @@ private static RemotingOkHttpClient createInternal(
         client.followRedirects(false); // We implement our own redirect logic.
 
         // SSL
-        client.sslSocketFactory(new KeepAliveSslSocketFactory(config.sslSocketFactory()), config.trustManager());
+        SSLSocketFactory sslSocketFactory = MetricRegistries.instrument(
+                config.taggedMetricRegistry(),
+                new KeepAliveSslSocketFactory(config.sslSocketFactory()),
+                serviceClass.getSimpleName());
```
sfackler, line 224: Going by how we handle this on the server side, should this just be `"client"` rather than the service class's name?

carterkozak, line 224: We could use `"client"` here, but I think it's helpful to differentiate our clients. We could alternatively do this wrapping ClientConfiguration creation with a SslSocketFactory instead, retaining the ServiceConfiguration name, and falling back to `"client"` when that information is unavailable. That would mean plumbing a service name through api bits.

carterkozak: The server uses "server" because all incoming connections are the same, we already have tags for the local service name. A single service may have many outgoing clients, so it may be helpful to see that FooService client is negotiating GCM ciphers, while BarService is not.

sfackler: We'll have to use the ServiceConfiguration name to handle this in conjure-rust-runtime, but it probably doesn't matter all that much that the tag name is exactly the same between different services.

iamdanfox:
Ok so we're expecting this to produce one extra `tls.handshake` meter, tagged by `context: com.palantir.foo.FooService`, `protocol: tlsv1.2`, `cipher: ...`

This will cause a bit more DD $ spend, but since it's only tagged per _client_ instead of per _endpoint_ it shouldn't be _too_ bonkers.

Let me try and put an exact $ cost on it...

iamdanfox:
Ok so the conjure_java_client.qos.slow_acquire_tagged metric is also tagged by service class, with so I looked at the metric explorer tab and found the number of distinct series, which multiplied by our price per metric comes out reasonably. Seems OK for an entire fleet.

iamdanfox (approved):

PR #1450, opened 2020-01-29
Update the conjure okhttp client connection pool

carterkozak:
1. Allow up to 1000 idle connections, increased from 100. Many services
   constantly saturate over 250 threads waiting on RPC from other services,
   unless http/2 is used this model cannot effectively reuse connections.
   This isn't a good reason to use a more complex protocol with more overhead
   and lower throughput.
2. Decreased the idle connection keep-alive duration from ten minutes to
   fifty-five seconds to avoid a race condition where the server closes
   an idle connection before the server. By reducing this timeout we are
   able to provide a higher maximum idle connection count without as much
   risk of holding resources unnecessarily long.

## After this PR
==COMMIT_MSG==
Update the conjure okhttp client connection pool to allow up to 1000 idle connections (increased from 100) and reduce the keepalive time to 55 seconds (from 10 minutes). This should provide greater throughput in RPC bound environments.
==COMMIT_MSG==

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -85,7 +85,12 @@ public final class OkHttpClients {
     private static final Dispatcher dispatcher;
 
     /** Shared connection pool. */
-    private static final ConnectionPool connectionPool = new ConnectionPool(100, 10, TimeUnit.MINUTES);
+    private static final ConnectionPool connectionPool = new ConnectionPool(
+            1000,
+            // Most servers use a one minute keepalive for idle connections, by using a shorter keepalive on
+            // clients we can avoid race conditions where the attempts to reuse a connection as the server
+            // closes it, resulting in unnecessary I/O exceptions and retrial.
+            55, TimeUnit.SECONDS);
 
     private static DispatcherMetricSet dispatcherMetricSet;
 
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -90,7 +90,8 @@ public final class OkHttpClients {
             // Most servers use a one minute keepalive for idle connections, by using a shorter keepalive on
             // clients we can avoid race conditions where the attempts to reuse a connection as the server
             // closes it, resulting in unnecessary I/O exceptions and retrial.
-            55, TimeUnit.SECONDS);
+            55,
+            TimeUnit.SECONDS);
 
     private static DispatcherMetricSet dispatcherMetricSet;
 
--- a/changelog/@unreleased/pr-1450.v2.yml
+++ b/changelog/@unreleased/pr-1450.v2.yml
@@ -1,0 +1,7 @@
+type: improvement
+improvement:
+  description: Update the conjure okhttp client connection pool to allow up to 1000
+    idle connections (increased from 100) and reduce the keepalive time to 55 seconds
+    (from 10 minutes). This should provide greater throughput in RPC bound environments.
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/1450


carterkozak:
Ohhhh I think we introduced a regression recently

carterkozak:
So I think we played ourselves. The shared connection pool is great when we can reuse SSLSocketFactory instances, but wrapping `config.sslSocketFactory()` with `new KeepAliveSslSocketFactory(` means two clients cannot share connections anymore.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -85,7 +85,13 @@
     private static final Dispatcher dispatcher;
 
     /** Shared connection pool. */
-    private static final ConnectionPool connectionPool = new ConnectionPool(100, 10, TimeUnit.MINUTES);
+    private static final ConnectionPool connectionPool = new ConnectionPool(
+            1000,
+            // Most servers use a one minute keepalive for idle connections, by using a shorter keepalive on
+            // clients we can avoid race conditions where the attempts to reuse a connection as the server
+            // closes it, resulting in unnecessary I/O exceptions and retrial.
+            55,
```
iamdanfox, line 93: I was looking for the code for this WC, but isn't that 1 minute timeout for worker threads?

carterkozak: We don't override the default, though we probably should override it with itself for discoverability.

https://github.com/undertow-io/undertow/blob/062e01fbbf23195cde76a9d59e274fccd534ae00/core/src/main/java/io/undertow/UndertowOptions.java#L75-L78

https://github.com/undertow-io/undertow/blob/8b243e1e75812bcef15fc1ec4bf0a837ec8ce4bb/core/src/main/java/io/undertow/Undertow.java#L144-L147

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -85,7 +85,13 @@
     private static final Dispatcher dispatcher;
 
     /** Shared connection pool. */
-    private static final ConnectionPool connectionPool = new ConnectionPool(100, 10, TimeUnit.MINUTES);
+    private static final ConnectionPool connectionPool = new ConnectionPool(
+            1000,
```
iamdanfox, line 89: In WC's ServerConfiguration, I think we have an upper bound of 2560 concurrent connections.  If client nodes start maxing out their connection pools and try to open 1000 each, doesn't this mean that 2 client nodes could hog an entire server node?

Is this a risk we're willing to take (i.e. because we think these connections will not be maxed out the whole time, and hope that servers are horizontally scaled sufficiently to accommodate their possible clients?)

```java
    /** The maximum number of threads to use for requests. */
    @Value.Default
    @JsonProperty("max-threads")
    default int maxThreads() {
        return Math.max(256, 32 * processors());
    }

    @Value.Default
    @JsonProperty("max-connections") 
    default int maxConnections() {
        return 10 * maxThreads();
    }
```

carterkozak: I don't think this limit stops us from making more requests, it only stops us from saving the idle connections. In the heavily loaded case we should try to reuse connections as much as possible to avoid handshake work.

iamdanfox: argh you're absolutely right - that first param is `maxIdleConnections`! OK totally fine with dialling this up, as this isn't going to suddenly unleash any floodgates that were being capped. 

So we're really only expecting this to eliminate some possible blockers for very bursty traffic.

iamdanfox (approved):
Sounds good, and shouldn't cause any drastic changes to inter-service RPC!

FWIW I looked through our `dispatcher.calls.running` metric at a big deployment and found rare spikes of ~200 simultaneous calls, but most of the time they were actually under still 100.

carterkozak:
It's unfortunate we can't record metrics describing the connection pool size.

PR #1451, opened 2020-01-29
Test connections are shared between clients

carterkozak:
==COMMIT_MSG==
Test connections are shared between clients
==COMMIT_MSG==


iamdanfox (approved):
Ok so we're just proving that the static okhttp shared ConnectionPool actually behaves as expected?

carterkozak:
That's correct

PR #1452, opened 2020-01-30
Fix: Deduplicate certificates deserialized from same security config amongst different clients

robert3005:
## Before this PR
We would store certificates read from sslconfiguration once per client instead of once per certificate/configuration

## After this PR
==COMMIT_MSG==
Store certificates in static cache in KeyStores class (similar to X509Factory in jdk) that keys certificates by encoded representation of certificate.
==COMMIT_MSG==

### DIFF
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
@@ -16,6 +16,8 @@
 
 package com.palantir.conjure.java.config.ssl;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.base.Throwables;
 import com.google.common.io.BaseEncoding;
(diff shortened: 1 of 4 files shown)


keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
```diff
@@ -319,8 +328,9 @@ private static KeyStore createKeyStore() {
     }
 
     private static List<Certificate> readX509Certificates(InputStream certificateIn) throws CertificateException {
-        CertificateFactory certFactory = CertificateFactory.getInstance("X.509");
-        return new ArrayList<>(certFactory.generateCertificates(certificateIn));
+        return CertificateFactory.getInstance("X.509").generateCertificates(certificateIn).stream()
```
robert3005, line 331: The difference here is `CertificateFactory#generateCertificates` for X509 doesn't deduplicate certificates globally  while `CertificateFactory#generateCertificate` does

CRogers (approved):

PR #1454, opened 2020-01-31
Implement the `enable-http2` client configuration flag

carterkozak:
This allows us to quickly opt into or out of http2 in
debugging scenarios, however the flag is not recommended
in standard use.
https://github.com/palantir/conjure-java-runtime-api/pull/443

## Before this PR
No way to opt clients out of http/2 negotiation, making client http/2 protocol bugs difficult to validate.

## After this PR
==COMMIT_MSG==
Implement the `enable-http2` client configuration flag
==COMMIT_MSG==

## Possible downsides?
Larger configuration surface.


carterkozak:
Requires a cjr-api release, once we're happy with this we can push both sides through.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -60,6 +60,8 @@
     @VisibleForTesting
     static final int NUM_SCHEDULING_THREADS = 5;
 
+    private static final boolean DEFAULT_ENABLE_HTTP2 = true;
```
iamdanfox, line 63: I think all of the other default values are added by the `ClientConfigurations.of` method - any reason to do this stuff inside the client here?

carterkozak:
Yep, I want the freedom to use different default values depending on the client implementation.

-ck

iamdanfox (approved):

PR #1479, opened 2020-02-14
Fix retrofit ListanableFuture when propagating QoS

dansanduleac:
## Before this PR

https://github.com/palantir/conjure-java-runtime/pull/1041's logic to propagate Server QOS exceptions to the caller only works when calling `retrofit2.Call#execute()`, but not `retrofit2.Call#enqueue()` which is what happens when using `ListenableFutures`.
Conjure-java only generates ListenableFuture endpoint methods since 5.0, so this means that retrofit2 APIs generated by conjure-java 5 (or with ListenableFuture enabled) would always return a null body to the caller if:
* ClientConfiguration is configured with `ServerQoS.PROPAGATE_429_and_503_TO_CALLER`
* the server returns a 429 or 503

## After this PR
==COMMIT_MSG==
Retrofit2 clients now correctly propagate QoS exceptions when client is configured with `ServerQoS.PROPAGATE_429_and_503_TO_CALLER`
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1479.v2.yml
+++ b/changelog/@unreleased/pr-1479.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Retrofit2 clients now correctly propagate QoS exceptions when client
+    is configured with `ServerQoS.PROPAGATE_429_and_503_TO_CALLER`
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/1479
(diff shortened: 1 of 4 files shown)


ferozco:
👍 Nice, thanks @dansanduleac

iamdanfox:
meta-point: I think this is another example of where adding more configurability/options results in a higher testing & maintainability burden and more bugs :/

PR #1484, opened 2020-02-18
Improvement: Include default set of CAs when constructing TrustManagers

robert3005:
This has been lifted from internal project that wraps conjure-java-runtime. I have unfortunately seen multiple places that can't use that wrapper and use conjure-java-runtime directly. In order to unify certificates behaviour I want to move that logic into cjr.

## After this PR
==COMMIT_MSG==
Include default root CAs when creating trustmanagers
==COMMIT_MSG==

Once this is merged we can get rid of the copy we have internally

## Possible downsides?
This is one more level removed but these certificates shouldn't change often.


robert3005:
@dansanduleac the thing I talked to you about last Friday

robert3005:
@iamdanfox ping

iamdanfox (approved):

robert3005:
🚄 ?

PR #1501, opened 2020-02-25
Include UserAgent in ClientConfiguration

iamdanfox:
## Before this PR

We have a bunch of signatures in dialogue that take in a ClientConfiguration and a UserAgent. This seems a little overly verbose, as semantically I think a user-agent is one of the pieces of information that are necessary to make outgoing requests, and therefore has a good claim to belong in ClientConfiguration.

## After this PR
==COMMIT_MSG==
ClientConfiguration contains an optional<UserAgent> field
==COMMIT_MSG==

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -153,13 +153,17 @@ public final class OkHttpClients {
             Class<?> serviceClass) {
         boolean reshuffle =
                 !config.nodeSelectionStrategy().equals(NodeSelectionStrategy.PIN_UNTIL_ERROR_WITHOUT_RESHUFFLE);
-        ClientConfiguration config1 = ClientConfiguration.builder()
-                .from(config)
-                .userAgent(userAgent)
-                .build();
+        ClientConfiguration config1 =
+                ClientConfiguration.builder().from(config).userAgent(userAgent).build();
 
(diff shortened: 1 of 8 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -152,24 +153,21 @@ public static OkHttpClient create(
...
+                .build();
+
+        return createInternal(client, config1, hostEventsSink, serviceClass, RANDOMIZE, reshuffle, () ->
+                new ExponentialBackoff(config1.maxNumRetries(), config1.backoffSlotSize()));
     }
 
     @VisibleForTesting
     static RemotingOkHttpClient withStableUris(
-            ClientConfiguration config, UserAgent userAgent, HostEventsSink hostEventsSink, Class<?> serviceClass) {
+            ClientConfiguration config, HostEventsSink hostEventsSink, Class<?> serviceClass) {
```
iamdanfox, line 171: thoughts on also including the HostEventsSink thingy in ClientConfiguration?

dansanduleac: I kinda like the idea. It's a good place to add new optional config especially as we can provide defaults too, and reduce the method argument inflation that's been going on here.

dansanduleac (approved):

carterkozak:
Snazzy :-)

iamdanfox:
It certainly makes dialogue pretty https://github.com/palantir/dialogue/pull/431

PR #1519, opened 2020-03-04
Add smile object mapper factory methods

carterkozak:
## After this PR
==COMMIT_MSG==
Add smile object mapper factory methods. These are intentionally not applied to frameworks which require manual action to enable specific encodings (feign, jersey, etc) and are reserved for negotiation where they can be transparently tested (conjure-undertow+dialogue).
==COMMIT_MSG==

## Possible downsides?
Additional dependency on smile, like cbor, and larger API footprint on factory methods. Like CBOR support this is worth the trade-off, especially given the breadth of smile use internally.


ferozco:
👍

PR #1533, opened 2020-03-12
Don't share connection pools

j-baker:
Internally, we've seen a number of issues that appear to be related to connection pool reuse between clients. What we think we see is that a client is hanging onto a connection (which after some period is actually dead, so holding no resources), but the ConnectionPool itself manages a deque of connections, which it synchronizes access too. After long enough runtime, this queue becomes so long that accessing it times out.

Now, the curious thing here is that this connection pool reuse is totally pointless! Reuse relies on the respective connections having the same SSLSocketFactory. And, the tritium instrumentation we do does identity equals - and so this cannot possibly cause any reuse at all.

### DIFF
--- a/changelog/@unreleased/pr-1533.v2.yml
+++ b/changelog/@unreleased/pr-1533.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Don't share connection pools between clients.
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/1533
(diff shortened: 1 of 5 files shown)


carterkozak:
we should talk :-)

carterkozak:
Generally in favor of this, more metrics tests to fix, and I don't want to merge unless/until @ferozco is happy with it.

ferozco:
To summarize the issue and proposed solution:

We previously had a global connection pool which was conceptually segmented per client instance since connections could not be reused across clients due to their SSLSocketFactory s not being equal [link](https://github.com/square/okhttp/blob/parent-3.13.1/okhttp/src/main/java/okhttp3/Address.java#L181). This resulted in slow lookups into the connection pool and low rates of re-use.

We're changing to having a per client connection pool, which will solve the problem of SSLSocketFactory s being equal. We expect this change to improve performance since connection will be re-usesed and look ups will be faster. Importantly, We will still have a global dispatcher which will continue limit the number of concurrent requests per host to 256.

Does that sound correct?

j-baker:
Exactly

ferozco:
👍

PR #1534, opened 2020-03-13
Revert "Don't share connection pools (#1533)"

j-baker:
This reverts commit 8371185592614760d96467849abc88298bfc3235.

Looking at the code some more, while my argument is correct, a thread is also booted up to clean up the pool periodically. This unfortunately means that in cases where you have issues here (and e.g. create loads and loads of clients unnecessarily), you'll also end up with a thread explosion.


CRogers:
:+1:

PR #1539, opened 2020-03-19
GCM ciphers are enabled by default on JRE9 and beyond

carterkozak:
## Before this PR

On newer Java runtimes GCM ciphers outperform the current default
cipher (TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384) due to hardware support.

Enabling GCM by default must be coupled by disabling http/2 by
default in our OkHttp client to prevent existing consumers from
switching to a protocol with known implementation bugs (20x slower
upload performance using okhttp h2, connection leaks, etc).

## After this PR
==COMMIT_MSG==
GCM ciphers are enabled by default on JRE9 and beyond
==COMMIT_MSG==

## Possible downsides?
Risks:
While existing servers allow GCM ciphers to be negotiated, including
those on java 8, clients using CJR regardless of JRE version will
not negotiate GCM ciphers by default. This allows servers running
on java 8 to avoid performance pitfalls of GCM. However, most
internal services, particularly those with heavy throughput
requirements, have upgraded to JDK 11. Applications which have
already taken the upgrade to 11 will become both more secure and
more performant.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -29,6 +29,8 @@ import java.net.Proxy;
 import java.net.ProxySelector;
 import java.net.SocketAddress;
 import java.net.URI;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
@@ -44,7 +46,7 @@ public final class ClientConfigurations {
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(5);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
     private static final Duration DEFAULT_FAILED_URL_COOLDOWN = Duration.ZERO;
-    private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
+    private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = !isJava8();
     private static final boolean DEFAULT_FALLBACK_TO_COMMON_NAME_VERIFICATION = false;
     private static final NodeSelectionStrategy DEFAULT_NODE_SELECTION_STRATEGY = NodeSelectionStrategy.PIN_UNTIL_ERROR;
     private static final int DEFAULT_MAX_NUM_RETRIES = 4;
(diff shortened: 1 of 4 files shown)


carterkozak:
Calling this a break because it will impact lock, and possibly others: https://github.com/palantir/atlasdb/pull/4655

carterkozak:
@iamdanfox @ferozco thoughts?

Now I can't add anyone as a reviewer, so that's exciting :-]

iamdanfox:
> Enabling GCM by default must be coupled by disabling http/2 by
default in our OkHttp client to prevent existing consumers from
switching to a protocol with known implementation bugs (20x slower
upload performance using okhttp h2, connection leaks, etc).

We don't have an external writeup of these problems right? Was this specifically caused by okhttp's http2 code or was it a particular cipher suite? Perhaps we could leave an `https://internal-jira` link to the relevant PDS ticket?

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -44,7 +46,7 @@
     private static final Duration DEFAULT_WRITE_TIMEOUT = Duration.ofMinutes(5);
     private static final Duration DEFAULT_BACKOFF_SLOT_SIZE = Duration.ofMillis(250);
     private static final Duration DEFAULT_FAILED_URL_COOLDOWN = Duration.ZERO;
-    private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = false;
+    private static final boolean DEFAULT_ENABLE_GCM_CIPHERS = !isJava8();
```
iamdanfox, line 51: Ideally, it would be great to provide a little comment and link here showing how GCM cipher suite performance got a lot better in java9, otherwise future readers might not remember why we special-cased java8 here

ferozco:
👍

iamdanfox:
I think it's worth being super clear that a possible downside here is all the straggling repos which haven't yet declared compatibility with java11 will likely burn more CPU cycles if all their clients start asking to communicate using GCM cipher suites.

I think it might be worth sending a direct email to these teams so that they know the ultimatum is upgrade or expect slowdowns if they transmit a lot of data...

iamdanfox:
@dvogelbacher is your analysis product running on java11 everywhere now? ~Also @pkoenig10 can I check the state of MP?~ (confirmed MP is running on 11)

dvogelbacher:
@iamdanfox it is not yet running on java 11, we should discuss how to make it so.

iamdanfox:
~As per our zoom earlier, adding `do not merge` until the Atlas folks merge https://github.com/palantir/atlasdb/pull/4655~

iamdanfox:
Ok so atlasdb [0.198.7](https://github.com/palantir/atlasdb/releases/tag/0.198.7) explicitly sets the `http:2` default they were relying on, and it looks like it's been adopted in our internal atlas-replacement service: https://internal-github/foundry/a---/pull/2250/files.

I think given that the majority of services have adopted j11, the benefits of GCM ciphers will outweigh the downsides... any last reservations before we pull the trigger?

carterkozak:
I'm happy with it. I'll plan to merge later unless there are remaining concerns.

PR #1540, opened 2020-03-20
Implement a JaxRsClient factory delegating to Dialogue

carterkozak:
This provides a bridge for existing and human-generated client code to take advantage of the superior observability and stability provided by Dialogue.

==COMMIT_MSG==
Implement a JaxRsClient factory delegating to Dialogue
==COMMIT_MSG==


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/JaxRsClient.java
```diff
@@ -55,4 +64,23 @@ private JaxRsClient() {}
                         config,
                         serviceConfiguration -> create(serviceClass, userAgent, hostEventsSink, serviceConfiguration)));
     }
+
+    /**
+     * Creates a {@code T client} for the given dialogue {@link Channel}.
+     */
+    public static <T> T create(Class<T> serviceClass, ConjureRuntime runtime, Channel channel) {
```
carterkozak: This is the only new public API.

Note that ConjureRuntime and Channel are from dialogue-target which is intentionally an `implementation` dependency to prevent leaking access to dialogue to consumers which don't already have it.

conjure-java-server-verifier/src/test/java/com/palantir/verification/server/VerificationClients.java
```diff
@@ -32,6 +36,16 @@ public static VerificationClientService verificationClientService(VerificationCl
                 verificationClientRule.getClientConfiguration());
     }
 
+    public static VerificationClientService verificationClientServiceDialogue(
+            VerificationClientRule verificationClientRule) {
+        Channel channel = ApacheHttpClientChannels.create(ClientConfiguration.builder()
+                .from(verificationClientRule.getClientConfiguration())
+                .userAgent(getUserAgent())
```
iamdanfox: thoughts on just pushing this `.userAgent()` into the `getClientConfiguration()` method?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/JaxRsClient.java
iamdanfox: Pretty minor, but could we order these `Class<T>, Channel, Runtime` to match the generated static factory methods? https://github.com/palantir/conjure-java/blob/160460802a55fd8e976ed5f3b636c58942336bbb/conjure-java-core/src/integrationInput/java/com/palantir/product/EteServiceBlocking.java#L108

iamdanfox: also I think people may well get confused and try to put a dialogue interface into the serviceClass param, when this method actually expects a jaxrs-annotated one...

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/JaxRsClient.java
```diff
@@ -55,4 +63,23 @@ private JaxRsClient() {}
...
+        ObjectMapper objectMapper = ObjectMappers.newClientObjectMapper();
+        ObjectMapper cborObjectMapper = ObjectMappers.newCborClientObjectMapper();
```
iamdanfox: could these be static fields to allow sharing between clients?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+        }
+        // Tracing path template is informational only
+        if (PATH_TEMPLATE.equalsIgnoreCase(headerName)) {
+            return false;
+        }
+        return true;
+    }
+
+    private static String urlDecode(String input) {
+        if (input.contains("%")) {
```
iamdanfox: I think the string `foo+bar` should also be urldecoded to `foo bar`. It seems like URLDecoder.decode already preserves its input in the case where there are no `+` or `%` chars, so probably safe to delete this if condition?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+            try {
+                return URLDecoder.decode(input, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+                throw new SafeRuntimeException("Failed to decode path segment", e);
```
iamdanfox: Might be worth putting `input` as an UnsafeArg?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+        private final Response response;
+
+        DialogueResponseBody(Response response) {
+            this.response = response;
+        }
+
+        @Override
+        public Integer length() {
+            return response.getFirstHeader(HttpHeaders.CONTENT_LENGTH)
+                    .map(Ints::tryParse)
```
iamdanfox, line 201: interestingly the implementation in OkhttpClient also verifies this is > 0, which presumably allows clients to keep working (and allocating properly) even if some bizarre server returns content-length: -1

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+        INSTANCE;
+
+        @Override
+        public Response deserialize(Response response) {
+            return response;
+        }
+
+        @Override
+        public Optional<String> accepts() {
+            return Optional.empty();
```
iamdanfox, line 246: this is used by DefaultClients.call to set the `Accepts` header right? Could we put a little comment saying this header is already set by the feign request so we don't want to override it?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * {@link DialogueFeignClient} is an adapter from {@link feign.Client} to {@link Channel Dialogue Channel}
+ * taking advantage of the superior observability and stability provided by Dialogue.
```
iamdanfox, line 60: Worth mentioning the buffering behaviour for requests & responses up front? I think it's probably also worth describing in what scenarios it's appropriate to use this (i.e. if you have a hand-rolled jaxrs interface) because we'd really prefer people to use the conjure-java generated code if possible. I want people to understand this is not the _ideal_ client to use, but still better than okhttp.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+ */
+final class DialogueFeignClient implements feign.Client {
+
+    private static final String PATH_TEMPLATE = "hr-path-template";
+    private static final Splitter pathSplitter = Splitter.on('/').omitEmptyStrings();
+    private static final Splitter.MapSplitter querySplitter =
+            Splitter.on('&').omitEmptyStrings().withKeyValueSeparator('=');
```
iamdanfox, line 67: neat :)

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+    private final String serviceName;
+    private final String version;
+
+    DialogueFeignClient(Class<?> jaxrsInterface, Channel channel, ConjureRuntime runtime, String baseUrl) {
+        this.channel = Preconditions.checkNotNull(channel, "Channel is required");
+        this.serviceName = jaxrsInterface.getSimpleName();
+        this.version = Optional.ofNullable(jaxrsInterface.getPackage().getImplementationVersion())
+                .orElse("0.0.0");
+        this.baseUrl = baseUrl;
+        this.clients = runtime.clients();
```
iamdanfox: could we just store the whole ConjureRuntime as a field? on the first read through the `Clients clients` thing made me worry this was doing some kind of internal caching!

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+    private static final Splitter.MapSplitter querySplitter =
+            Splitter.on('&').omitEmptyStrings().withKeyValueSeparator('=');
+
+    private final Clients clients;
+    private final Channel channel;
+    private final String baseUrl;
+    private final String serviceName;
+    private final String version;
+
+    DialogueFeignClient(Class<?> jaxrsInterface, Channel channel, ConjureRuntime runtime, String baseUrl) {
```
iamdanfox, line 75: given that we always pass in `baseUrl` of `dialogue://feign`, could we ditch this param & field and just make it a constant?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+            }
+        });
+
+        try {
+            Response response = clients.block(clients.call(
+                    channel, new FeignRequestEndpoint(request), builder.build(), IdentityDeserializer.INSTANCE));
+            return feign.Response.create(
+                    response.code(),
+                    null,
+                    Multimaps.asMap((Multimap<String, String>) response.headers()),
```
iamdanfox: why the casting here? couldn't we just do `response.headers().asMap()` ?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+            return true;
+        }
+
+        @Override
+        public void close() {
+            // nothing to do
+        }
+    }
+
+    private static final class DialogueResponseBody implements feign.Response.Body {
```
iamdanfox, line 190: is it worth having some kind of toString implementation on this guy? not sure where this might show up in logging / exception messages

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+
+        @Override
+        public Optional<String> accepts() {
+            return Optional.empty();
+        }
+    }
+
+    /** Implements exception handling equivalent dialogue decoders. */
+    static final class RemoteExceptionDecoder implements feign.codec.ErrorDecoder {
+        static final feign.codec.ErrorDecoder INSTANCE = new RemoteExceptionDecoder();
```
iamdanfox: any reason not to use the enum pattern here?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+        private boolean isError(feign.Response response) {
+            return 300 <= response.status() && response.status() <= 599;
+        }
+
+        private Exception parse(feign.Response response) {
+            String body;
+            try {
+                body = toString(response.body().asInputStream());
+            } catch (NullPointerException | IOException e) {
+                UnknownRemoteException exception = new UnknownRemoteException(response.status(), "<unparseable>");
```
iamdanfox: I _think_ if we wanted to exactly match the old behaviour, we'd pass in null for the second arg if the response.body() is null rather than the magic string `"<unparseable>"`?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+            endpoint = getFirstHeader(request, PATH_TEMPLATE).orElse("feign");
+        }
+
+        @Override
+        public void renderPath(Map<String, String> _params, UrlBuilder url) {
+            String target = request.url();
+            Preconditions.checkState(
+                    target.startsWith(baseUrl),
+                    "Request URL must start with base url",
+                    UnsafeArg.of("baseUrl", baseUrl));
```
iamdanfox, line 328: might need to include the `target` here as we presumably always know what the desired baseUrl is?

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -0,0 +1,346 @@
...
+                String querySegments = trailing.substring(queryParamsStart + 1);
+                querySplitter
+                        .split(querySegments)
+                        .forEach((name, value) -> url.queryParam(urlDecode(name), urlDecode(value)));
+            }
+        }
+
+        @Override
+        public HttpMethod httpMethod() {
+            return HttpMethod.valueOf(request.method().toUpperCase(Locale.ENGLISH));
```
iamdanfox: could store this in a field to avoid recomputing the same thing every time we call this method?

iamdanfox (commented):
This is pretty awesome! Hoping this will make the migration waaay easier

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: The okhttp UnknownRemoteExceptionResponseHandler uses `"<empty>"` and `"Failed to read response body"`. I can match those.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: Good call, originally I had implemented this as a channel that transformed responses into exceptions.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: Always helpful to have toString :-)

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: Good catch!

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/JaxRsClient.java
carterkozak: Might as well

conjure-java-server-verifier/src/test/java/com/palantir/verification/server/VerificationClients.java
carterkozak: :+1:

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: ListMultimap.asMap returns `Map<K, List<V>>`, by casting to Multimap we get the requested `Map<K, Collection<V>>`

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: Sort of -- if we use a constant value it inverts the coupling in an odd way, the special string is odd, and the way this is written it's isolated to a single method.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: good call, while I'm here I think I'll rename this class and use it to transform from dialogue response to feign response instead of doing it inline.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/JaxRsClient.java
carterkozak: `JaxRsClient.class` is relatively specific, I've added a test to show that this case fails with `IllegalStateException: Method <method name> not annotated with HTTP method type (ex. GET, POST)`

iamdanfox (approved):
LGTM, I think this will be a good stepping stone to pure dialogue!

PR #1543, opened 2020-03-21
ClientConfiguration no longer enforces failedUrlCooldown

iamdanfox:
## Before this PR

Dialogue can handle NodeSelectionStrategy.ROUND_ROBIN, but as it no longer has a blacklisting channel, it can't do anything with failedUrlCooldown.  I'd like to encourage people to leave this set to zero (and give people a warning/fail fast if they're setting an option we know will do nothing). Unfortunately validation currently prevents us from even constructing a ClientConfiguration object with round robin & failedUrlCooldown of zero.

## After this PR
==COMMIT_MSG==
The requirement for a positive failedUrlCooldown when round_robin is chosen is now enforced in OkhttpClients.create, not when constructing a ClientConfiguration object.
==COMMIT_MSG==

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -128,11 +128,6 @@ public interface ClientConfiguration {
             checkArgument(maxNumRetries() == 0, "If meshProxy is configured then maxNumRetries must be 0");
             checkArgument(uris().size() == 1, "If meshProxy is configured then uris must contain exactly 1 URI");
         }
-        if (nodeSelectionStrategy().equals(NodeSelectionStrategy.ROUND_ROBIN)) {
-            checkArgument(
-                    !failedUrlCooldown().isNegative() && !failedUrlCooldown().isZero(),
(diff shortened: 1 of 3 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -128,11 +128,6 @@ default void check() {
             checkArgument(maxNumRetries() == 0, "If meshProxy is configured then maxNumRetries must be 0");
             checkArgument(uris().size() == 1, "If meshProxy is configured then uris must contain exactly 1 URI");
         }
-        if (nodeSelectionStrategy().equals(NodeSelectionStrategy.ROUND_ROBIN)) {
-            checkArgument(
-                    !failedUrlCooldown().isNegative() && !failedUrlCooldown().isZero(),
```
carterkozak, line 133: Thoughts on keeping the non-negative enforcement?

iamdanfox: Done

carterkozak (approved):

PR #1545, opened 2020-03-23
JaxRsClient dialogue implementation preserves endpoint names

carterkozak:
==COMMIT_MSG==
JaxRsClient dialogue implementation preserves endpoint names
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EndpointNameHeaderEnrichmentContract.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EndpointNameHeaderEnrichmentContract.java
@@ -1,0 +1,40 @@
+/*
+ * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EndpointNameHeaderEnrichmentContract.java
```diff
@@ -0,0 +1,40 @@
...
+import feign.Contract;
+import feign.MethodMetadata;
+import java.lang.reflect.Method;
+
+/**
+ * Contract to capture the endpoint name (method name) and pass it to a feign client.
+ *
+ * This should be considered internal API and should not be depended upon.
+ */
+public final class EndpointNameHeaderEnrichmentContract extends AbstractDelegatingContract {
```
iamdanfox, line 28: out of curiosity, why does this have to be public? could we just pop it in the same package as the JaxRsClient class?

carterkozak: AbstractDelegatingContract Is package private, we could duplicate it or make it public instead.

iamdanfox (approved):

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/EndpointNameHeaderEnrichmentContract.java
iamdanfox: Eh, no strong preference tbh

PR #1548, opened 2020-03-25
dialogue feign client assumes json when no consumes annotation is present

carterkozak:
==COMMIT_MSG==
dialogue feign client assumes json when no consumes annotation is present
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
@@ -149,9 +149,12 @@ final class DialogueFeignClient implements feign.Client {
             if (requestBodyContent.length == 0) {
                 return Optional.empty();
             }
-            throw new SafeIllegalStateException("A Content-Type header was not present on feign request with a body");
         }
-        return Optional.of(new ByteArrayRequestBody(requestBodyContent, maybeContentType.get()));
+        return Optional.of(new ByteArrayRequestBody(
+                requestBodyContent,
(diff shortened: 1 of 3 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -149,9 +149,12 @@ private static String urlDecode(String input) {
...
                 return Optional.empty();
             }
-            throw new SafeIllegalStateException("A Content-Type header was not present on feign request with a body");
         }
-        return Optional.of(new ByteArrayRequestBody(requestBodyContent, maybeContentType.get()));
+        return Optional.of(new ByteArrayRequestBody(
+                requestBodyContent,
+                // A Content-Type header was not present on feign request
+                // with a body, use the default application/json.
+                maybeContentType.orElse("application/json")));
```
iamdanfox, line 157: Out of curiosity, does this mean feign was previously just sending requests without a content-type header?  I think it's fine to say we'll infer application/json and if people don't like that they are free to declare it explicitly!

iamdanfox (approved):

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: Yup, the feign JAXRSContract applies the content-type header only when a `@Consumes` annotation is encountered, the Encoder implementations don't add content-type headers.

PR #1562, opened 2020-03-31
Implement overloads in SslSocketFactories for specific security providers

carterkozak:
## Before this PR
No way to specify a particular provider without jvm-wide configuration.

## After this PR
==COMMIT_MSG==
Implement overloads in SslSocketFactories for specific security providers
==COMMIT_MSG==

## Possible downsides?
Larger API surface.


ferozco:
👍

PR #1566, opened 2020-04-06
Use FROM_ENVIRONMENT proxy config type

LorenzoMartini:
In https://github.com/palantir/conjure-java-runtime-api/pull/535 we add the `FROM_ENVIRONMENT` type for proxy configuration. Here we add the case in creation of proxy selector to pick from JVM's default proxy selector.


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -127,6 +130,14 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
         switch (proxyConfig.type()) {
             case DIRECT:
                 return fixedProxySelectorFor(Proxy.NO_PROXY);
+            case FROM_ENVIRONMENT:
+                String defaultEnvProxy = System.getenv(ENV_HTTPS_PROXY);
+                Preconditions.checkNotNull(
```
gatesn: Put this on one line `String defaultEnvProxy = Preconditions.checkNotNull...`

iamdanfox:
Seems like the example pcloud docs mention some other environment variables which we might want to read too?

```
http_proxy=http://zomp-ovc-gw-1:8888/
https_proxy=http://zomp-ovc-gw-1:8888/
no_proxy=localhost,127.0.0.1,s3.amazonaws.com,s3-external-1.amazonaws.com
```

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
LorenzoMartini: 👍

LorenzoMartini:
@iamdanfox I think for this PR's purpose only https_proxy  is of interest. Otherwise we would need a more clever selection in the api for `FROM_ENVIRONMENT` I guess? Cause we wouldn't know which one to choose otherwise.

iamdanfox:
What about the `no_proxy` environment variable?

gatesn:
I don't think `no_proxy` applies here, because you explicitly set the proxy type on the service configuration instead (therefore choosing which URIs it applies to).

I guess it's a question of what does "FROM_ENVIRONMENT" mean. If we're trying to replicate the behaviour of e.g. curl where no_proxy is respected. Or if we prefer to be more explicit and just say, for this ServiceConfiguration we want to use the proxy declared by env vars.

iamdanfox (approved):
Ok yep makes sense - if you've explicitly opted-in to a proxy for s3 calls, then it seems simplest to just do that.  Given you're just using this for vault calls, should be fine. Thanks for upstreaming this guys!

PR #1568, opened 2020-04-06
Implement a host metrics registry callback

carterkozak:
This simplifies the complexity for callers which can be composed
around individual targets.

## After this PR
==COMMIT_MSG==
Implement a host metrics registry callback
==COMMIT_MSG==

## Possible downsides?
Larger API surface

### DIFF
--- a/changelog/@unreleased/pr-1568.v2.yml
+++ b/changelog/@unreleased/pr-1568.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Implement a host metrics registry callback
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/1568
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
@@ -26,4 +26,25 @@ public interface HostEventsSink {
     void record(String serviceName, String hostname, int port, int statusCode, long micros);
 
     void recordIoException(String serviceName, String hostname, int port);
+
+    default HostEventCallback callback(String serviceName, String hostname, int port) {
+        return new HostEventCallback() {
+            @Override
+            public void record(int statusCode, long micros) {
+                HostEventsSink.this.record(serviceName, hostname, port, statusCode, micros);
+            }
+
+            @Override
+            public void recordIoException() {
+                HostEventsSink.this.recordIoException(serviceName, hostname, port);
+            }
+        };
+    }
+
+    interface HostEventCallback {
+
+        void record(int statusCode, long micros);
+
+        void recordIoException();
+    }
 }
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.java.okhttp;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.google.common.base.Suppliers;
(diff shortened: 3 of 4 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
```diff
@@ -26,4 +26,25 @@
     void record(String serviceName, String hostname, int port, int statusCode, long micros);
 
     void recordIoException(String serviceName, String hostname, int port);
+
+    default HostEventCallback callback(String serviceName, String hostname, int port) {
```
iamdanfox, line 30: javadoc saying to prefer this version for performance reasons?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
```diff
@@ -89,4 +96,49 @@ public void recordIoException(String serviceName, String hostname, int port) {
...
+
+        private final ServiceHostAndPort target;
+        private final Supplier<DefaultHostMetrics> hostMetricsSupplier;
+
+        HostEventRegistryCallback(String serviceName, String hostname, int port) {
+            this.target = ImmutableServiceHostAndPort.of(serviceName, hostname, port);
+            // recomputed each hour, must be shorter than the 1 day timeout used by the cache.
+            // This avoids an unnecessary cache load to record metrics.
+            this.hostMetricsSupplier =
+                    Suppliers.memoizeWithExpiration(() -> hostMetrics.get(target), 1, TimeUnit.HOURS);
```
iamdanfox, line 110: two layers of expiry time seems a little much for me :/ we're already micro-optimising pretty hard to avoid allocation and GC here - would it really be so bad to just take the cache read?

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
carterkozak: It depends on the caller, the way okhttp is built it wouldn't make sense to construct a single-use callback, but if you have a channel representation that's reused, it's much better to use the callback approach.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
carterkozak: Since we're inside a single class here I think it should be fine. We could reduce the timeout to 1 minute and still reduce map lookups by several orders of magnitude.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
iamdanfox: Ok sounds like that context you just shared would be helpful info for callers of this interface!

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
iamdanfox: Argh ok

changelog/@unreleased/pr-1568.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Implement a host metrics registry callback
```
iamdanfox, line 3: as a performance optimization to minimize allocation (and thereby GC) on the hot codepath of every server response

iamdanfox (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostMetricsRegistry.java
carterkozak: Do you want this to merge as is, or with a reduced memoize duration?

PR #1572, opened 2020-04-09
Use dialogue exception mapping for the dialogue feign shim

carterkozak:
Less logic, moar shim.

==COMMIT_MSG==
Use dialogue exception mapping for the dialogue feign shim
==COMMIT_MSG==

I was thinking about this when we fixed up the QosException handling in dialogue, this way we don't have to maintain logic in the CJR shim as improvements make their way into Dialogue. Funnily enough the dialogue based feign client already did handle qos exceptions properly.


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -256,63 +255,77 @@ public String toString() {
...
+
+        RemoteExceptionDecoder(ConjureRuntime runtime) {
+            this.runtime = runtime;
+        }
+
+        @Override
+        public Exception decode(String _methodKey, feign.Response response) {
+            try {
+                // The dialogue empty body deserializer properly handles exception mapping
+                runtime.bodySerDe().emptyBodyDeserializer().deserialize(new FeignDialogueResponse(response));
```
iamdanfox, line 324: haha amusing way to get hold of the ErrorDecoder!

iamdanfox (approved):

PR #1573, opened 2020-04-13
Add user agent to ClientConfigurations functions

leonz:
## Before this PR
ClientConfigurations doesn't provide a straightforward utility to provide user agent.  Because Dialogue requires that the userAgent field is set, you end up having to do something a bit clunky:

```
        return ClientConfiguration.builder()
                .from(ClientConfigurations.of(uris, sslSocket, trustManager))
                .userAgent(userAgent)
                .build();
```

## After this PR
==COMMIT_MSG==
Adds a utility function to create a `ClientConfiguration` with a user agent.
==COMMIT_MSG==


ferozco (approved):
Thanks @leonz! Think this a nice quality of life improvement

PR #1578, opened 2020-04-16
Change proxySelector FROM_ENVIRONMENT creation to parse env var as URI

LorenzoMartini:
## Before this PR
The way system.getEnv returns the host and port doesn't parse correctly with `HostAndPort.fromString`, but it does with `URI.create`.

## After this PR
==COMMIT_MSG==
Change proxySelector FROM_ENVIRONMENT creation to parse env var as URI.
==COMMIT_MSG==

Add separate method for testing not to mess with env vars

### DIFF
--- a/changelog/@unreleased/pr-1578.v2.yml
+++ b/changelog/@unreleased/pr-1578.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Change proxySelector FROM_ENVIRONMENT creation to parse env var as
+    URI.
+  links:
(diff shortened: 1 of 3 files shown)


CRogers:
As lorenzo direct slacked me, here is some more info:

> the way we parsed the environment variable is not compatible with the way those variables are in our environment (https://github.palantir.build/pcloud/pcloud-docs/blob/master/docs/usage/egress.rst#using-the-http-egress-proxy). With this change we parse them correctly

CRogers (approved):
Looks to be in line with internal specification document.

PR #1580, opened 2020-04-23
Fix Dialogue feign client empty path paramter handling

carterkozak:
==COMMIT_MSG==
Fix Dialogue feign client empty path parameter handling, no longer excluding empty paths
==COMMIT_MSG==

There's a second issue here: https://github.com/palantir/dialogue/pull/668 for non-final-segment empty path parameters.


carterkozak:
Tests are expected to fail until https://github.com/palantir/dialogue/pull/668 is available

iamdanfox:
Thing it needs a `./gradlew test -Drecreate=true` to fix the traces_on_retries.log

carterkozak:
that's odd, I don't think that test should be impacted!

iamdanfox:
Dw I posted the comment on the wrong PR - should have been on the dialogue one!

carterkozak:
@iamdanfox Mind taking a look at this when you have a moment? I'd like to get it into a quick WC release.

conjure-java-jaxrs-client/build.gradle
```diff
@@ -29,7 +29,12 @@ dependencies {
     testImplementation project(":conjure-java-retrofit2-client")
     testImplementation project(":conjure-java-jersey-server")
     testImplementation project(':keystores')
-    testImplementation "com.palantir.dialogue:dialogue-serde"
+    testImplementation "com.palantir.dialogue:dialogue-serde", {
+        exclude group: 'com.palantir.conjure.java.runtime'
```
ferozco: why do we need this?

carterkozak: Circular dependency on this project which breaks GCV

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientDialogueEndpointTest.java
```diff
@@ -177,6 +178,79 @@ public void testUnsupportedHttpMethod_arbitrary() {
...
+        ArgumentCaptor<Endpoint> endpointCaptor = ArgumentCaptor.forClass(Endpoint.class);
+        ArgumentCaptor<Request> requestCaptor = ArgumentCaptor.forClass(Request.class);
+        verify(channel).execute(endpointCaptor.capture(), requestCaptor.capture());
+        UrlBuilder urlBuilder = mock(UrlBuilder.class);
+        endpointCaptor.getValue().renderPath(ImmutableMap.of(), urlBuilder);
+        verify(urlBuilder).pathSegment("foo"); // context path
+        verify(urlBuilder).pathSegment("begin");
+        verify(urlBuilder).pathSegment("");
+        verify(urlBuilder).pathSegment("end");
+    }
```
iamdanfox, line 211: There's so much set up code in these tests, I find them a little hard to read.  Could we extract the mock channel setup into a function?

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientPathParamHandlingTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void wildcardPathParameterAllowsEmptyString() throws Exception {
+        client.complexPath("");
+        RecordedRequest takeRequest = server.takeRequest();
+        assertThat(takeRequest.getRequestLine()).isEqualTo("GET /complex/ HTTP/1.1");
+    }
+
+    @Test
+    public void innerPathParameterAllowsEmptyString() throws Exception {
+        client.innerPath("");
```
iamdanfox, line 110: if we just stuck in a "/" would we expect it to just get urlencoded in both these cases? or would one of them possibly do `complex//` ?

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientDialogueEndpointTest.java
```diff
@@ -177,6 +178,79 @@ public void testUnsupportedHttpMethod_arbitrary() {
                 .hasMessageContaining("ARBITRARY");
     }
 
+    @Test
+    public void testTrailingWildcardParameter_slashes() {
+        ConjureRuntime runtime = DefaultConjureRuntime.builder().build();
+        Channel channel = mock(Channel.class);
```
ferozco: its probably cleaner to create the mocks at the class level instead of inside almost every test

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientPathParamHandlingTest.java
carterkozak: the `/` character would be encoded to `%2F`. I can add a case for it if you like.

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientDialogueEndpointTest.java
carterkozak: I find it difficult to use tests that configure mocks at the class level :-/
They make it more difficult to add tests for the same functionality with slightly different structure, and make it easy to accidentally invalidate existing tests while changing the common setup.

conjure-java-jaxrs-client/build.gradle
ferozco: You should be able to do:
```
configurations.all {
  resolutionStrategy {
    preferProjectModules()
  }
}
```

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientDialogueEndpointTest.java
ferozco: Sorry, I was just suggesting creating the mocks i.e. `@Mock Channel channel` not configuring the responses. As @iamdanfox suggests we should factor out setup logic

conjure-java-jaxrs-client/build.gradle
carterkozak: Any reason we don't include that in GCV by default?

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientDialogueEndpointTest.java
carterkozak: updated

iamdanfox (approved):

versions.lock
```diff
@@ -22,7 +22,7 @@ com.google.guava:guava:28.1-jre (15 constraints: 8ef6e5ed)
 com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava (1 constraints: bd17c918)
 com.google.j2objc:j2objc-annotations:1.3 (1 constraints: b809eda0)
 com.jcraft:jzlib:1.1.3 (1 constraints: 0705f635)
-com.netflix.concurrency-limits:concurrency-limits-core:0.2.2 (2 constraints: 3b148789)
+com.netflix.concurrency-limits:concurrency-limits-core:0.2.2 (1 constraints: 0605f335)
```
uschi2000, line 25: @carterkozak didn't your re-implementation kill this dependency?

carterkozak: Only for dialogue, I didn’t replace it from cjr-okttp.

PR #1587, opened 2020-05-05
Interfaces for client factories

iamdanfox:
## Before this PR

See https://internal-github/foundry/witchcraft/pull/2427

## After this PR
==COMMIT_MSG==
The client-config jar now publishes some standard interfaces for implementing client factories. This should allow libraries to rely on these interfaces (and thereby take wc-provided clients) rather than embedding their own client creation.
==COMMIT_MSG==

Usage is intended to be _fluent_, e.g.

```java
AuthService auth = wc.clients()
        .withNodeSelectionStrategy(NodeSelectionStrategy.ROUND_ROBIN)
        .get(AuthService.class, "multipass");
```

This PR does not contain any _implementation_ of a client factory, as these are quite fiddly and often include caching, and may want to switch between dialogue/retrofit and jaxrs.

here's an example implementation on the dialogue side: https://github.com/palantir/dialogue/pull/694


client-config/build.gradle
```diff
@@ -5,6 +5,8 @@ apply plugin: 'com.palantir.revapi'
 dependencies {
     api 'com.palantir.conjure.java.api:service-config'
     api 'com.palantir.tritium:tritium-registry'
+    api 'com.google.errorprone:error_prone_annotations'
+    api 'com.palantir.refreshable:refreshable'
```
iamdanfox, line 9: I could move this stuff to a separate jar if we want? like `com.palantir.conjure.java.runtime:client-factories`?

carterkozak: client-factory-api maybe?

carterkozak: I don't think we really buy anything from splitting them though, we're forced into the api dependency on `ClientCofiguration.<enum>`. If we are interested in breaking that, we should put the client factory api in cjr-api instead.

carterkozak: What if we didn't depend on the ClientConfiguration enums and allowed those via mixins elsewhere?

iamdanfox: I mean it's feasible to put n-1 interfaces in c-j-r-api but leave the enum-related mixin here, just seems like the sprawl might end up a bit of a pain.

carterkozak (approved):

PR #1588, opened 2020-05-05
Fix: Wrap certificate input stream in BufferedInputStream to avoid single byte reads from X509Factory 

robert3005:
## Before this PR
We defer certificate reading to certificate factory which does single byte reads

## After this PR
==COMMIT_MSG==
Copy certificate files to memory before delegating to avoid inefficient single byte disk reads from X509Factory
==COMMIT_MSG==

The problematic method is https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/provider/X509Factory.java#L547

## Possible downsides?
We consume few hundred kilobytes extra memory when reading certificates

### DIFF
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/DefaultCas.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/DefaultCas.java
@@ -19,13 +19,11 @@ package com.palantir.conjure.java.config.ssl;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.hash.Hashing;
-import com.google.common.io.ByteStreams;
+import com.google.common.io.Resources;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


carterkozak:
Would a BufferedInputStream be a reasonable alternative?

robert3005:
Sure, I wrote both versions and thought the amount of data is small that reading all of it shouldn't matter but also if it's small bufferedinputstream should do the trick

keystores/src/main/java/com/palantir/conjure/java/config/ssl/DefaultCas.java
```diff
@@ -51,8 +56,12 @@
     private static Map<String, X509Certificate> getTrustedCertificates() {
         ImmutableMap.Builder<String, X509Certificate> certificateMap = ImmutableMap.builder();
         try {
+            InputStream certificateResource = DefaultCas.class.getResourceAsStream(CA_CERTIFICATES_CRT);
```
carterkozak: Resource streams must be closed, this can be added to the try

keystores/src/main/java/com/palantir/conjure/java/config/ssl/DefaultCas.java
```diff
@@ -51,8 +56,12 @@
...
+            ByteArrayOutputStream inMemoryCertificates = new ByteArrayOutputStream();
+            ByteStreams.copy(certificateResource, inMemoryCertificates);
```
carterkozak: ByteStreams.toByteArray avoids the ByteArrayOutputStream, alternatively guava Resources can give us `Resources.toByteArray(Resources.getResource(CA_CERTIFICATES_CRT))`.

I like the BufferedInputStream approach better because it avoids stream->bytea->bais, we simply wrap the original stream and don't need to worry about it.

carterkozak (approved):

PR #1591, opened 2020-05-06
ClientConfiguration includes Optional<HostEventsSink>

iamdanfox:
## Before this PR

I'm trying to pick up the new fluent client factory interface in WC, but it turns out while 99% of dialogue client logic was in open-source land, the HostEventsSink wiring was kept internal to WC.

I don't really want to regress on this.

## After this PR
==COMMIT_MSG==
ClientConfiguration includes `Optional<HostEventsSink>`
==COMMIT_MSG==

## Possible downsides?
- All the JaxRsClient methods are confusing af because they now have two ways to specify a HostEventsSink. I am so done with this repo.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
@@ -23,7 +23,6 @@ import com.palantir.conjure.java.api.config.service.BasicCredentials;
 import com.palantir.conjure.java.api.config.service.PartialServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.UserAgent;
-import com.palantir.conjure.java.okhttp.HostEventsSink;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.UnsafeArg;
 import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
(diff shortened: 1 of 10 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -23,6 +23,7 @@
 import com.palantir.conjure.java.api.config.service.PartialServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.UserAgent;
+import com.palantir.conjure.java.okhttp.HostEventsSink;
```
carterkozak: okhttp :-/

iamdanfox: Yep. This is gross. Unfortunately I think changing this package name is gonna be a whole _world_ of pain.

carterkozak: How terrible would it be to redefine in a new package and update the old interfaces to extend the new?

iamdanfox: Hmm ok I think we might actually be able to do that - let me give it a shot

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/HostEventsSink.java
```diff
@@ -41,10 +46,55 @@ public void recordIoException() {
...
+
+                @Override
+                public void recordIoException() {
+                    other.recordIoException();
+                }
+            };
+        }
+    }
+
+    static HostEventsSink from(com.palantir.conjure.java.client.config.HostEventsSink other) {
```
iamdanfox, line 80: I had to add all these gross converter things because OkHttpClients.java has methods which revapi otherwise says would ABI break.

```
public static OkHttpClient create(
            OkHttpClient.Builder client,
            ClientConfiguration config,
            UserAgent userAgent,
            HostEventsSink hostEventsSink,
            Class<?> serviceClass) {
```

carterkozak: I like it, thanks!

carterkozak (approved):
lgtm pending CI style fixes

PR #1596, opened 2020-05-13
All jaxrs clients use dialogue transport

carterkozak:
In a future change we may be able to remove the dependency from
jaxrs-clients on okhttp, but that will take more work and potentially
break upgrades when folks depend on the transitive detail.

## Before this PR
jaxrs clients used okhttp unless otherwise requested using a separate API.

## After this PR
==COMMIT_MSG==
All jaxrs clients use dialogue transport
==COMMIT_MSG==

## Possible downsides?
Change always brings risk.
Larger dependency footprint from the `conjure-java-jaxrs-client` module.


iamdanfox:
Looking at the atlasdb analysis I made yesterday, it seems like the majority of foundry services have picked up [atlasdb 0.226.0](https://github.com/palantir/atlasdb/releases/tag/0.226.0) on develop (which is the release that uses dialogue by default)

image

Given that this transition has gone smoothly so far, I think it's time to merge and release this, allowing us to chop off the long-tail of feign+jaxrs+okhttp usage.

iamdanfox (approved):
LGTM. Think this'll need another approval because I pushed a commit

carterkozak:
:+1:

PR #1666, opened 2020-06-26
Support TLSv1.3 ciphers

carterkozak:
Remove the legacy cipher marker for pre-tls1.2 as nothing
allows older protocols anymore.

==COMMIT_MSG==
Support TLSv1.3 ciphers
==COMMIT_MSG==


client-config/src/main/java/com/palantir/conjure/java/client/config/CipherSuites.java
```diff
@@ -46,8 +48,8 @@
             "TLS_RSA_WITH_AES_128_GCM_SHA256");
 
     private static final ImmutableList<String> ALL_CIPHER_SUITES = ImmutableList.<String>builder()
-            .addAll(FAST_CIPHER_SUITES)
             .addAll(GCM_CIPHER_SUITES)
+            .addAll(FAST_CIPHER_SUITES)
```
carterkozak, line 52: @robert3005 this is for you

ferozco (approved):

PR #1678, opened 2020-07-01
Prune dialogue hc4 dependencies

carterkozak:
This also deduplicates testing now that all jaxrs clients use
the dialogue transport.

==COMMIT_MSG==
Prune dialogue hc4 dependencies
==COMMIT_MSG==


iamdanfox (approved):

carterkozak:
no changelog because this only impacts tests, jaxrsclient.create already used the httpclient 5 implementation.

PR #1765, opened 2020-09-04
Update conjure-scala-jaxrs-client to use Scala 2.12

dsw2127:
## Before this PR

conjure-scala-jaxrs-client uses scala 2.11

## After this PR

==COMMIT_MSG==
Update conjure-scala-jaxrs-client to use Scala 2.12
==COMMIT_MSG==


ferozco:
This will break things at runtime for people who are not using scala `2.12` right? Can you confirm that internally people have upgraded?

dsw2127:
The projects internally that use conjure-scala-jaxrs-client are either fusion-related - which is motivating this change as I am working to upgrade them to Scala 2.12 - or witchcraft, which has it and scala generally only as a test dependency. I am handling the fusion repos, and I can take care of the bump in witchcraft after we have a version of conjure-scala-jaxrs-client for Scala 2.12

iamdanfox (approved):
Confirmed there are literally only two usages in all of internal github, let's do it.

https://sourcegraph.p.b/search?q=conjure-scala-jaxrs-client+file:versions.lock&patternType=literal

PR #1772, opened 2020-09-14
ConjureJerseyFeature.INSTANCE allows plugging in internal error observability

iamdanfox:
## Before this PR

There's work in progress internally to drive down the rates of _internal errors_ across servers (i.e. things that are probably a mistake on the the server-author's part, not just weird user input / strange network conditions). This has been implemented for WC's conjure-undertow endpoints, but as @samrogerson pointed out, there are many servers which have extensive API surfaces that aren't currently conjurized (and aren't feasibly able to become conjurized).

Currently, WC's `EndpointExceptionMetricsHandler` ends up using some smart heuristics to decide whether something is an internal error or not, then adds some params to the request logs: 
- `_internalErrorIdentifier` (a short hash of the actual stacktrace)
- `_internalErrorType` `(qos|internal|service_internal|rpc|other)`.

It also uses our MDC magic string to make sure that these params are also appear in the associated service.log `log.info` / `log.error` lines.

c-j-r already had some logic to instrument internal errors added by @parker89 in https://github.com/palantir/conjure-java-runtime/pull/1336 and extended with a tag in https://github.com/palantir/conjure-java-runtime/pull/1420. _Clearly_ we do not want to maintain two implementations of WC's fancy logic, so I think we're going to have to suck it up and make this stuff pluggable.

## After this PR
==COMMIT_MSG==
Rather than using `ConjureJerseyFeature.INSTANCE`, it is now recommended to use `ConjureJerseyFeature.builder().exceptionListener(...).build()` to allow WC's smarter 'internal error' detection logic to hook up the necessary request and service log params.
==COMMIT_MSG==

## Possible downsides?
- ~I think this approach doesn't actually clear MDCs yet~


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -17,23 +17,21 @@
...
     INSTANCE;
 
-    private final JerseyServerMetrics metrics;
+    private final Consumer<Throwable> listener;
 
     ConjureJerseyFeature() {
-        this.metrics = JerseyServerMetrics.of(SharedTaggedMetricRegistries.getSingleton());
-    }
-
-    public JerseyServerMetrics getMetrics() {
```
iamdanfox, line 35: I've deleted this method because I can't find any usages of it internally:

https://sourcegraph.p.b/search?q=ConjureJerseyFeature.INSTANCE.getMetrics%28%29+lang:java&patternType=literal

carterkozak:
This appears to duplicate a nontrivial amount of code with our internal server project. What if we add a provider interface for exception listeners which allows us to register implementations with jersey from the internal project and reuse code?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/InternalErrorExceptionListener.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public void accept(Throwable exception) {
+        Optional<ErrorCause> errorCause = classifyTypeOfInternalError(exception);
+        if (errorCause.isPresent()) {
+            ErrorCause cause = errorCause.get();
+            metrics.internalerrorAll(cause.toString()).mark();
+        }
+    }
```
iamdanfox: @carterkozak this class is just preserving the behaviour that was already working in c-j-r, but I consider this literally as a stopgap thing because I want to _immediately_ swap it out for WC's implementation.

carterkozak: Ack, can we take advantage of Jerseys dependency injection framework here?

iamdanfox: Chatted over VTC and figured that there aren't actually _that_ many prod usages of ConjureJerseyFeature.INSTANCE (since 99% of servers get it for free in WC). I think this means it's reasonable for any stragglers to just use the builder.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -17,42 +17,43 @@
...
-        this.metrics = JerseyServerMetrics.of(SharedTaggedMetricRegistries.getSingleton());
-    }
 
-    public JerseyServerMetrics getMetrics() {
-        return metrics;
-    }
+    /**
+     * Prefer {@link #builder()} which allows passing in a real {@link ExceptionListener} which records internal errors.
+     */
+    INSTANCE;
```
iamdanfox, line 32: Thoughts on explicitly deprecating this? I didn't do it originally because I didn't want to flag the test-only usages, but looking at sourcegraph there are actually a few prod users... https://sourcegraph.p.b/search?q=ConjureJerseyFeature.INSTANCE+lang:java&patternType=literal

carterkozak: I'm okay with a deprecation if you'd like to PR uses out of "large internal" :-)

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -86,4 +87,81 @@ public boolean configure(FeatureContext context) {
...
+
+                @Override
+                public String toString() {
+                    return "ConjureJerseyFeature{}";
+                }
+            };
+        }
+    }
+
+    /** Implementors of this interface can. */
```
carterkozak: What can implementors do?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -86,4 +87,81 @@ public boolean configure(FeatureContext context) {
...
+
+        default <T extends Throwable> ExceptionMapper<T> augment(ExceptionMapper<T> delegate) {
+            return new ExceptionMapper<T>() {
+                @Override
+                public Response toResponse(T exception) {
+                    try {
+                        onException(exception);
+                        return delegate.toResponse(exception);
+                    } finally {
+                        afterResponseBuilt();
```
carterkozak: Not sure the abstract methods really make sense here, I think this interface only needs to have a single method which consumes a Throwable.

iamdanfox: That's what I had originally, are you thinking we just do MDC.clear() afterwards?

carterkozak: Sorry, wasn't quite a complete thought.

Since this interface deals directly with ExceptionMapper, I was thinking it might as well operate as an ExceptionMapper wrapper so implementations can use try/finally. Not a strong preference though, and it's probably fine as is.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -86,4 +87,84 @@ public boolean configure(FeatureContext context) {
...
         return true;
     }
+
+    @CheckReturnValue
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @CheckReturnValue
+    public static final class Builder {
```
carterkozak, line 95: private ctor. I'll write an error prone rule for this today.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
iamdanfox: Ugh, seems like this 100% does not work with jersey (I think due to type erasure). 

```
Caused by: javax.ws.rs.ProcessingException: Could not find exception type for given ExceptionMapper class: class com.palantir.conjure.java.server.jersey.ConjureJerseyFeature$ExceptionListener$1.
	at org.glassfish.jersey.internal.ExceptionMapperFactory.getType(ExceptionMapperFactory.java:260)
	at org.glassfish.jersey.internal.ExceptionMapperFactory.getExceptionType(ExceptionMapperFactory.java:230)
	at org.glassfish.jersey.internal.ExceptionMapperFactory.lambda$createLazyExceptionMappers$1(ExceptionMapperFactory.java:204)
	at org.glassfish.jersey.internal.util.collection.Values$LazyValueImpl.get(Values.java:317)
	at org.glassfish.jersey.internal.ExceptionMapperFactory.find(ExceptionMapperFactory.java:115)
	at org.glassfish.jersey.internal.ExceptionMapperFactory.findMapping(ExceptionMapperFactory.java:102)
	at org.glassfish.jersey.server.ServerRuntime$Responder.mapException(ServerRuntime.java:523)
	at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:403)
	at org.glassfish.jersey.server.ServerRuntime$1.run(ServerRuntime.java:261)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
```

Gonna plumb these in manually and see if that does the trick.

carterkozak: Ya, that makes sense. I think we need to provide the listener to each exception mapper.

carterkozak: Can reproduce the issue by removing the `ExceptionListener.augment` override from `NoOpListener`.

carterkozak:
:+1:

PR #1802, opened 2020-10-23
Make dialogue-target an api dependency

jawshooah:
`ConjureRuntime` and `Channel` classes are exposed in `JaxRsClient` public API. Unless a consumer pulls in dialogue-target from some other dependency, this results in a `NoClassDefFoundError` at runtime.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure-java-runtime/1802)

### DIFF
--- a/changelog/@unreleased/pr-1802.v2.yml
+++ b/changelog/@unreleased/pr-1802.v2.yml
@@ -1,0 +1,8 @@
+type: fix
+fix:
+  description: |-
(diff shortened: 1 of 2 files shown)


iamdanfox (approved):

iamdanfox:
Oof - I wish we actually had tooling to detect this... maybe one day java modules will save us.

PR #1804, opened 2020-10-23
Always enable secure GCM ciphers by default

carterkozak:
Previously GCM ciphers were disabled by default on java 8 due
to performance problems, however at this point the vast majority
of services have upgraded and there's greater risk in continuing
to use less secure cipher suites than the performance impact on a
few straggler services.
For now clients may still override `enable-gcm-cipher-suites: false`
to opt out of GCM ciphers, however we have no intention of allowing
servers to use CBC ciphers moving forward.

==COMMIT_MSG==
Always enable secure GCM ciphers by default, previously java 8 clients avoided GCM ciphers by default.
==COMMIT_MSG==

## Possible downsides?
Some java8 services may experience degraded performance.

### DIFF
--- a/changelog/@unreleased/pr-1804.v2.yml
+++ b/changelog/@unreleased/pr-1804.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Always enable secure GCM ciphers by default, previously java 8 clients
+    avoided GCM ciphers by default.
+  links:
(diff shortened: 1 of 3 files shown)


iamdanfox (commented):
Looks good, especially as there's literally just a couple of service still on java8 floating around.

iamdanfox (approved):

PR #1832, opened 2020-12-02
Fix string-to-long coercion when afterburner is enabled

carterkozak:
Reported upstream:
https://github.com/FasterXML/jackson-modules-base/issues/120

## Before this PR
The CoercionConfig is fantastic, unfortunately it is not yet
supported by the AfterBurner module. Separately we may consider
moving to the new Blackbird module on java 11 runtimes.

For now I've updated our existing long deserializers for 2.12.0
which has the benefit of rejecting coercion from string to SafeLong.

## After this PR
==COMMIT_MSG==
Fix string-to-long coercion when afterburner is enabled
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
+++ b/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
@@ -19,12 +19,14 @@ package com.palantir.conjure.java.serialization;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.exc.InputCoercionException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.palantir.logsafe.Preconditions;
 import java.io.File;
 import java.io.IOException;
 import java.math.BigInteger;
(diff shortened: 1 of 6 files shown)


conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -128,11 +131,115 @@ public void testLongTypeDeserializationFromString() throws IOException {
...
+    public void testLongBeanTypeDeserializationFromString() throws IOException {
+        assertThat(MAPPER.readValue("{\"value\":\"1\"}", LongBean.class)).isEqualTo(new LongBean(1L));
+    }
+
+    @Test
+    public void testLongBeanTypeDeserializationFromNumber() throws IOException {
+        assertThat(MAPPER.readValue("{\"value\":\"1\"}", LongBean.class)).isEqualTo(new LongBean(1L));
+    }
+
+    static final class LongBean {
```
ferozco, line 144: can we use something we generate from Conjure here? or an immutable here

carterkozak: Conjure is not impacted by the bug, this requires direct bean field access via afterburner. Immutables generated objects may work, but there's no guarantee around their implementation and it may change between versions.

ferozco: ok, I think that having class definitions (especially verbose ones) scattered throughout the test class makes it harder to read. Could we either push these to the bottom or pull them out into separate files

carterkozak: I suppose, I tend to prefer having classes for tests colocated with the test so it's easy to quickly grok the entire case, but we can put them at the bottom of the class ¯\_(ツ)_/¯

ferozco:
👍

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
ferozco: ok I don't have strong feelings here

PR #1836, opened 2020-12-07
Restore joda deserialization

iamdanfox:
## Before this PR

@rahij reported in #dev-foundry-infra (/archives/C5MKHTMD4/p1607350000042200) that after the jackson 2.12.0 upgrade, deserializing some classes from our `<analysis product>` which previously worked with a vanilla c-j-r object mapper started failing with:

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Joda date/time type `org.joda.time.DateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-joda" to enable handling
```

## After this PR
==COMMIT_MSG==
conjure-java-jackson-serialization: joda types can once again be deserialized (as this was a behaviour change of the jackson 2.12 upgrade) 
==COMMIT_MSG==

https://github.com/FasterXML/jackson-databind/blob/39fab67429c0f23399f031e0c0dbd4a13a73622d/release-notes/VERSION-2.x#L71-L72

https://github.com/FasterXML/jackson-databind/issues/2776

## Possible downsides?
- I'd really prefer people not to use joda


versions.lock
```diff
@@ -65,6 +66,7 @@ jakarta.el:jakarta.el-api:3.0.3 (1 constraints: fe135465)
 jakarta.validation:jakarta.validation-api:2.0.2 (3 constraints: 8e372576)
 jakarta.ws.rs:jakarta.ws.rs-api:2.1.6 (12 constraints: 4cd52d0a)
 jakarta.xml.bind:jakarta.xml.bind-api:2.3.2 (1 constraints: 30198ba6)
+joda-time:joda-time:2.10.8 (2 constraints: b4238380)
```
carterkozak, line 69: not a huge fan of including joda in cjr, but I don't want to regress folks

carterkozak (approved):

versions.lock
iamdanfox: agreed - if we were being more brutal I think we should just wait for the P0s and then ask people to manually add the dep, but I think this would be too mean :/

PR #1839, opened 2020-12-09
Fix: Downgrade Jackson to 2.11.1

ferozco:
## Before this PR
We encountered an issue with de serializing doubles as BigDecimal which caused services to fail to accept valid requests 

## After this PR
==COMMIT_MSG==
Downgrade Jackson to 2.11.1
==COMMIT_MSG==


ferozco:
pending: https://github.com/palantir/dialogue/pull/1077

carterkozak (approved):

PR #1853, opened 2021-01-06
Break: Remove Retrofit Support

carterkozak:
Retrofit does not load on jre14+
Retrofit has known security vulnerabilities
Retrofit cannot be upgraded without breaking conjure

Internal burndown tracking: https://pl.ntr/cjr-internal-burndown
[External usage search](https://github.com/search?q=com.palantir.conjure.java.client.retrofit2+-repo%3Apalantir%2Fconjure-java-runtime++NOT+is%3Afork&type=code) (0 hits)

## Before this PR
retrofit and okhttp on the runtime classpath despite not being used in the majority of cases.

## After this PR
==COMMIT_MSG==
Break: remove retrofit support, remove dependency on okhttp
==COMMIT_MSG==

## Possible downsides?
Big break, but we built dialogue to replace okhttp, retrofit, and feign.
Consumers may need to add a dependency on the conjure okhttp-clients project and migrate retrofit clients to dialogue.


carterkozak:
I'm not sure I like the dependency inversion from `jaxrs depends on okhttp` to `okhttp depends on jaxrs`. Might be better to add a new legacy-clients jar with the classes that are migrated out of okhttp-clients.

carterkozak:
I've split out the okhttp decoupling component here: #1955

bmarcaur (approved):

carterkozak:
👍

PR #1874, opened 2021-01-26
FROM_ENVIRONMENT proxy selector to behave as no proxy if env var not set

rkaram:
## Before this PR
`ClientConfigurations` built with a `FROM_ENVIRONMENT` proxy would throw if the `https_proxy` environment variable was not set.

## After this PR

==COMMIT_MSG==
`ClientConfigurations` built with a `FROM_ENVIRONMENT` proxy will now behave as if there is no proxy if the `https_proxy` environment variable is not set, instead of throwing.
==COMMIT_MSG==

## Possible downsides?
Users who were relying on the exception being thrown as a validation that the proxy is set might now see the client silently be constructed without a proxy.

### DIFF
--- a/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
+++ b/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
@@ -23,8 +23,6 @@ import com.palantir.conjure.java.api.config.service.ProxyConfiguration;
 import com.palantir.conjure.java.api.config.service.ServiceConfiguration;
 import com.palantir.conjure.java.api.config.service.UserAgent;
 import com.palantir.conjure.java.config.ssl.SslSocketFactories;
-import com.palantir.logsafe.Preconditions;
-import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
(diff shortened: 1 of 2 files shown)


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -151,10 +149,10 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
             case DIRECT:
                 return fixedProxySelectorFor(Proxy.NO_PROXY);
             case FROM_ENVIRONMENT:
-                String defaultEnvProxy = Preconditions.checkNotNull(
-                        System.getenv(ENV_HTTPS_PROXY),
-                        "Missing environment variable",
-                        SafeArg.of("name", ENV_HTTPS_PROXY));
+                String defaultEnvProxy = System.getenv(ENV_HTTPS_PROXY);
+                if (defaultEnvProxy == null) {
+                    return fixedProxySelectorFor(Proxy.NO_PROXY);
```
carterkozak, line 160: Let's add an info level log line in either case so we can be confident we understand the environment when things go wrong.

rkaram: Done

carterkozak (approved):

PR #1897, opened 2021-03-01
Use SOURCE retention for immutables style annotations

pkoenig10:
See https://github.com/immutables/immutables/issues/291.

Example of similar fix in https://github.com/palantir/conjure-java-runtime-api/pull/369.

This eliminates compilation warnings for consumers that do not directly depend on immutables.

### DIFF
--- a/conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
+++ b/conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
@@ -320,7 +320,7 @@ public final class Retrofit2ClientApiTest extends TestBase {
                 AGENT,
                 new HostMetricsRegistry(),
                 ClientConfiguration.builder()
-                        .from(createTestConfig("https://invalid.service.dev"))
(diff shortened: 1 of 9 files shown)


pkoenig10:
Not sure why the tests are failing, but looks like they are failing on develop also.

CRogers (approved):
Nice thanks.

conjure-java-retrofit2-client/src/test/java/com/palantir/conjure/java/client/retrofit2/Retrofit2ClientApiTest.java
```diff
@@ -320,7 +320,7 @@ private void connectionFailureWithFuture(Supplier<Future<String>> futureSupplier
                 AGENT,
                 new HostMetricsRegistry(),
                 ClientConfiguration.builder()
-                        .from(createTestConfig("https://invalid.service.dev"))
+                        .from(createTestConfig("https://service.invalid"))
```
pkoenig10, line 323: The `.dev` TLD was causing test failures in CircleCI. Something must have changed and now this domain correctly resolves and whatever server is there returns a 404.

Switched to `.invalid` which is a reserved TLD for invalid domains.
https://en.wikipedia.org/wiki/.invalid

PR #1955, opened 2021-04-26
conjure-java-jaxrs-client no longer provides okhttp-clients

carterkozak:
Our use of okhttp is deprecated, the library will no longer be
included on the classpath by default.

## Before this PR
okhttp dependencies everywhere.

## After this PR
No okhttp dependency from conjure-java-jaxrs-client.
==COMMIT_MSG==
conjure-java-jaxrs-client no longer provides okhttp-clients
==COMMIT_MSG==

## Possible downsides?
Consumers may depend on okhttp-clients via conjure-java-jaxrs-clients. Our modern dependency validation requests that developers list the implementation dependencies they use, so our users shouldn't be broken.


carterkozak:
This is a stepping-stone toward #1853

conjure-java-jaxrs-client/build.gradle
```diff
@@ -4,7 +4,7 @@ apply plugin: 'com.palantir.revapi'
 dependencies {
     api project(":extras:refresh-utils")
     api project(":client-config")
-    api project(":okhttp-clients")
+    api project(":conjure-java-legacy-clients")
```
jkozlowski, line 7: Doesn't API mean consumers can still get it? I'm assuming this is a first PR in a chain

okhttp-clients/build.gradle
```diff
@@ -6,6 +6,7 @@ apply plugin: 'com.palantir.metric-schema'
 
 dependencies {
     api project(':client-config')
+    api project(':conjure-java-legacy-clients')
```
jkozlowski, line 9: legacy vs deprecated?

jkozlowski (approved):

conjure-java-jaxrs-client/build.gradle
carterkozak: Yep, this new module doesn't provide okhttp, rather two public classes that used to be provided by `okhttp-clients`: `HostEventsSink` (the deprecated version, long story) and `QosExceptionResponseMapper` which is shared between jaxrs and retrofit.

PR #2019, opened 2021-06-25
Propagate error code and name for 401 and 403 responses

pkoenig10:
## Context
Our internal authorization service throws specific, useful errors when writes are denied due to lack of permissions or constraint violations.

We often want these errors to be visible to end users so they can understand why a particular request failed. Unfortunately, there are often multiple services between end users and the authorization service. In order to communicate these errors to end users, we would have to define these errors in a very large number of services and add manual logic to propagate them.

The corresponding conjure-java PR for Undertow endpoints is https://github.com/palantir/conjure-java/pull/1399.

## Before this PR
Remote exceptions with 401 and 403 statuses are propagate as the corresponding generic `ErrorType`.

## After this PR
Remote exceptions with 401 and 403 statuses are propagated preserving the error code and error name. Parameters are still not propagated.

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
@@ -19,10 +19,10 @@ package com.palantir.conjure.java.server.jersey;
 import com.palantir.conjure.java.api.errors.ErrorType;
 import com.palantir.conjure.java.api.errors.RemoteException;
 import com.palantir.conjure.java.api.errors.SerializableError;
+import com.palantir.conjure.java.api.errors.ServiceException;
 import com.palantir.logsafe.SafeArg;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 import org.slf4j.Logger;
@@ -54,64 +54,35 @@ final class RemoteExceptionMapper extends ListenableExceptionMapper<RemoteExcept
 
     @Override
     public Response toResponseInner(RemoteException exception) {
-        ErrorType errorType = mapErrorType(exception);
-        Response.ResponseBuilder builder = Response.status(errorType.httpErrorCode());
+        if (exception.getStatus() == 401 || exception.getStatus() == 403) {
+            log.info(
+                    "Encountered a remote exception",
+                    SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
+                    SafeArg.of("errorName", exception.getError().errorName()),
+                    SafeArg.of("status", exception.getStatus()),
+                    exception);
 
-        try {
-            SerializableError error = SerializableError.builder()
-                    .errorName(errorType.name())
-                    .errorCode(errorType.code().toString())
-                    .errorInstanceId(exception.getError().errorInstanceId())
+            return Response.status(exception.getStatus())
+                    .type(MediaType.APPLICATION_JSON)
+                    .entity(SerializableError.builder()
+                            .errorCode(exception.getError().errorCode())
+                            .errorName(exception.getError().errorName())
+                            .errorInstanceId(exception.getError().errorInstanceId())
+                            .build())
                     .build();
-            builder.type(MediaType.APPLICATION_JSON);
-            builder.entity(error);
-        } catch (RuntimeException e) {
-            log.warn(
-                    "Unable to translate exception to json",
-                    SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
-                    SafeArg.of("errorName", exception.getError().errorName()),
-                    e);
-            // simply write out the exception message
-            builder.type(MediaType.TEXT_PLAIN);
-            builder.entity("Unable to translate exception to json. Refer to the server logs with this errorInstanceId: "
-                    + exception.getError().errorInstanceId());
-        }
-        return builder.build();
-    }
-
-    private ErrorType mapErrorType(RemoteException exception) {
-        Status status = Status.fromStatusCode(exception.getStatus());
-
-        if (status.getStatusCode() == 401) {
-            log.info(
-                    "Encountered a remote unauthorized exception."
-                            + " Mapping to a default unauthorized exception before propagating",
-                    SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
-                    SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
-                    exception);
-
-            return ErrorType.UNAUTHORIZED;
-        } else if (status.getStatusCode() == 403) {
-            log.info(
-                    "Encountered a remote permission denied exception."
-                            + " Mapping to a default permission denied exception before propagating",
-                    SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
-                    SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
-                    exception);
-
-            return ErrorType.PERMISSION_DENIED;
         } else {
-            // log at WARN instead of ERROR because this indicates an issue in a remote server
             log.warn(
                     "Encountered a remote exception. Mapping to an internal error before propagating",
                     SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
                     SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
+                    SafeArg.of("status", exception.getStatus()),
                     exception);
 
-            return ErrorType.INTERNAL;
+            ServiceException propagatedException = new ServiceException(ErrorType.INTERNAL, exception);
+            return Response.status(propagatedException.getErrorType().httpErrorCode())
+                    .type(MediaType.APPLICATION_JSON)
+                    .entity(SerializableError.forException(propagatedException))
+                    .build();
         }
     }
 }
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
@@ -59,7 +59,7 @@ final class RemoteExceptionMapper extends ListenableExceptionMapper<RemoteExcept
                     "Encountered a remote exception",
                     SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
                     SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("status", exception.getStatus()),
+                    SafeArg.of("statusCode", exception.getStatus()),
                     exception);
 
             return Response.status(exception.getStatus())
@@ -75,7 +75,7 @@ final class RemoteExceptionMapper extends ListenableExceptionMapper<RemoteExcept
                     "Encountered a remote exception. Mapping to an internal error before propagating",
                     SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
                     SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("status", exception.getStatus()),
+                    SafeArg.of("statusCode", exception.getStatus()),
                     exception);
 
             ServiceException propagatedException = new ServiceException(ErrorType.INTERNAL, exception);
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
@@ -40,8 +40,8 @@ import org.slf4j.LoggerFactory;
  * response is then thrown as a {@link RemoteException} at the call point in server A. If server A does not catch this
  * exception and it raises up the call stack back into Jersey, execution enters this {@link RemoteExceptionMapper}.
  *
- * <p>To preserve debuggability, the exception and HTTP status code from B's exception are logged at WARN level, but not
- * propagated to caller to avoid an unintentional dependency on the remote exception.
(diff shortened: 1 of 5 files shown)


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
-                    "Encountered a remote unauthorized exception."
-                            + " Mapping to a default unauthorized exception before propagating",
-                    SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
-                    SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
-                    exception);
-
-            return ErrorType.UNAUTHORIZED;
-        } else if (status.getStatusCode() == 403) {
+        if (exception.getStatus() == 401 || exception.getStatus() == 403) {
```
pkoenig10, line 57: We could limit this a bit further here and only propagate remote exceptions whose error code matches the corresponding generic `ErrorType`. But I don't think this would make a significant difference in practice. If we receive a 401 or 403 error it's almost certainly going to be with those error codes.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
-                    e);
-            // simply write out the exception message
-            builder.type(MediaType.TEXT_PLAIN);
-            builder.entity("Unable to translate exception to json. Refer to the server logs with this errorInstanceId: "
-                    + exception.getError().errorInstanceId());
-        }
-        return builder.build();
-    }
-
-    private ErrorType mapErrorType(RemoteException exception) {
```
pkoenig10, line 82: Had to refactor this a little bit since the error code in the remote exception might not satisfy the error name regex in `ErrorType`.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
-
-        try {
-            SerializableError error = SerializableError.builder()
-                    .errorName(errorType.name())
-                    .errorCode(errorType.code().toString())
-                    .errorInstanceId(exception.getError().errorInstanceId())
-                    .build();
-            builder.type(MediaType.APPLICATION_JSON);
-            builder.entity(error);
-        } catch (RuntimeException e) {
```
pkoenig10, line 68: These catch blocks appear to be a holdover from when we performed the serialization manually in the exception mapper. This was changed in https://github.com/palantir/conjure-java-runtime/pull/675 and these error messages are now misleading/incorrect.

This code should never throw exceptions, so these catch blocks can be removed.

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/ExceptionMappingTest.java
```diff
@@ -121,6 +121,7 @@ public void testRemoteException() {
         assertThat(error.errorInstanceId()).isEqualTo("errorInstanceId");
         assertThat(error.errorCode()).isEqualTo(ErrorType.INTERNAL.code().toString());
         assertThat(error.errorName()).isEqualTo(ErrorType.INTERNAL.name());
+        assertThat(error.parameters()).isEmpty();
```
pkoenig10, line 124: Adding these asserts to ensure we are not propagating parameters.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
carterkozak: Most likely scenario is a new `Code` (several existing codes share status codes). In that case we wouldn't want older services to fail to propagate that data :+1:

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
-        }
-        return builder.build();
-    }
-
-    private ErrorType mapErrorType(RemoteException exception) {
-        Status status = Status.fromStatusCode(exception.getStatus());
-
-        if (status.getStatusCode() == 401) {
-            log.info(
-                    "Encountered a remote unauthorized exception."
```
carterkozak, line 87: I'd prefer to keep around the more specific log lines based on status code. Makes it easier to sift through the data, and decouples their log levels.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
-        } else if (status.getStatusCode() == 403) {
+        if (exception.getStatus() == 401 || exception.getStatus() == 403) {
             log.info(
-                    "Encountered a remote permission denied exception."
-                            + " Mapping to a default permission denied exception before propagating",
+                    "Encountered a remote exception",
                     SafeArg.of("errorInstanceId", exception.getError().errorInstanceId()),
                     SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
+                    SafeArg.of("status", exception.getStatus()),
```
carterkozak: Any reason you renamed the args? Slight preference to keep them aligned

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
pkoenig10: We already log the status as an arg here, shouldn't that be sufficient to filter logs?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ServiceExceptionMapper.java
```diff
@@ -55,23 +54,13 @@ public Response toResponseInner(ServiceException exception) {
...
-                    e);
-            // simply write out the exception message
-            builder.type(MediaType.TEXT_PLAIN);
-            builder.entity("Unable to translate exception to json. Refer to the server logs with this errorInstanceId: "
-                    + exception.getErrorInstanceId());
-        }
-        return builder.build();
+        SerializableError error = SerializableError.builder()
+                .from(SerializableError.forException(exception))
+                .build();
```
carterkozak: I realize this was the same before, but is there any reason for the additional builder?
```suggestion
        SerializableError error = SerializableError.forException(exception);
```

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
pkoenig10: I noticed that we use `status` in other places so I was trying to be internally consistent.
```
conjure-java-runtime pkoenig/propagate .$ rg Arg.*status
conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
62:                    SafeArg.of("status", exception.getStatus()),
78:                    SafeArg.of("status", exception.getStatus()),

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/NeverReturnNullDecoder.java
37:        Preconditions.checkNotNull(object, "Unexpected null body", SafeArg.of("status", response.status()));

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/feignimpl/NeverReturnNullDecoderTest.java
49:                .containsArgs(SafeArg.of("status", 200));
59:                .containsArgs(SafeArg.of("status", 200));
69:                .containsArgs(SafeArg.of("status", 200));
```

But changing this is just going to create unnecessary pain for log consumers so I'll revert this.

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ServiceExceptionMapper.java
pkoenig10: No, will remove

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
carterkozak: It should be, although filtering on args only seems to work intermittently for me. Numeric value handling is wonky due to the floating point coercion. Not a strong opinion on this, just what I would have done.

carterkozak:
I think this change is reasonable, and makes auth failure causes more obvious. I'm going to give folks a couple days to review and provide feedback before merging in case there are sharp edges I haven't considered.
@uschi2000 and @jkozlowski, what do you think?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
```
jkozlowski, line 54: There's some Javadoc in this class:

```
<p>To preserve debuggability, the exception and HTTP status code from B's exception are logged at WARN level, but not
 * propagated to caller to avoid an unintentional dependency on the remote exception.
```

Perhaps needs an update?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
jkozlowski: Ok, so for 401/403 just propagate the upstream errorCode and errorName, because it's potentially got useful information. +1

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
```diff
@@ -54,64 +54,35 @@
...
                     SafeArg.of("errorName", exception.getError().errorName()),
-                    SafeArg.of("statusCode", status.getStatusCode()),
+                    SafeArg.of("statusCode", exception.getStatus()),
                     exception);
 
-            return ErrorType.INTERNAL;
+            ServiceException propagatedException = new ServiceException(ErrorType.INTERNAL, exception);
+            return Response.status(propagatedException.getErrorType().httpErrorCode())
+                    .type(MediaType.APPLICATION_JSON)
+                    .entity(SerializableError.forException(propagatedException))
```
jkozlowski, line 84: This bit is a bit more difficult for me to reason through, but I think I get it: by wrapping it in ServiceException, we're basically not propagating whatever information there was in the RemoteException, instead we're sending a nice clean 500, we're keeping the errorInstanceId from RemoteException and we're not propagating any args.

Perhaps this would make it more obvious?

```
            ErrorType internalError = ErrorType.INTERNAL;
            return Response.status(internalError.httpErrorCode())
                    .type(MediaType.APPLICATION_JSON)
                    .entity(SerializableError.builder()
                            .errorCode(internalError.code().name())
                            .errorName(internalError.name())
                            .errorInstanceId(exception.getError().errorInstanceId())
                            .build())
                    .build();
```

I dunno if it's just because I'm less familiar with this particular codepath, but I had to grok what #forException and ServiceException constructor to check what this is going to do. Creating an exception to then destroy it and return a SerializableError really threw me off, I wanted to confirm we're not sending back some stacktraces and args.

conjure-java-jersey-server/src/test/java/com/palantir/conjure/java/server/jersey/ExceptionMappingTest.java
jkozlowski: Does the test server actually provide any arguments?

```
            throw new RemoteException(
                    SerializableError.builder()
                            .errorInstanceId("errorInstanceId")
                            .errorCode("errorCode")
                            .errorName("errorName")
                            .build(),
                    REMOTE_EXCEPTION_STATUS_CODE);
```

jkozlowski (commented):
Change sounds useful, I've added some nits. And mostly have questions:

1. My only concern is that this will appear as if it's coming from the first-hop service, perhaps we could indicate something on the response to show that it's propagated from some upstream? 
2. Why shouldn't we propagate all the RemoteException error codes?
3. I am not so familiar with the types of errors that the internal authorization service throws, but are they actually grokkable to the user? I'm basically curious how much MORE useful is the internal error code, without seeing the params?

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/RemoteExceptionMapper.java
pkoenig10: This pattern is used in the Undertow exception handler:
https://github.com/palantir/conjure-java/blob/develop/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureExceptions.java

I slightly prefer this approach because it delegates some of the tricker bits to centralized static factory methods.

pkoenig10:
Answering questions:
1. I think the required namespacing of error names should be sufficient here. Let me know if you disagree.
2. For other 400 responses we don't know whether it is the end user or an intermediary service that provided bad arguments. For a 401 and 403 we can be reasonably confident that the root cause is that the end user is unauthenticated or unauthorized.
3. These errors are definitely grokkable to end users and refer to user-facing concepts - things like lacking access to markings, attempting to create resources that violate marking constraints, etc. However at the end of the day, we still should/will rely on the FE to display these errors in a more legible way than the raw response. But the benefit here is that there is a single error type for these things, rather than every service redefining these errors just to propagate them.

uschi2000:
I can see pros and cons. Pragmatic decision to propagate them (even though it makes the overall behavior a little more complex) is probably OK, in particular if this information is to be used to provide users with better explanations of permission issues.

carterkozak (approved):

PR #2022, opened 2021-07-06
Replace aged afterburner with the sr71 blackbird module

carterkozak:
Afterburner uses several internal methods which result in warnings in
modern jdks, and result in failures using jdk16+.
The replacement Blackbird module uses LambdaMetafactory for a more
sustainable solution.

Documentation:
https://github.com/FasterXML/jackson-modules-base/tree/2.13/blackbird#jackson-blackbird

==COMMIT_MSG==
Replace aged afterburner with the sr71 blackbird module
==COMMIT_MSG==

## Possible downsides?
Sometimes jackson changes can be brittle, it's possible bugs exist in the blackbird module that weren't present in afterburner. We have a great deal of test coverage, so I'm not worried as long as tests pass.


fawind (approved):

iamdanfox:
@carterkozak I vaguely recall the folks who work on our internal auth'd elasticsearch caring a _lot_ about how fast serde is... do we have any predictions about how fast blackbird is compared to afterburner?? (is it the same codegen just loaded in a different way? or a whole different approach)

carterkozak:
More or less equivalent, except that it only applies to method calls and not direct field access. Fortunately conjure exclusively uses methods. There may be some variation in types with annotated fields, but I don't expect it to be substantial.

PR #2051, opened 2021-07-23
Excavator: Migrate to the logsafe SafeLogger API

carterkozak:
###### _excavator_ is a bot for automating changes across repositories.

Changes produced by the roomba/logsafe-logger-migration check.

To enable or disable this check, please contact the maintainers of Excavator.


carterkozak:
Recreated manually since the excavator hasn't been enabled for oss repos yet.

fawind (approved):

conjure-java-jaxrs-client/build.gradle
```diff
@@ -23,8 +24,6 @@ dependencies {
         // the shipped version clashes with the newer javax.ws.rs:javax.ws.rs-api used by (e.g.) dropwizard
         exclude group: "javax.ws.rs", module: "jsr311-api"
     }
-    implementation "org.slf4j:slf4j-api"
```
jkozlowski, line 26: I think excavator needs an update to handle this?

jkozlowski (approved):

conjure-java-jaxrs-client/build.gradle
carterkozak: Considering that out of scope for now, we don't guarantee all slf4j uses are removed, only those that we can cleanly migrate over

PR #2104, opened 2021-09-13
CJR CipherSuites list is more restrictive, avoiding less secure ciphers

carterkozak:
==COMMIT_MSG==
CJR CipherSuites list is more restrictive, avoiding less secure ciphers
==COMMIT_MSG==

## Possible downsides?
Could fail to connect to insecure servers. This is a feature.

### DIFF
--- a/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
+++ b/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
@@ -277,13 +277,8 @@ public final class OkHttpClients {
                     .build());
         }
 
-        // cipher setup
-        client.connectionSpecs(createConnectionSpecs(config.enableGcmCipherSuites()));
-        // gcm ciphers are required for http/2 per https://tools.ietf.org/html/rfc7540#section-9.2.2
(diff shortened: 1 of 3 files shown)


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -339,14 +334,11 @@ private static UserAgent augmentUserAgent(UserAgent agent, Class<?> serviceClass
         return augmentedAgent;
     }
 
-    private static ImmutableList<ConnectionSpec> createConnectionSpecs(boolean enableGcmCipherSuites) {
+    private static ImmutableList<ConnectionSpec> createConnectionSpecs() {
```
jkozlowski, line 337: What are the blockers for removing okhttp support entirely?

jkozlowski (approved):

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
carterkozak: I think a lot of folks still use it in tests, I’ll do a more thorough investigation later though

PR #2134, opened 2021-10-14
Use raw binary in Smile encoding

sfackler:
## Before this PR
Binary data was encoded in Smile's 7-bit format, resulting in a 14% increase in the encoded size and requiring more computation to encode and decode.

## After this PR

==COMMIT_MSG==
Smile encoding now uses the raw binary format.
==COMMIT_MSG==

The raw format just directly copies the bytes through unmodified. This does mean that the encoded representation can include embedded control characters like the smile EOS marker, but we don't use that when serializing anyway.

### DIFF
--- a/changelog/@unreleased/pr-2134.v2.yml
+++ b/changelog/@unreleased/pr-2134.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Smile encoding now uses the raw binary format.
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/2134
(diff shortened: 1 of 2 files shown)


carterkozak:
@sfackler Can you confirm that this wouldn't result in any incompatibility between old and new versions of CJR? I imagine there should be a field that tells the receiver how binary data is encoded?

sfackler:
Yep, safe binary and raw binary are have separate type tokens identifying them (0xe8 and 0xfd respectively). Additionally, there's a bit in the Smile message header saying that the value may contain raw binary values. The only thing you lose from enabling this is the ability to easily skip over entire objects by searching for the 0xff end-of-stream token, but we don't include that in our serialized output anyway:

https://github.com/FasterXML/smile-format-specification/blob/master/smile-specification.md#token-class-misc-binary--text--structure-markers

The raw binary feature has been part of the spec for at least 4 years, and judging by the changelog has probably been there since creation.

carterkozak (approved):
Sounds great, thanks for clarifying :-)

PR #2175, opened 2021-11-29
Use ObjectMapper subtypes and builders

pkoenig10:
In Jackson 2.13.0, in preparation for Jackson 3.0, some `ObjectMapper` configuration methods have been deprecated in favor of using the `ObjectMapper` subtype builders.

Because the existing `ObjectMappers` methods return `ObjectMapper`, consumers cannot modify/extend these object mapper using the `rebuild` functionality. For example, Witchcraft uses these methods to build a configuration deserialization object mapper.

This PR adds `ObjectMappers` methods that return the `ObjectMapper` subtypes and construct the object mappers using builders. In future PRs, we should deprecate and ultimately remove the existing methods.

For more details see [this Jackson 3.0 blog post](https://cowtowncoder.medium.com/jackson-3-0-immutability-w-builders-d9c532860d88).

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
@@ -20,9 +20,12 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
-import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
+import com.fasterxml.jackson.databind.cfg.MapperBuilder;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import com.fasterxml.jackson.dataformat.cbor.databind.CBORMapper;
 import com.fasterxml.jackson.dataformat.smile.SmileFactory;
 import com.fasterxml.jackson.dataformat.smile.SmileGenerator;
+import com.fasterxml.jackson.dataformat.smile.databind.SmileMapper;
 import com.fasterxml.jackson.datatype.guava.GuavaModule;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.joda.JodaModule;
@@ -33,6 +36,10 @@ public final class ObjectMappers {
 
     private ObjectMappers() {}
 
+    private static final SmileFactory SMILE_FACTORY = SmileFactory.builder()
+            .disable(SmileGenerator.Feature.ENCODE_BINARY_AS_7BIT)
+            .build();
+
     /**
      * Returns a default ObjectMapper with settings adjusted for use in clients.
      *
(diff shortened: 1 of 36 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -113,10 +126,68 @@ public static ObjectMapper newCborServerObjectMapper() {
      *   <li>Throw on unknown properties found during deserialization.
      * </ul>
      */
+    public static SmileMapper newServerSmileMapper() {
+        return withDefaultModules(SmileMapper.builder(SMILE_FACTORY))
+                .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+                .build();
+    }
+
+    public static ObjectMapper newClientObjectMapper() {
```
pkoenig10, line 135: Would like to eventually deprecate these methods. I think we can fix this automatically with the `@InlineMe` annotation. Do you know if our infra automatically applies suggested fixes for that check?

http://errorprone.info/docs/inlineme

conjure-scala-jaxrs-client/src/main/java/com/palantir/conjure/java/serialization/ScalaObjectMappers.java
```diff
@@ -19,38 +19,65 @@
...
+    public static JsonMapper newServerJsonMapper() {
+        return withScalaSupport(ObjectMappers.newServerJsonMapper());
+    }
+
+    public static CBORMapper newServerCborMapper() {
+        return withScalaSupport(ObjectMappers.newServerCborMapper());
+    }
+
+    public static SmileMapper newServerSmileMapper() {
+        return withScalaSupport(ObjectMappers.newServerSmileMapper());
```
carterkozak, line 53: no scala on new apis please

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
carterkozak: It does not -- we'd need to handle running fixes on all dep updates in addition to baseline upgrades, otherwise some excavators would fail.

I'm not sure inline-me would produce code that actually compiles, it depends how the result is used. Prefer to handle automated migrations in isolation before deprecation

conjure-scala-jaxrs-client/src/main/java/com/palantir/conjure/java/serialization/ScalaObjectMappers.java
pkoenig10: This is used here:
https://github.com/palantir/conjure-java-runtime/blob/e10e3578a38cb07310ddbb31b6416f1e6dd811de/conjure-scala-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/FeignJaxRsScalaClientBuilder.java#L27-L28

I originally changed the return type of the `AbstractFeignJaxRsClientBuilder` methods since they were package private. But given that someone _could_ extend this, I decided not to.

Would you prefer to override these methods to use the more specific types wherever possible? Or just change the `FeignJaxRsScalaClientBuilder` constants to `ObjectMapper`?

pkoenig10: This is also used in the Witchcraft PR that follows from this PR.

pkoenig10:
@carterkozak If you have time to take another look at this, I'd really appreciate it!

carterkozak (approved):
Thanks, @pkoenig10!

PR #2304, opened 2022-03-21
Metrics to determine when a feign client dangerously buffers data

carterkozak:
We are phasing out feign clients, this allows us to detect the
most critical clients to migrate to dialogue.

==COMMIT_MSG==
Metrics to determine when a feign client dangerously buffers data
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jaxrs-client/src/main/metrics/dangerous-buffering.yml
+++ b/conjure-java-jaxrs-client/src/main/metrics/dangerous-buffering.yml
@@ -13,7 +13,8 @@ namespaces:
           - name: direction
             values: [request, response]
         docs: |
-          Meter incremented each time a feign client is used to send or receive streamed binary data. Feign should
-          never be used in this way because it fully buffers the request and response internally, resulting in heavy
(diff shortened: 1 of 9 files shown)


conjure-java-jaxrs-client/src/main/metrics/dangerous-buffering.yml
```diff
@@ -0,0 +1,20 @@
+options:
+  javaPackage: com.palantir.conjure.java.client.jaxrs.feignimpl
+  javaVisibility: packagePrivate
+namespaces:
+  feign.client:
+    docs: Metrics produced by Feign clients
+    metrics:
+      dangerous.buffering:
+        type: meter
```
fawind, line 9: Curious, why isn't this a counter given that we only ever increment by 1?

carterkozak: It's incremented by the number of bytes we write -- meter and counter are the same in that regard, but meter cannot be decremented

fawind (approved):

PR #2316, opened 2022-03-29
WebApplicationExceptionMapper treats redirection as informational

schlosna:
## Before this PR
RedirectionExceptions were being logged at `ERROR`, especially HTTP 304 Not Modified

## After this PR

==COMMIT_MSG==
WebApplicationExceptionMapper treats redirection as informational
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/WebApplicationExceptionMapper.java
+++ b/conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/WebApplicationExceptionMapper.java
@@ -52,10 +52,10 @@ final class WebApplicationExceptionMapper extends ListenableExceptionMapper<WebA
     public Response toResponseInner(WebApplicationException exception) {
         String errorInstanceId = UUID.randomUUID().toString();
 
-        if (exception.getResponse().getStatus() / 100 == 4 /* client error */) {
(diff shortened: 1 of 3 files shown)


carterkozak (approved):

conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/WebApplicationExceptionMapper.java
```diff
@@ -52,10 +52,10 @@
     public Response toResponseInner(WebApplicationException exception) {
         String errorInstanceId = UUID.randomUUID().toString();
 
-        if (exception.getResponse().getStatus() / 100 == 4 /* client error */) {
-            log.info("Error handling request", SafeArg.of("errorInstanceId", errorInstanceId), exception);
-        } else {
+        if (exception.getResponse().getStatusInfo().getFamily() == Response.Status.Family.SERVER_ERROR) {
             log.error("Error handling request", SafeArg.of("errorInstanceId", errorInstanceId), exception);
+        } else {
+            log.info("Error handling request", SafeArg.of("errorInstanceId", errorInstanceId), exception);
```
schlosna, line 58: I didn't change the behavior on the info logging here, but a case could be made we probably don't need the exception for non-5xx responses

PR #2320, opened 2022-03-31
Add marker annotations to indicate JaxRs client or service usage

fawind:
## Before this PR

When migrating hand-rolled services to the [conjure-undertow annotation processor](https://github.com/palantir/conjure-java/tree/develop/conjure-undertow-annotations), we are currently missing the information if a JAX-RS interface is used as service or client definition (or both).

## After this PR

Add marker annotations that allow us to indicate whether a JAX-RS interface is used as a client or service.

==COMMIT_MSG==
Add marker annotations to indicate JaxRs client or service usage
==COMMIT_MSG==

### DIFF
--- a/conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyClient.java
+++ b/conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyClient.java
@@ -1,0 +1,30 @@
+/*
+ * (c) Copyright 2022 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicator annotation for marking that a JAX-RS interface is used to construct a Jersey client.
+ * See also {@link JerseyService}.
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
+public @interface JerseyClient {}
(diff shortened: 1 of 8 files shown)


conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyClient.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicator annotation for marking that a JAX-RS interface is used to construct a Jersey client.
+ * See also {@link JerseyService}.
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
+public @interface JerseyClient {}
```
fawind: Open to naming suggestions. This is currently specific to Jersey. We might want to make this framework agnostic and name it something like `ClientDefinition` and `ServerDefinition`. But not sure if we have further use-cases outside the jersey migration in mind yet.

carterkozak: Maybe `JaxRsClient`? Jersey is a jax-rs implementation, and we don't use any of its client functionality. Feign is our jax-rs client-side implementation.

We could call the other annotation `JaxRsServer` for a clear client/server distinction.

Thoughts on using `RUNTIME` retention, that way our client factories can be updated to throw if an interface is annotated incorrectly?

conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyService.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicator annotation for marking that a JAX-RS interface is used for registering a Jersey resource. See also
+ * {@link JerseyClient}.
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.TYPE)
```
carterkozak: We may want to make these `@Inherited` for ease of checking resource types.

conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyClient.java
fawind: `JaxRs...` sounds good, updated!

Also like the validation bit! Updated retention to runtime.

conjure-java-annotations/src/main/java/com/palantir/conjure/java/annotations/JerseyService.java
fawind: Added

carterkozak (approved):

fawind:
> Thoughts on using RUNTIME retention, that way our client factories can be updated to throw if an interface is annotated incorrectly?

Will follow up adding that check in a separate PR!

PR #2435, opened 2022-08-09
Create jakarta equivalents of conjure-java-runtime

mglazer:
## Before this PR

There was no way to use this library in the context of jakarta upstream libraries.

## After this PR

==COMMIT_MSG==
Create jakarta equivalents of conjure-java-runtime

This change creates -jakarta equivalents of conjure-java-jaxrs-client
and conjure-java-jaxrs-server. This enables a gradual migration to the
jakarta EE versions of upstream libraries.

In addition to these changes, a few other minor changes had to occur:

* Migrated to Junit5 for tests which needed to depend on a newer version
  of Jakarta
* Upgraded tracing-java and conjure-java since we needed to take
  on those upstream changes.
* Removed revapi from conjure-java-jersey-server since it was causing no end of issues with GCV. I suspect because it locks into the produced API configuration which is kind of funky when using GCV and our method of disabling it on a per-project basis.
* Deleted tests from conjure-java-jersey-server since they now cause conflicts due to Dropwizard + Jersey being bumped elsewhere in the classpath.
* This change bumps okhttp in the test dependencies, but the locked version is still 3.13.1 which should prevent the documented 4.x bug from appearing. The reason for this bump was the 3.x line of okhttp still used legacy junit rules, and with the upgrade to junit5 I had to take a newer version of the mockhttpserver.
* This change bumps the metrics-core dependency to 4.x. This is probably the most dramatic version bump change and will affect downstream projects. Since tritium pins to 4.1.1 (https://github.com/palantir/tritium/blob/develop/versions.props#L32) and thus that lines up with Spark, the comment that was left behind in this repository no longer seems to apply.
==COMMIT_MSG==

## Possible downsides?

The metrics-core bump could be unsafe for downstream consumers, but given this aligns it with Tritium that seems like it should be fine and the comment is no longer correct.


settings.gradle
```diff
@@ -2,6 +2,7 @@ rootProject.name = 'conjure-java-runtime'
 
 include 'conjure-java-annotations'
 include 'conjure-java-jaxrs-client'
+include 'conjure-java-jaxrs-jakarta-client'
```
carterkozak: I'm not sure that we can get away with branching conjure-java-jaxrs-clients given the migration won't be typesafe (input is a `Class<?>`). I've meant to look into replacing our feign contract implementation to support both sets of annotations so that we can support existing interfaces more easily.

conjure-java-jaxrs-client/src/test/java/com/palantir/conjure/java/client/jaxrs/JaxRsClientFailoverTest.java
```diff
@@ -30,44 +30,50 @@
...
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
-import org.junit.Test;
-import org.junit.experimental.theories.DataPoints;
-import org.junit.experimental.theories.FromDataPoints;
-import org.junit.experimental.theories.Theories;
-import org.junit.experimental.theories.Theory;
-import org.junit.runner.RunWith;
-
-@RunWith(Theories.class)
```
mglazer, line 42: This was Theories before...I honestly cannot tell how theories was different from parameterized tests in this case.

conjure-java-jersey-jakarta-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -0,0 +1,146 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.server.jersey;
+
+import com.fasterxml.jackson.jakarta.rs.cbor.JacksonCBORProvider;
```
mglazer, line 19: this import has changed.

conjure-java-jersey-server/build.gradle
```diff
@@ -1,14 +1,18 @@
...
 apply plugin: 'com.palantir.metric-schema'
 
+versionsLock {
+    disableJavaPluginDefaults()
+}
+
 dependencies {
     api "com.palantir.conjure.java.api:errors"
-    api "org.glassfish.jersey.core:jersey-server"
+    api "org.glassfish.jersey.core:jersey-server:2.26"
```
mglazer, line 10: The previous jersey version was 2.22.2, this bumps to 2.26 because I was having significant issues with version alignment otherwise.

conjure-java-server-verifier/build.gradle
```diff
@@ -6,14 +6,13 @@ dependencies {
     verifier "com.palantir.conjure.verification:verification-client::${osClassifier}@tgz"
 
     testImplementation project(':conjure-java-jaxrs-client')
-    testImplementation project(':conjure-java-jersey-server')
+    testImplementation project(':conjure-java-jersey-jakarta-server')
```
mglazer, line 9: this means we're not verifying against the legacy server anymore.

gradle/verifier.gradle
```diff
@@ -94,7 +94,7 @@ task conjureJavaObjects(type: Exec, dependsOn: unpackGenerator) {
 
 task conjureJavaJersey(type: Exec, dependsOn: unpackGenerator) {
     executable "${-> unpackGenerator.getExecutable()}"
-    args 'generate',  "${-> configurations.verificationApi.singleFile}", 'src/generatedJersey/java', '--jersey'
+    args 'generate',  "${-> configurations.verificationApi.singleFile}", 'src/generatedJersey/java', '--jersey', "--jakartaPackages"
```
mglazer, line 97: generate the packages in the new format.

versions.lock
```diff
@@ -1,172 +1,175 @@
...
+com.palantir.tracing:tracing-api:6.13.0 (5 constraints: 9e527e9d)
+com.palantir.tracing:tracing-jersey-jakarta:6.13.0 (1 constraints: 3c05503b)
+com.palantir.tracing:tracing-okhttp3:6.13.0 (1 constraints: 3c05503b)
 com.palantir.tritium:tritium-api:0.51.0 (2 constraints: 2d1f6cbc)
 com.palantir.tritium:tritium-core:0.51.0 (1 constraints: 3e103fa2)
 com.palantir.tritium:tritium-metrics:0.51.0 (5 constraints: 054d2158)
 com.palantir.tritium:tritium-registry:0.51.0 (6 constraints: 425dead1)
 com.squareup.okhttp3:logging-interceptor:3.13.1 (1 constraints: 3a053f3b)
-com.squareup.okhttp3:okhttp:3.13.1 (6 constraints: 8459d8f9)
+com.squareup.okhttp3:okhttp:3.13.1 (6 constraints: 5151cdf6)
```
mglazer, line 46: actually looks like we're still pinned to a 3.13.1 which is great.

versions.lock
```diff
@@ -1,172 +1,175 @@
...
-joda-time:joda-time:2.10.8 (2 constraints: b4238380)
-org.apache.httpcomponents.client5:httpclient5:5.1.3 (1 constraints: cd13956e)
-org.apache.httpcomponents.core5:httpcore5:5.1.4 (3 constraints: 6139a70d)
-org.apache.httpcomponents.core5:httpcore5-h2:5.1.3 (1 constraints: 3d13093c)
-org.checkerframework:checker-qual:3.24.0 (3 constraints: ae1f0252)
-org.glassfish:jakarta.el:3.0.3 (2 constraints: bf231f5c)
-org.glassfish.hk2:hk2-api:2.6.1 (1 constraints: 6e0c620c)
-org.glassfish.hk2:hk2-locator:2.6.1 (1 constraints: 4410e0ad)
-org.glassfish.hk2:hk2-utils:2.6.1 (2 constraints: 21173277)
+io.dropwizard.metrics:metrics-core:4.2.10 (22 constraints: 035b00eb)
```
mglazer, line 52: this bumps our metrics-core dependency since it was conflicting all over the place at runtime.

mglazer:
👍

mglazer:
Closing via: https://github.com/palantir/conjure-java-runtime/pull/2440

PR #2436, opened 2022-08-09
conjure-java-jaxrs-client supports javax and jakarta annotations

carterkozak:
==COMMIT_MSG==
conjure-java-jaxrs-client supports javax and jakarta annotations
==COMMIT_MSG==

Note there are still javax jaxrs dependencies on the classpath, however that won't prevent both sets of annotations from functioning correctly.

## Possible downsides?
No test coverage of the jakarta path due to dependency issues.


mglazer (approved):

carterkozak:
Merging based on Glazer’s approval

PR #2438, opened 2022-08-11
Rewrite all tests away from Dropwizard

mglazer:
## After this PR

==COMMIT_MSG==
Rewrite all tests away from Dropwizard

Dropwizard behaves poorly with GCV since it inadvertently locks
a number of our dependencies into versions that we don't want.
Concretely this has a problem when we are trying to upgrade
our dependencies to the Jakarta EE 9 versions, which only very
recent versions of Dropwizard support, and those versions bring
in dependencies of libraries like slfj4 2.0.0 ALPHA which conflicts
with other services.

This is a large change...I hope I never have to do this again.

Other changes while I was in the code:

* Migrated to Junit5 where possible
==COMMIT_MSG==


conjure-java-server-verifier/src/test/java/com/palantir/verification/server/AutoDeserializeTest.java
```diff
@@ -20,87 +20,81 @@
...
+                                                .getNegative()
+                                                .get(i))));
+            });
+        }
     }
 
-    @Test
-    public void runTestCase() {
+    @ParameterizedTest(name = "{0}({3}) -> should succeed {2} with client {4}")
+    @ArgumentsSource(Parameters.class)
```
carterkozak, line 84: TIL: I hadn't used class-based argument sources, but I like it a lot better than specifying a method name as a string. Thanks!

mglazer: Oh ya, I've always used ArgumentsProvider instead of the string based method. Typesafety is so much nicer, and the string based methods _only_ make sense if you're using a language like Groovy which support table based spec testing.

versions.lock
```diff
@@ -64,125 +64,83 @@ jakarta.activation:jakarta.activation-api:1.2.2 (2 constraints: ba28fcbd)
...
-org.glassfish.jersey.media:jersey-media-jaxb:2.34 (1 constraints: e01957ca)
 org.glassfish.jersey.media:jersey-media-json-jackson:2.34 (1 constraints: dd04fd30)
-org.glassfish.jersey.test-framework:jersey-test-framework-core:2.34 (1 constraints: 4b23d8ba)
-org.glassfish.jersey.test-framework.providers:jersey-test-framework-provider-inmemory:2.34 (1 constraints: cd0ef25f)
 org.hamcrest:hamcrest:2.1 (2 constraints: 8512932d)
 org.hamcrest:hamcrest-all:1.3 (1 constraints: a8041d2c)
 org.hamcrest:hamcrest-core:2.1 (2 constraints: 88126e10)
 org.hamcrest:hamcrest-library:2.1 (1 constraints: 1507415c)
+org.jboss.spec.javax.annotation:jboss-annotations-api_1.3_spec:2.0.1.Final (1 constraints: 150fc165)
+org.jboss.spec.javax.servlet:jboss-servlet-api_4.0_spec:2.0.0.Final (1 constraints: 140fba65)
```
carterkozak: nit: let's exclude these in favor of the standard servlet/annotation-apis. I'd expect classpath uniqueness checks to fail here, I'm not sure why they aren't...

undertow-testing/build.gradle
```diff
@@ -0,0 +1,17 @@
+// Avoid including test dependencies from this module in
+// gradle module metadata.
+versionsLock {
+    testProject()
+}
+
+dependencies {
+    api 'org.junit.jupiter:junit-jupiter'
+    api 'io.undertow:undertow-core'
+    api 'io.undertow:undertow-servlet'
```
carterkozak:
```suggestion
    api 'io.undertow:undertow-servlet', {
        // Use  standard servlet-api instead
        exclude group: 'org.jboss.spec.javax.servlet', module: 'jboss-servlet-api_4.0_spec'
        exclude group: 'org.jboss.spec.javax.annotation', module: 'jboss-annotations-api_1.2_spec'
    }
    api <servlet-api>
    implementation <annotation-api>
```

versions.lock
mglazer: I have stanzas in my build.gradle which are doing the module replacements...

undertow-testing/build.gradle
mglazer: Done. It was actually api_1.3_spec....I don't know how anyone is expected to remember these package names.

carterkozak (approved):
Love it, thanks! This is a much better representation of our internal services :-)

mglazer:
👍

PR #2569, opened 2023-03-22
Report metrics when unusually large strings are parsed

carterkozak:
==COMMIT_MSG==
Report metrics when unusually large strings are parsed
==COMMIT_MSG==

Unfortunately this only covers JSON at the moment, and only JSON parsed using mappers from Conjure factory methods. I'd like to capture data from the smile mappers as well, but I don't want to block beginning to capture data on that.

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
@@ -1,0 +1,710 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
+import com.codahale.metrics.Histogram;
+import com.fasterxml.jackson.core.Base64Variant;
+import com.fasterxml.jackson.core.FormatSchema;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonFactoryBuilder;
+import com.fasterxml.jackson.core.JsonLocation;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonStreamContext;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.ObjectCodec;
+import com.fasterxml.jackson.core.SerializableString;
+import com.fasterxml.jackson.core.StreamReadCapability;
+import com.fasterxml.jackson.core.StreamReadFeature;
+import com.fasterxml.jackson.core.TSFBuilder;
+import com.fasterxml.jackson.core.TreeNode;
+import com.fasterxml.jackson.core.Version;
+import com.fasterxml.jackson.core.async.NonBlockingInputFeeder;
+import com.fasterxml.jackson.core.io.IOContext;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.core.util.JacksonFeatureSet;
+import com.fasterxml.jackson.core.util.RequestPayload;
+import com.google.common.util.concurrent.RateLimiter;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalStateException;
+import com.palantir.logsafe.exceptions.SafeRuntimeException;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
+import com.palantir.tritium.metrics.registry.SharedTaggedMetricRegistries;
+import com.palantir.tritium.metrics.registry.TaggedMetricRegistry;
+import java.io.DataInput;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.io.Writer;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Iterator;
+import javax.annotation.Nullable;
+
+final class InstrumentedJsonFactory extends JsonFactory {
+
+    // Log at most once per second
+    private static final RateLimiter LOGGING_RATE_LIMITER = RateLimiter.create(1);
+    private static final SafeLogger log = SafeLoggerFactory.get(InstrumentedJsonFactory.class);
+    private final Histogram parsedStringLength;
+    private final Throwable creationStackTrace;
+
+    // Using the shared metric registry singleton to avoid API churn in methods that use this instrumentation.
+    @SuppressWarnings("deprecation")
+    InstrumentedJsonFactory() {
+        this(SharedTaggedMetricRegistries.getSingleton());
+    }
+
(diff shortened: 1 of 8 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+        }
+
+        @Override
+        public boolean nextFieldName(SerializableString str) throws IOException {
+            return delegate.nextFieldName(str);
+        }
+
+        @Override
+        public String nextFieldName() throws IOException {
+            return delegate.nextFieldName();
```
schlosna, line 336: Do we want to track field name lengths as well?

[Looking at `StreamReadConstraints`](https://github.com/FasterXML/jackson-core/blob/2.15/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java#L30-L43) we probably want to track number length (default limit `1000`), nesting depth (default limit `1000`), and string length (default limit `5_000_000`)

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+        }
+
+        @Override
+        public double getDoubleValue() throws IOException {
+            return delegate.getDoubleValue();
+        }
+
+        @Override
+        public BigDecimal getDecimalValue() throws IOException {
+            return delegate.getDecimalValue();
```
schlosna, line 533: Do we need to track max length here as well (default limit `1000`)? Would we want to log warning if we exceed default limit?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+
+        @Override
+        public byte[] getBinaryValue(Base64Variant bv) throws IOException {
+            return delegate.getBinaryValue(bv);
+        }
+
+        @Override
+        public byte[] getBinaryValue() throws IOException {
+            return delegate.getBinaryValue();
+        }
```
schlosna, line 554: Do we need to track byte array lengths?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+
+        @Override
+        public long getLongValue() throws IOException {
+            return delegate.getLongValue();
+        }
+
+        @Override
+        public BigInteger getBigIntegerValue() throws IOException {
+            return delegate.getBigIntegerValue();
```
schlosna, line 518: Do we need to track max length here as well (default limit `1000`)? Would we want to log warning if we exceed default limit?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+
+        private String recordStringLength(String value) {
+            if (value != null) {
+                int length = value.length();
+                parsedStringLength.update(length);
+                if (length > 1_000_000 && LOGGING_RATE_LIMITER.tryAcquire()) {
+                    log.info(
+                            "Detected an unusually large JSON string value",
+                            SafeArg.of("length", length),
+                            new SafeRuntimeException("Pared here", factoryCreationStackTrace));
```
schlosna: nit:
```suggestion
                            new SafeRuntimeException("Parsed here", factoryCreationStackTrace));
```

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
```diff
@@ -0,0 +1,710 @@
...
+            this.parsedStringLength = parsedStringLength;
+            this.factoryCreationStackTrace = factoryCreationStackTrace;
+        }
+
+        private String recordStringLength(String value) {
+            if (value != null) {
+                int length = value.length();
+                parsedStringLength.update(length);
+                if (length > 1_000_000 && LOGGING_RATE_LIMITER.tryAcquire()) {
+                    log.info(
```
schlosna: warn?

schlosna (commented):
Overall looks good, had a few questions

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
carterkozak: Ideally I'd like to track number length, however it's tricky because precision may be lost when we read a value, and we may need to work backwards to understand the input data size.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
carterkozak: 👍 Fixed, and moved to a new class which manages the observability pieces, shared between the smile and json implementations

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
carterkozak: I could go either way, initially I want to understand the state of the world before I ask teams to investigate themselves, but it's unlikely folks will take action based on a new `log.warn` any more than a `log.info` until we add monitors.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -81,7 +77,7 @@ public static CBORMapper newClientCborMapper() {
      * </ul>
```
schlosna, line 77: Do we need to instrument CBOR mappers as well?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
carterkozak: I don't think this is enforced in Jackson. I agree that json array sizes (including but not limited to byte-arrays) should be bounded similarly to string length, but I'd prefer to build up to that later if that's alright with you.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
carterkozak: It's the least commonly used of the three by a large margin, so I've opted not to include instrumentation in this change. I'll send you our encoding metrics internally

schlosna: I'm fine deferring CBOR for now, we create those outside of Conjure anyway

schlosna (approved):

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/InstrumentedJsonFactory.java
cowtowncoder: FWTW String value length limitations should apply for case of JSON (since physically they are JSON Strings).

But no limit checking for native Binary for Smile or CBOR etc. Could add limits in future, but if so those would be handled separately.

carterkozak: Thanks for the heads up!

In most places these are used we have some form of protocol negotiation and expect smile and json to behave equivalently, so the data should still provide valid signal, but I will differentiate the metrics based on the format in a follow-up

PR #2574, opened 2023-03-28
Increase string-length observability thresholds

carterkozak:
This is meant to increase the signal to noise ratio of the data we collect in a couple ways:
We collect stack trace information when values are within 20% of the upcoming limit in Jackson (4mb), helping us focus our efforts on places that are most likely to be impacted.
We increase the minimum threshold to update our histogram in order to vastly reduce interactions with the metric because it uses a recency-biased random sampling reservoir which holds ~1k samples at a time. With a high rate of interactions, it's likely that we will fail to sample the largest value's we observe (which I've found in practice as logged warnings have length values which exceed reported metrics).

==COMMIT_MSG==
Increase string-length observability thresholds
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
+++ b/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
@@ -323,8 +323,8 @@ public final class ObjectMappersTest {
         registry.forEachMetric((name, _value) -> registry.remove(name));
         Histogram stringLength = JsonParserMetrics.of(registry).stringLength(JsonFactory.FORMAT_NAME_JSON);
         assertThat(stringLength.getSnapshot().size()).isZero();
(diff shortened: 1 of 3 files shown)


conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -352,7 +352,7 @@ public void testStringMetrics_smile() throws IOException {
         Histogram stringLength = JsonParserMetrics.of(registry).stringLength(SmileFactory.FORMAT_NAME_SMILE);
         assertThat(stringLength.getSnapshot().size()).isZero();
         // Length must exceed the minimum threshold for metrics (64 characters)
```
carterkozak:
```suggestion
        // Length must exceed the minimum threshold for metrics
```

schlosna (approved):

PR #2580, opened 2023-04-03
Upgrade Undertow to 2.2.24.Final

carterkozak:
==COMMIT_MSG==
Upgrade Undertow to 2.2.24.Final
==COMMIT_MSG==

## Possible downsides?
Oddness due to interactions between the undertow javax/jakarta migration scheme and the way gradle-consistent-versions works. We test with an older undertow-servlet release, which is dangerous, but a reasonable risk in test-only code. Previously we sat on an old release.


versions.props
```diff
@@ -22,7 +22,13 @@ org.jmock:jmock = 2.12.0
...
 org.slf4j:* = 1.7.36
 io.leangen.geantyref:geantyref = 1.3.4
-io.undertow:* = 2.2.20.Final
+io.undertow:* = 2.2.24.Final
+# dependency-upgrader:OFF
+# Last version of undertow with both Jakarta+Javax support before 'undertow-servlet' was
+# updated to jakarta only. This should NEVER be done in production code, it is only used
+# here to enable existing test coverage.
+io.undertow:undertow-servlet-jakarta = 2.2.20.Final
+# dependency-upgrader:ON
```
schlosna, line 31: @carterkozak should we also bump [`com.palantir.conjure.java` to 7.5.0](https://github.com/palantir/conjure-java/releases/tag/7.5.0) in this release? I'm also not sure why we have separate pin for `com.palantir.conjure.java:conjure-lib = 7.4.0`

```
com.palantir.conjure.java:* = 7.5.0
```

carterkozak: Good call on the unnecessary `conjure-lib`, I'm not sure where that came from. I don't think conjure-java 7.5.0 is published yet

schlosna: yeah, I got 404s locally for `com.palantir.conjure.java:conjure-lib:7.5.0` just a second ago. Will +1 and defer to you if you want to wait for the publish to complete before merging this & cutting CJR release as well

schlosna (approved):

versions.props
carterkozak: Removing the duplication causes consistent CI failures. I'm going to undo that piece and investigate later on.

versions.props
```diff
@@ -4,8 +4,8 @@ com.google.code.findbugs:jsr305 = 3.0.2
 com.google.guava:guava = 31.1-jre
 com.netflix.concurrency-limits:* = 0.2.2
 com.netflix.feign:feign-*= 8.18.0
-com.palantir.conjure.java:* = 7.4.0
-com.palantir.conjure.java:conjure-lib = 7.4.0
+com.palantir.conjure.java:* = 7.5.0
+om.palantir.conjure.java:conjure-lib = 7.5.0
```
schlosna: sorry about the wild goose chase on this, something seems odd with `:conjure-java-client-verifier` requiring this pin, but fine leaving this extra pin here


```suggestion
com.palantir.conjure.java:conjure-lib = 7.5.0
```

carterkozak: 👍 I'll follow up on this one separately

schlosna: yeah, not urgent, just seemed odd when looking through pins, but here be 🐉

PR #2586, opened 2023-04-05
Add public JsonFactory factories to ObjectMappers

carterkozak:
==COMMIT_MSG==
Add public JsonFactory factories to ObjectMappers
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
@@ -17,12 +17,14 @@
 package com.palantir.conjure.java.serialization;
 
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.TSFBuilder;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.cfg.MapperBuilder;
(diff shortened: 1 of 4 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -221,18 +223,26 @@ public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
...
+    public static CBORFactory cborFactory() {
+        return withDefaults(CBORFactory.builder()).build();
+    }
+
+    /** Configures provided JsonFactory with Conjure default settings. */
+    private static <F extends JsonFactory, B extends TSFBuilder<F, B>> B withDefaults(B builder) {
+        return builder
+                // Interning introduces excessive contention https://github.com/FasterXML/jackson-core/issues/946
+                .disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
+    }
```
carterkozak, line 247: Unclear if we want this to be public like `withDefaultModules(ObjectMapper)`/etc. I opted to keep it private to push consumers toward concrete factories where we can add instrumentation.

schlosna: Based on some other usages, making this public seems beneficial to consolidate functionality

schlosna (approved):
Looks good, defer to you if/when you want to make `withDefaults` public for external consumption

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
carterkozak: Let's revisit making this public later on since the string length metrics from our concrete factories are valuable to capture in the short term, and using `withDefaults` will bypass them.

PR #2587, opened 2023-04-06
Use a TypeFactory backed by a Caffeine Cache

carterkozak:
This is meant to avoid contention observed in the Jackson LRUMap while we investigate and build a reproducer to submit upstream.

==COMMIT_MSG==
Use a TypeFactory backed by a Caffeine Cache
==COMMIT_MSG==

Note that in addition to using a caffeine cache, we set the bounds larger than the default jackson values.
At first glance it appears that we're moving from the default singleton TypeFactory to a new factory per-mapper, however this wasn't the case to begin with due to modules (guava) adding a `TypeModifier` which produces a new factory+cache entirely.
Also note that the SerializerCache and DeserailizerCache use the same LRUMap cache as TypeFactory, which I haven't found an easy way to modify. It's possible that places with contention on TypeFactory will shift contention to ser/de factories (although I suspect this is less likely as more types are visited by the TypeFactory than have serializers or deserializers).

## Possible downsides?
More to maintain on our end, we don't have a way to validate this configuration.

### DIFF
--- a/conjure-java-jackson-serialization/build.gradle
+++ b/conjure-java-jackson-serialization/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     api "com.palantir.safe-logging:preconditions"
     implementation "com.palantir.safe-logging:logger"
     implementation 'com.palantir.tritium:tritium-registry'
+    implementation 'com.github.ben-manes.caffeine:caffeine'
 
     testImplementation "junit:junit"
     testImplementation "org.assertj:assertj-core"
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CaffeineCachingTypeFactory.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CaffeineCachingTypeFactory.java
@@ -1,0 +1,145 @@
+/*
+ * (c) Copyright 2023 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.type.TypeFactory;
+import com.fasterxml.jackson.databind.type.TypeModifier;
+import com.fasterxml.jackson.databind.type.TypeParser;
+import com.fasterxml.jackson.databind.util.ArrayBuilders;
+import com.fasterxml.jackson.databind.util.LRUMap;
+import com.fasterxml.jackson.databind.util.LookupCache;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.primitives.Ints;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.logger.SafeLogger;
+import com.palantir.logsafe.logger.SafeLoggerFactory;
+import javax.annotation.Nullable;
+
+final class CaffeineCachingTypeFactory extends TypeFactory {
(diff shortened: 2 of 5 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CaffeineCachingTypeFactory.java
```diff
@@ -0,0 +1,145 @@
...
+    public CaffeineCachingTypeFactory withCache(LRUMap<Object, JavaType> cache) {
+        return withCache((LookupCache<Object, JavaType>) cache);
+    }
+
+    @Override
+    public CaffeineCachingTypeFactory withCache(LookupCache<Object, JavaType> cache) {
+        return new CaffeineCachingTypeFactory(asCaffeine(cache), _parser, _modifiers, _classLoader);
+    }
+
+    private static final class CaffeineLookupCache implements LookupCache<Object, JavaType> {
```
carterkozak, line 102: Initially I opted to implement only the cache component, however we must extend `TypeFactory` so that calls to `withModifier(modifier)`don't drop our cache in favor of the default implementation.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CaffeineCachingTypeFactory.java
```diff
@@ -0,0 +1,145 @@
...
+        private final Cache<Object, JavaType> cache;
+
+        CaffeineLookupCache() {
+            this.cache = Caffeine.newBuilder()
+                    // max-size 1000 up from 200 default as of 2.14.2
+                    .maximumSize(1000)
+                    // initial-size 128 up from 16 default as of 2.14.2
+                    .initialCapacity(128)
```
carterkozak, line 110: I'd really like to add a max-age here to allow eventual class unloading, but it's not clear what a good value would be.

schlosna: Should we use a `expireAfterAccess` of relatively short duration?

conjure-java-jackson-serialization/build.gradle
```diff
@@ -14,6 +14,7 @@ dependencies {
     api "com.palantir.safe-logging:preconditions"
     implementation "com.palantir.safe-logging:logger"
     implementation 'com.palantir.tritium:tritium-registry'
+    implementation 'com.github.ben-manes.caffeine:caffeine'
```
schlosna, line 17: I think we're probably ok to add this as a dependency given it is already used in many places. If/when there are any major version bumps/API changes those may be tougher to roll out.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/CaffeineCachingTypeFactory.java
carterkozak: That's what I was had in mind, sorry, not sure why I said 'max-age' 🤦

The caffeine cache implementation appears equivalent to the cache used in jackson, where contention occurs within ConcurrentHashMaps, so my current theory is that we have a hot rotating set of types beyond the 200 element size of the cache. Since endpoints are called frequently, we end up churning the same items in and out. Each time an item is expired, several threads attempt to recompute and re-add it to the cache in parallel.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -204,6 +205,7 @@ public static <M extends ObjectMapper, B extends MapperBuilder<M, B>> B withDefa
      * </ul>
      */
     public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
+        mapper.setTypeFactory(new CaffeineCachingTypeFactory());
         return mapper.registerModule(new GuavaModule())
```
schlosna, line 209: format nit:

```suggestion
        return mapper.setTypeFactory(new CaffeineCachingTypeFactory())
                .registerModule(new GuavaModule())
```

schlosna (approved):
Thanks for this @carterkozak ! I'm not sure exactly where the right place for this would be, but I'm wondering if we should add some serDe benchmarks for various shapes of data

schlosna:
Hmm, I thought I had removed `merge when ready` label, but that didn't seem to happen

schlosna:
....and there it goes

carterkozak:
> I'm not sure exactly where the right place for this would be, but I'm wondering if we should add some serDe benchmarks for various shapes of data

Completely agree this would be helpful. Perhaps internally where we can sample large realistic examples.

I'm going to push a PR shortly adding instrumentation to this so we have some way to verify assumptions described in https://github.com/palantir/conjure-java-runtime/pull/2587#discussion_r1160203511

PR #2588, opened 2023-04-06
Add TypeFactory cache instrumentation

carterkozak:
==COMMIT_MSG==
Add TypeFactory cache instrumentation
==COMMIT_MSG==


conjure-java-jackson-serialization/src/main/metrics/jackson-registry-metrics.yml
```diff
@@ -1,12 +0,0 @@
...
-  javaVisibility: packagePrivate
-namespaces:
-  json.parser:
-    docs: Metrics produced instrumented Jackson components.
-    metrics:
-      string.length:
-        type: histogram
-        tags:
-          - format
-        docs: Histogram describing the length of strings parsed from input.
```
carterkozak, line 12: This was a rename to `jackson-metrics` since `-registry` didn't make a lot of sense in the filename.

schlosna (approved):

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -421,6 +423,39 @@ private void testTypeFactory(ObjectMapper mapper) {
...
+        ObjectMapper mapper = ObjectMappers.newServerJsonMapper();
+        long hitBefore = hit.getCount();
+        long missBefore = miss.getCount();
+
+        mapper.writeValue(ByteStreams.nullOutputStream(), new SimpleSerializable());
+
+        long hits = hit.getCount() - hitBefore;
+        long misses = miss.getCount() - missBefore;
+
+        assertThat(hits).isGreaterThan(0);
```
schlosna: should this be 0 hits?

```suggestion
        assertThat(hits).isZero();
```

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
```diff
@@ -421,6 +423,39 @@ private void testTypeFactory(ObjectMapper mapper) {
...
+        assertThat(misses).isOne();
+
+        // After writing the same type again, we should observe no additional misses.
+        missBefore = miss.getCount();
+        mapper.writeValue(ByteStreams.nullOutputStream(), new SimpleSerializable());
+        misses = miss.getCount() - missBefore;
+        assertThat(misses).isZero();
```
schlosna: should we check that hits > 0 here?

```suggestion
        assertThat(hit.getCount() - hits).isGreaterThan(0);
```

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/ObjectMappersTest.java
carterkozak: Internally this ends up around 3 due to various components mapping the same type. It's probably best not to make any asserts at all at this point, then assert the value has increased in the second round

carterkozak: I'll update the test to operate directly on the TypeFactory since the ObjectMapper complicates things

PR #2595, opened 2023-04-12
Opt out of the TypeFactory cache until we can resolve cache contention

carterkozak:
## Before this PR

https://github.com/FasterXML/jackson-databind/issues/3876
https://github.com/FasterXML/jackson-benchmarks/pull/5

The linked benchmarks show a small advantage of caching over not caching in most cases, in the best case for a cache: when a single operation is repeatedly executed. In practice, the cache is used for all operations and sees heavy evictions, where contention is more likely and has greater impact on users.

Given the hash collision issues that we're aware of, and the impacts we've observed, we will remove the cache until a version is available which isn't impacted by the linked issue for performance that's more predictable.

## After this PR

==COMMIT_MSG==
Opt out of the TypeFactory cache until we can resolve cache contention
==COMMIT_MSG==

## Possible downsides?
No caching may result in worse performance in the best cases, but avoids cross-thread contention on overloaded buckets which result in exceedingly long operations.


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/NonCachingTypeFactory.java
```diff
@@ -0,0 +1,110 @@
...
+        super(NoCacheLookupCache.INSTANCE);
+    }
+
+    NonCachingTypeFactory(TypeParser parser, @Nullable TypeModifier[] modifiers, ClassLoader classLoader) {
+        super(NoCacheLookupCache.INSTANCE, parser, modifiers, classLoader);
+    }
+
+    @Override
+    public NonCachingTypeFactory withModifier(TypeModifier mod) {
+        return new NonCachingTypeFactory(_parser, computeModifiers(_modifiers, mod), _classLoader);
```
schlosna, line 45: Meta-non-actionable observation: the `_modifiers` and `_classloader` prefixed with `_` threw me at first as I forgot that's the Jackson protected field style.

schlosna (approved):

schlosna:
Based on the benchmarks and profiles I have seen, this tradeoff seems right for now to avoid the major contention and ConcurrentHashMap bin traversals on high throughput code paths.

PR #2597, opened 2023-04-14
ObjectMappers.withDefaultModules(ObjectMapper) does not replace existing TypeFactories

carterkozak:
Setting the TypeFactory does not cause the new TypeFactory to be updated with existing TypeModifier instances, even if modules are re-registered (due to IGNORE_DUPLICATE_MODULE_REGISTRATIONS). We can identify cases where the default TypeFactory instance is used, and handle those correctly, leaving edge cases using existing TypeFactory instances.

==COMMIT_MSG==
ObjectMappers.withDefaultModules(ObjectMapper) does not replace existing TypeFactories with state from registered modules
==COMMIT_MSG==

This resolves an issue impacting components which call `withDefaultModules(mapper)` on mappers that have been created and configured elsewhere prior to being passed to this method.

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
@@ -170,7 +170,7 @@ public final class ObjectMappers {
      * </ul>
      */
     public static <M extends ObjectMapper, B extends MapperBuilder<M, B>> B withDefaultModules(B builder) {
-        return builder.typeFactory(new NonCachingTypeFactory())
+        return builder.typeFactory(NonCachingTypeFactory.from(builder.build().getTypeFactory()))
                 .addModule(new GuavaModule())
                 .addModule(new ShimJdk7Module())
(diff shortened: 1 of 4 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -205,7 +205,7 @@ public static <M extends ObjectMapper, B extends MapperBuilder<M, B>> B withDefa
      * </ul>
      */
     public static ObjectMapper withDefaultModules(ObjectMapper mapper) {
-        return mapper.setTypeFactory(new NonCachingTypeFactory())
+        return mapper.setTypeFactory(NonCachingTypeFactory.from(mapper.getTypeFactory()))
```
carterkozak, line 208: I considered unsetting `IGNORE_DUPLICATE_MODULE_REGISTRATIONS` and always using `NonCachingTypeFactory`, however that's unsound if modules that CJR doesn't apply which add TypeModifier instances have already been registered.

The most common case is that the `new*Mapper` methods are used initially, so I think what I've done here is the most reasonable trade-off.

mpritham (approved):

PR #2601, opened 2023-04-19
Apply an excessively permissive large string length limit by default

carterkozak:
This allows us to bridge the difference between jackson 2.14 and 2.15 such that we are able to safely upgrade, even if the upgrade occurs before CJR is prepared to take the upgrade itself.

We will ratchet down the default over time.

==COMMIT_MSG==
Apply a large StreamReadConstraints maxStringLength to reduce friction in preparation for jackson 2.15 adoption
==COMMIT_MSG==

## Possible downsides?
This is a very large value, nearly equivalent to no limit. However, we limit the total request size for structured data already.

Testing: It's difficult to test this change because we do not want to add a dependency on 2.15.0-rc3+, however I have verified that the new test fails without this patch, and passes once the patch has been applied.

### DIFF
--- a/changelog/@unreleased/pr-2601.v2.yml
+++ b/changelog/@unreleased/pr-2601.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Apply a large StreamReadConstraints maxStringLength to reduce friction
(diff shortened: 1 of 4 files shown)


bjlaub (approved):
lgtm

PR #2603, opened 2023-04-20
Opt out of canonicalization in Conjure JsonFactories

carterkozak:
We expect unlimited keyspace due to maps keyed by random IDs which are usually UUID-based. This interacts poorly with string canonicalization components in Jackson in ways that cause heavy heap churn. See the discussion here for the details:
https://github.com/FasterXML/jackson-benchmarks/pull/6

==COMMIT_MSG==
Opt out of canonicalization in Conjure JsonFactories
==COMMIT_MSG==

## Possible downsides?
The primary risk is that without canonicalization, we allocate reused strings much more heavily. Strings used to match setter methods should be very short lived and should not escape the parsing thread, so the JIT may be able to do terribly clever things to help us out.

Note: it may be possible to canonicalize contextually in jackson 3, which would give us the best of both worlds!

### DIFF
--- a/changelog/@unreleased/pr-2603.v2.yml
+++ b/changelog/@unreleased/pr-2603.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Opt out of canonicalization in Conjure JsonFactories
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/2603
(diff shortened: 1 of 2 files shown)


schlosna (approved):

schlosna:
Example snapshot from `com.fasterxml.jackson.perf.json.JsonArbitraryFieldNameBenchmark.parse-AverageTime-mode-NO_INTERN-shape-RANDOM_KEY_MAP-type-INPUT_STREAM/profile.jfr` shows significant `Arrays.copyOf` via `void com.fasterxml.jackson.core.sym.ByteQuadsCanonicalizer._verifySharing():1005` which we have seen in Conure based services.

image

PR #2608, opened 2023-04-24
Improve ClientConfiguration creation performance by reusing already created trust managers

jpcorreia99:
## Before this PR
In the creation of the client configuration from the service configuration we currently first create a SSLSocketFactory and later a X509TrustManager. 

This is redundant work since the [creation of the SSLSocketFactory](https://github.com/palantir/conjure-java-runtime/blob/d4ece2245a9a61f013b0e37c01c9cb2e3897651b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java#L51) already has to [instantiate all the necessary TrustManagers](https://github.com/palantir/conjure-java-runtime/blob/d4ece2245a9a61f013b0e37c01c9cb2e3897651b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java#L98-L102).

Since the creation of the managers involves an I/O call, removing the second creation creates a verifiable performance improvement. 

## After this PR
==COMMIT_MSG==
Extract creation of X509TrustManager for reutilization on client config creation.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-2608.v2.yml
+++ b/changelog/@unreleased/pr-2608.v2.yml
@@ -1,0 +1,6 @@
+type: improvement
+improvement:
+  description: Extract creation of X509TrustManager for reutilization on client config
+    creation.
+  links:
+  - https://github.com/palantir/conjure-java-runtime/pull/2608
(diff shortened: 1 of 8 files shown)


jpcorreia99:
After discussion, a better abstracted approach was presented in https://github.com/palantir/conjure-java-runtime/pull/2609

carterkozak:
Let's close this and merge #2609

PR #2609, opened 2023-04-24
SslSocketFactories avoids duplicate effort creating TLS components

carterkozak:
Concretely, this simplifies creation of TrustContext and creating ClientConfiguration from a ServiceConfiguration to load the TrustManagerFactory from disk a single time rather than once for the X509TrustManager and a second time for the SSLContext to create an SSLSocketFactory.

Additionally, we provide public utility functionality such that consumers may also avoid duplicating efforts with slight code changes, without reimplementing the logic to create SSLContext instances themselves.

## After this PR

==COMMIT_MSG==
SslSocketFactories avoids duplicate effort creating TLS components
==COMMIT_MSG==

### DIFF
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
@@ -26,8 +26,10 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.Provider;
+import java.security.SecureRandom;
 import java.util.Map;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import javax.net.ssl.KeyManager;
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLContext;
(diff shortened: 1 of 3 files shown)


keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
```diff
@@ -194,44 +204,48 @@ public static TrustManager[] createTrustManagers(Map<String, PemX509Certificate>
      * {@link #createSslSocketFactory}.
```
mpritham, line 204: nit: Maybe we want to update this comment? `createX509TrustManager` seems like it does something similar to this method, but I don't think there's a version of `createSslSocketFactory` which constructs an `sslContext` using key managers.

mpritham (approved):
lgtm. took off auto-merge tag.

keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
carterkozak: `createSslSocketFactory` does so here: https://github.com/palantir/conjure-java-runtime/blob/989486d6b7162c35f50d84e5dc259ae8c4573b3f/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java#L47-L56
calling into this function:
https://github.com/palantir/conjure-java-runtime/blob/989486d6b7162c35f50d84e5dc259ae8c4573b3f/keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java#L94-L105

I think the comment is still the simplest way to describe the functionality, even if the implementation avoids some overhead that would be incurred by calling each of those utilities individually. What do you think?

mpritham: I see, didn't go into the `createSslContext` function. Ok, that sounds good. 👍

PR #2611, opened 2023-04-25
Client configuration build accepts TrustContextFactory 

jpcorreia99:
Currently, if consumers of this library want to build a client configuration but also be able to provide their own SSLSocketFactory and TrustManagers, the only option they have to do so, is via [this call](https://github.com/palantir/conjure-java-runtime/blob/f1e0ddb4f8a13886a9a024ff36047a6c7d230a95/client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java#L111-L113) which does not allow the consumers to provide any extra configuration.

This PR overloads the `of` method to allow for a `TrustContextFactory` to be passed alongside the normal `ServiceConfiguration`, letting the caller have control both over SSL related classes and configuration.

Also overloads the `createTrustContext` methods to allow a security `Provider` class to be passed.

## After this PR
==COMMIT_MSG==
Overload client config creation method to allow a TrustContextFactory to be passed as well.
==COMMIT_MSG==

## Possible downsides?
The factory may get abused and provide SSL related classes that do not exactly match the security settings in the provided config.


keystores/src/main/java/com/palantir/conjure/java/config/ssl/TrustContextFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.config.ssl;
+
+import com.palantir.conjure.java.api.config.ssl.SslConfiguration;
+
+public interface TrustContextFactory {
+    TrustContext create(SslConfiguration configuration);
+}
```
carterkozak: This could use some javadoc. I might suggest moving it into `ClientConfigurations` as a nested class since it's meant specifically for use with that class.

keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
```diff
@@ -210,6 +210,20 @@ public static TrustContext createTrustContext(SslConfiguration config) {
...
+     *
+     * @param config an {@link SslConfiguration} describing the trust store and key store configuration
+     * @param provider The preferred security {@link Provider}
+     */
+    public static TrustContext createTrustContext(SslConfiguration config, Provider provider) {
+        TrustManager[] trustManagers = createTrustManagers(config);
+        KeyManager[] keyManagers = createKeyManagers(config);
+        SSLContext context = createSslContext(trustManagers, keyManagers, provider);
+        return TrustContext.of(context.getSocketFactory(), getX509TrustManager(trustManagers));
+    }
```
carterkozak: I suspect we won't actually use the new overload methods, let's undo this part for this PR

jpcorreia99: handled in dc3995258d697797736c97f45a18a24c2da4fae8

keystores/src/main/java/com/palantir/conjure/java/config/ssl/TrustContextFactory.java
jpcorreia99: Handled in https://github.com/palantir/conjure-java-runtime/pull/2611/commits/a543cf6fce5a665a0c1eaae42cc3f43943c49ae2

carterkozak (approved):

PR #2677, opened 2023-08-14
Add method for creating an SSLContext that merges both the trust store and custom certificates

Eric-Alvarez:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
@@ -122,6 +122,12 @@ final class KeyStores {
         KeyStore keyStore;
         keyStore = createKeyStore();
 
+        addCertificatesToKeystore(certificatesByAlias, keyStore);
+
+        return keyStore;
+    }
+
+    static void addCertificatesToKeystore(Map<String, PemX509Certificate> certificatesByAlias, KeyStore keyStore) {
         for (Map.Entry<String, PemX509Certificate> entry : certificatesByAlias.entrySet()) {
             try (InputStream certIn =
                     new ByteArrayInputStream(entry.getValue().pemCertificate().getBytes(StandardCharsets.UTF_8))) {
@@ -135,8 +141,6 @@ final class KeyStores {
                         e);
             }
         }
-
-        return keyStore;
     }
 
     private static void addCertificatesToKeystore(
--- a/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
+++ b/keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
@@ -20,7 +20,9 @@ import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.base.Throwables;
 import com.google.common.io.BaseEncoding;
+import com.palantir.conjure.java.api.config.ssl.SslConfiguration.StoreType;
 import com.palantir.conjure.java.config.ssl.pkcs1.Pkcs1PrivateKeyReader;
+import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeRuntimeException;
 import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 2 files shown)


keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
```diff
@@ -91,6 +92,20 @@ public static SSLSocketFactory createSslSocketFactory(
...
 
+    /**
+     * Create a {@link SSLSocketFactory} from the provided certificates and configuration.
+     *
+     * @param trustCertificatesByAlias a map of X.509 certificate in PEM or DER format by the alias to load the
+     *     certificate as.
+     * @param config an {@link SslConfiguration} describing the trust store configuration
+     * @param provider The preferred security {@link Provider}
+     */
+    public static SSLSocketFactory createSslSocketFactory(
```
Eric-Alvarez, line 103: I noticed most of these methods aren't explicitly tested, do I need to add tests for my additions?

keystores/src/main/java/com/palantir/conjure/java/config/ssl/SslSocketFactories.java
```diff
@@ -269,6 +311,23 @@ public static KeyManager[] createKeyManagers(SslConfiguration config) {
...
+            TrustManagerFactory trustManagerFactory =
+                    TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            trustManagerFactory.init(keyStore);
+            return trustManagerFactory;
+        } catch (GeneralSecurityException e) {
+            throw Throwables.propagate(e);
+        }
+    }
+
+    private static KeyStore getCombinedTrustStoreAndDefaultCas(Path trustStorePath, StoreType trustStoreType) {
```
Eric-Alvarez, line 331: This would be convenient if we made it public, this PR would not be necessary if this method was available.

pkoenig10 (dismissed):
I would prefer not to add these methods.

The use case seems very niche and they introduce a fairly large footgun. I don't want to implicitly endorse adding custom certificates to truststore. Doing so makes it very hard to reason about the security of our services because can no longer be sure that the truststore being used by the application matches what is rendered by our production infrastructure.

For your use case, I'd recommend just writing this yourself.

Eric-Alvarez:
@pkoenig10 I'm happy to not add the top level methods, but exposing the ability to load a keystore from a truststore with default ca certs is innocuous and very useful for handling these edge cases.

Eric-Alvarez:
@pkoenig10 Could you take a look at the update? Making this method accessible will eliminate the need to copy paste a lot of code for loading truststores and default ca certs (although the latter can be handled with adding the jvm_certs into the trust store). I'd rather not copy paste the method for loading truststores as it will become out of date and might fail in the future.

keystores/src/main/java/com/palantir/conjure/java/config/ssl/KeyStores.java
```diff
@@ -317,6 +323,48 @@ static KeyStore newKeyStoreWithEntry(KeyStore original, Optional<String> passwor
...
 
+    /**
+     * Return a new {@link KeyStore} that contains the contents of the trust store and all default ca certificates.
+     *
+     * @param trustStorePath The path to the trust store.
+     * @param trustStoreType The type of trust store.
+     * @return a newly constructed key store of the type trustStoreType that contains the contents of the trust store
+     * and all default ca certificates.
+     */
+    public static KeyStore getCombinedTrustStoreAndDefaultCas(Path trustStorePath, StoreType trustStoreType) {
```
Eric-Alvarez: Unsure if it would be preferred to put this in a different class rather than making this class public.

pkoenig10:
I'd like to better understand why we think this behavior should be provided by this library. This library is intended to be a utility to create Conjure clients, not a general keystore library.

APIs live forever and can be (ab)used in unexpected ways, so I generally hold a pretty high bar when adding new APIs. I'm not convinced that this behavior crosses that threshold. This behavior is fairly niche and I don't think it's unreasonable to require consumers who want this behavior to implement it themselves.

Addressing some specific comments:

> Making this method accessible will eliminate the need to copy paste a lot of code for loading truststores and default ca certs

What is the problem with duplicating this code? Duplicating code is generally preferable to providing API that we are unsure about or may later regret.

> I'd rather not copy paste the method for loading truststores as it will become out of date and might fail in the future.

Can you explain what you mean by "out of date"? What specific failure modes are you concerned about?

Eric-Alvarez:
Security vulnerabilities, performance improvements, and adding/removing keystore types are all areas that could be a concern with duplicating the code. 
It seems this library _is_ a generic trust store/key store library for the most part. There's no mention of conjure outside of package names in the whole thing. And the return types are generic java types rather than anything specific to conjure clients.

We have a standard truststore format, it would be nice if we were able to load it without copy pasting code from another repo that could break or become insecure. Loading a truststore or keystore seems like a pretty normal use-case that could be re-used in many places. Taking a hard line that we're not allowed to iterate on or use any keystore utilities and everything needs to be copy pasted seems like an objectively worse state of the world.
I'm not sure I see the risk of abuse with simply providing a convenience method to return a loaded keystore, as you mentioned I am completely able to do this already, just in a worse way that could add bugs or be insecure.

PR #2701, opened 2023-10-13
Remove the ForwardingOkHttpClient

bmarcaur:
## Before this PR
`ForwardingOkHttpClient` is incompatible with 4.0.0 because all of the fields are made final. This removes the abstraction (as it was not used, more info on that below) and has `RemotingOkHttpClient` simply override the `newCall` method that it was concerned about.

## After this PR
We should be able to upgrade our version of OkHttp.

## Possible downsides?
Subtle behavior change? Very unlikely that this breaks anyones usage of `RemotingOkHttpClient`

## Info on usage
[Public usage of `ForwardingOkHttpClient`](https://github.com/search?q=ForwardingOkHttpClient++NOT+is%3Afork&type=code)
[Public usage of `RemotingOkHttpClient`](https://github.com/search?q=RemotingOkHttpClient++NOT+is%3Afork&type=code)

[Internal usage of `ForwardingOkHttpClient`](https://pl.ntr/internal-forwarding-v2)
[Internal usage of `RemotingOkHttpClient`](https://pl.ntr/internal-remoting)

As you can see from all of the links, the only usages are internal to this repo (or internal forks) making this ABI change a strictly internal concern.

There are currently [258 internal repos](https://pl.ntr/affected-okhttp-clients) that have both a version 4.x+ of `OkHttp` and `okhttp-clients` on their classpath, i.e. have the potential for a runtime error should they ever call `OkHttpClients#create`.


okhttp-clients/src/main/java/okhttp3/OpeningOkHttpClient.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// CHECKSTYLE:OFF
+package okhttp3;
+// CHECKSTYLE:ON
+
+// Opens up the constructor, this is made public in later versions of OkHttp
+public class OpeningOkHttpClient extends OkHttpClient {
```
bmarcaur, line 21: I tried a couple ways of trying to have `RemotingOkHttpClient` masquerade as a regular `OkHttpClient` but without internal access to the constructor and a closed garden builder API I was not able to do so.

This is also a temporary opening because as soon as we upgrade to 4.X the constructor is made public, but I wanted to separate that concern.

bmarcaur:
Gonna make this OBE by investing in https://github.com/palantir/conjure-java-runtime/pull/1853

PR #2709, opened 2023-10-20
Bump to OkHttp 4.X

### LINKED ISSUE #1933: Not compatible with okhttp 4.x
## What happened?

We had a repo where `com.squareup.okhttp3:okhttp:4.9.0` was in use due to a dependency on `org.cometd.java:cometd-java-client-websocket-okhttp:5.0.2`. This resulted in the following runtime error:

```
java.lang.VerifyError: class com.palantir.conjure.java.okhttp.ForwardingOkHttpClient overrides final method cookieJar.()Lokhttp3/CookieJar;
```

Per https://square.github.io/okhttp/upgrading_to_okhttp_4/#backwards-incompatible-changes:

> OkHttpClient has 26 accessors like interceptors() and writeTimeoutMillis() that were non-final in OkHttp 3.x and are final in 4.x. These were made non-final for use with mocking frameworks like Mockito. We believe subtyping OkHttpClient is the wrong way to test with OkHttp. If you must, mock Call.Factory which is the interface that OkHttpClient implements.

This obviously doesn't cover what the [RemotingOkHttpClient](https://github.com/palantir/conjure-java-runtime/blob/7e38c2a94868090ede5ab546b13e0248d3dfcf03/okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpClient.java) is doing (as it's not test code), so I don't know if this will turn out to be hard to fix.

Currently I'm working around this by down-pinning to a 3.x version of okhttp.

Compatibility with okhttp 4.x.

bmarcaur:
## Before this PR
`ForwardingOkHttpClient` is incompatible with 4.0.0 because all of the fields are made final. This removes the abstraction (as it was not used, more info on that below) and has `RemotingOkHttpClient` simply override the `newCall` method that it was concerned about.

## After this PR
We should be able to upgrade our version of OkHttp. This should fix #1933.

## Possible downsides?
Subtle behavior change? Very unlikely that this breaks anyones usage of `RemotingOkHttpClient`

## Info on usage
[Public usage of `ForwardingOkHttpClient`](https://github.com/search?q=ForwardingOkHttpClient++NOT+is%3Afork&type=code)
[Public usage of `RemotingOkHttpClient`](https://github.com/search?q=RemotingOkHttpClient++NOT+is%3Afork&type=code)

[Internal usage of `ForwardingOkHttpClient`](https://pl.ntr/internal-forwarding-v2)
[Internal usage of `RemotingOkHttpClient`](https://pl.ntr/internal-remoting)

As you can see from all of the links, the only usages are internal to this repo (or internal forks) making this ABI change a strictly internal concern.

There are currently [258 internal repos](https://pl.ntr/affected-okhttp-clients) that have both a version 4.x+ of `OkHttp` and `okhttp-clients` on their classpath, i.e. have the potential for a runtime error should they ever call `OkHttpClients#create`.

So I had a shower thought thought, we have so many "incompatible" repos (see above) I wondered if this was hitting us in the field and we just didn't know it. So I did an [Aries search](https://apollo.palantircloud.com/workspace/aries/logList/v2/%7B%22environmentIds%22%3A%5B%5D%2C%22logTypes%22%3A%5B%22SERVICE_LOG%22%5D%2C%22queryStrings%22%3A%5B%22ForwardingOkHttpClient%22%5D%2C%22startTime%22%3A%222023-10-18T13%3A28%3A04.000%22%7D) and did not find any errors similar to the one described in #1933. Which makes sense as I imagine that would be devastating and crash most services, but also means that `okhttp-clients` is being overly broadcasted by `$something`. This PR feels like a strict improvement, but the thing it is improving _smells_.

Edit: I have spent more time burning down OkHttp internally and have two `$somethings`. `magritte-sdk` but their movement off of Okhttp/retrofit is a WIP. There is a draft PR. But the biggest offender are references to old versions of libraries i.e. stalled excavators are inhibiting flow through the system. I have internal links to back this up if you need.


okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/Okhttp39HostnameVerifier.java
```diff
@@ -79,7 +79,7 @@ public boolean verify(String host, SSLSession session) {
     }
 
     public boolean verify(String host, X509Certificate certificate) {
-        return verifyAsIpAddress(host) ? verifyIpAddress(host, certificate) : verifyHostname(host, certificate);
+        return canParseAsIpAddress(host) ? verifyIpAddress(host, certificate) : verifyHostname(host, certificate);
```
bmarcaur: [Weird rename?](https://github.com/square/okhttp/commit/42a41f3857fb8a7f93fc2e1f5a8b4c534329fbff#diff-1d7cffa64af32fd577609a790c990e9ae9728cca3b1bbfd5175adf532a320567R235)

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/RemotingOkHttpCall.java
```diff
@@ -586,7 +585,7 @@ private static void retryIfAllowed(Callback callback, Call call, Exception excep
     }
 
     private static boolean isStreamingBody(Call call) {
-        return call.request().body() instanceof UnrepeatableRequestBody;
+        return call.request().body() != null && call.request().body().isOneShot();
```
bmarcaur, line 588: [PR](https://github.com/square/okhttp/pull/4676/commits/0f624b52bebbcac83f7cf8566dc25a694b6adbed#diff-d216814317bd8f987b94647a7bb31afaac312e9c50dd7b9bf05dcce7b1f039b9R92) removing the type and introducing the new API

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/Okhttp39HostnameVerifier.java
schlosna: nit: can we remove the static import and instead `import okhttp3.internal.Util;`

```suggestion
        return Util.canParseAsIpAddress(host) ? verifyIpAddress(host, certificate) : verifyHostname(host, certificate);
```

versions.props
```diff
@@ -13,6 +13,7 @@ com.palantir.safe-logging:* = 3.6.0
 com.palantir.tokens:* = 3.18.0
 com.palantir.tracing:* = 6.18.0
 com.palantir.tritium:* = 0.79.0
+com.squareup.okhttp3:* = 4.11.0
```
schlosna: [4.12.0 is out as of 2023-10-16](https://square.github.io/okhttp/changelogs/changelog_4x/) and bumps okio to 3.6.0 to fix [CVE-2023-3635](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-3635)

```suggestion
com.squareup.okhttp3:* = 4.12.0
```

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/Okhttp39HostnameVerifier.java
```diff
@@ -32,7 +32,7 @@
  */
 package com.palantir.conjure.java.okhttp;
 
-import static okhttp3.internal.Util.verifyAsIpAddress;
+import static okhttp3.internal.Util.canParseAsIpAddress;
```
schlosna: nit & sort

```suggestion
import okhttp3.internal.Util;
```

schlosna (approved):

carterkozak:
:+1:

PR #2714, opened 2023-10-21
Replace deprecated mappingException

schlosna:
## Before this PR
Jackson deprecated `DeserializationContext#mappingException` and removes it in 2.16, leading to compilation failure in https://github.com/palantir/conjure-java-runtime/pull/2712 .

## After this PR
==COMMIT_MSG==
Replace deprecated mappingException
==COMMIT_MSG==


carterkozak (approved):
Thanks!

PR #2727, opened 2023-11-02
Enable checkImplicitDependencies

pkoenig10:
In https://github.com/palantir/conjure-java-runtime/pull/2709 we upgraded OkHttp to 4.x, but `okhttp-clients` only has an implicit dependency on OkHttp through `tracing-okhttp`. Consumers without their own explicit dependency on OkHttp 4.x were not upgraded to 4.x and experienced runtime failures.


conjure-java-jersey-server/build.gradle
```diff
@@ -51,3 +57,7 @@ configurations.all {
         exclude group: 'org.glassfish.hk2.external', module: 'jakarta.inject'
     }
 }
+
+checkImplicitDependencies {
+    ignore 'com.palantir.conjure.java.runtime', 'conjure-java-jersey-jakarta-server'
+}
```
pkoenig10, line 67: Not sure why this false positive is happening, I presume because we have duplicate classes.

```
Execution failed for task ':conjure-java-jersey-server:checkImplicitDependenciesMain'.
> Found 1 implicit dependencies - consider adding the following explicit dependencies to 'conjure-java-jersey-server/build.gradle', or avoid using classes from these jars:
      dependencies {
          implementation project(':conjure-java-jersey-jakarta-server')
      }
```

bmarcaur (approved):

conjure-java-jersey-server/build.gradle
```diff
@@ -20,15 +20,25 @@ dependencies {
...
     implementation "com.netflix.feign:feign-core"
     implementation "com.palantir.safe-logging:safe-logging"
     implementation "com.palantir.safe-logging:logger"
     implementation 'com.palantir.tokens:auth-tokens'
     implementation "com.palantir.tracing:tracing-jersey"
+    implementation 'org.glassfish.jersey.core:jersey-common', {
+        version {
+            strictly jerseyVersion
+        }
+    }
```
pkoenig10, line 41: Please double check that I didn't miss anything else for the legacy javax projects.

schlosna (approved):

PR #2740, opened 2023-11-29
Changes Necessary to Fix 8.0 Conjure Update

bmarcaur:
Fixes #2735, but will mean that `conjure-java-retrofit2-client` is untested, and while we intend to remove that it seems that the lesser evil is blocking the 8.x upgrade IMO.


carterkozak:
Should be unnecessary now that #1853 has merged

PR #2744, opened 2023-12-01
Clean up version constraints

pkoenig10:


versions.props
```diff
@@ -46,11 +48,3 @@ javax.annotation:javax.annotation-api = 1.3.2
 javax.ws.rs:javax.ws.rs-api = 2.1.1
 javax.inject:javax.inject = 1
 
-# dependency-upgrader:OFF
-
-com.fasterxml.jackson.*:* = 2.13.1
-
-# Pin to same version as spark
-io.dropwizard.metrics:metrics-core = 3.2.6
```
carterkozak, line 54: When a new dropwizard-metrics version is released, we don't want this repo to to cause everything else to resolve it upward -- that has historically caused issues for spark.

pkoenig10: But that already happens - we're currently using a 4.x version which is being bumped by transitive dependencies. So this constraint is just confusing.

carterkozak: the dw-metrics dep should ideally be produced by tritium, however at the limit we'd end up splitting out that dependency into a separate excavator if we ran into that sort of api split again

carterkozak:
:+1: you'll need to rebase in order to merge

pkoenig10:
@carterkozak not sure why your 👍 isn't satisfying policy-bot

carterkozak (approved):
I think the policy files were updated recently to invalidate on push

PR #2760, opened 2023-12-16
Enable Werror

pkoenig10:
Fix some Error Prone warnings and enforce that warnings are treated as errors going forward.

We do this in our other projects, [such as Dialogue](https://github.com/palantir/dialogue/blob/e887725af7ee847c986949c008c69b2de44fb9e0/build.gradle#L81-L83).

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/InputStreamDelegateDecoder.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/InputStreamDelegateDecoder.java
@@ -22,6 +22,7 @@ import com.palantir.logsafe.Safe;
 import com.palantir.tritium.metrics.registry.SharedTaggedMetricRegistries;
 import feign.FeignException;
 import feign.Response;
+import feign.Util;
 import feign.codec.Decoder;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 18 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/InputStreamDelegateDecoder.java
```diff
@@ -52,7 +51,7 @@ public InputStreamDelegateDecoder(@Safe String clientNameForLogging, Decoder del
     public Object decode(Response response, Type type) throws IOException, FeignException {
         if (type.equals(InputStream.class)) {
             byte[] body =
-                    response.body() != null ? Util.toByteArray(response.body().asInputStream()) : new byte[0];
+                    response.body() != null ? response.body().asInputStream().readAllBytes() : new byte[0];
```
carterkozak: Unlike `Util.toByteArray`, `InputStream.readAllBytes()` does not close the InputStream. Feign wraps calls with a try/finally+close, so this should be fine

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/InputStreamDelegateEncoder.java
```diff
@@ -52,7 +51,7 @@ public InputStreamDelegateEncoder(@Safe String clientNameForLogging, Encoder del
     public void encode(Object object, Type bodyType, RequestTemplate template) throws EncodeException {
         if (bodyType.equals(InputStream.class)) {
             try {
-                byte[] bytes = Util.toByteArray((InputStream) object);
+                byte[] bytes = ((InputStream) object).readAllBytes();
```
carterkozak: This may expose resource leaks since it no longer closes the input, we should update this to preserve existing behavior

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/InputStreamDelegateDecoder.java
pkoenig10: Ah sorry, I didn't catch that. I'll revert these changes.

carterkozak (approved):

PR #2879, opened 2024-04-10
Remove Dependency on Jackson Afterburner

carterkozak:
Afterburner is not used on supported JDKs, there's no reason to include the dependency.

==COMMIT_MSG==
Remove Dependency on Jackson Afterburner
==COMMIT_MSG==

## Possible downsides?
Slight possibility that code running on jdk-11 may observe a very slight performance regression. In practice we haven't seen much difference between registering afterburner and not.


pkoenig10 (dismissed):

carterkozak:
woops, mind giving this another +1?

tpetracca:
👍🏻

PR #2903, opened 2024-05-13
Add support for creating an HTTPS proxy

Eric-Alvarez:
Relates to https://github.com/palantir/dialogue/issues/2278
## Before this PR

No support for new ProxyType HTTPS
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==
A special Proxy instance will be created when the proxy type is HTTPS. Consumers need to check for this type to determine if the proxy should be connected to with HTTPS


client-config/src/main/java/com/palantir/conjure/java/client/config/HttpsProxy.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.client.config;
+
+import java.net.Proxy;
+import java.net.SocketAddress;
+
+public class HttpsProxy extends Proxy {
```
carterkozak: I might suggest a slight change in the public API portion of this change such that we expose a bit less to consumers. Instead of exposing a public implementation of `Proxy`, what if we added a public utility class along these lines?

```java
public final class HttpsProxies {

    public static Proxy create(InetSocketAddress address) {
        // TODO
    }

    public static boolean isHttps(Proxy proxy) {
        // TODO
    }

    private HttpsProxies() {}
    
    private static final class HttpsProxy extends Proxy {
        // TODO
    }
}
```

Eric-Alvarez: Makes sense, updated, let me know if that's what you were thinking

client-config/src/main/java/com/palantir/conjure/java/client/config/HttpsProxies.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Utility class for creating and identifying HTTPS proxy instances.
+ * <p>
+ * This class provides a way to create proxy instances that represent HTTPS connections, as there is no native support
+ * for an HTTPS proxy type in the Java {@link Proxy} class. When using a proxy created with this class, consumers should
+ * use the provided {@link #isHttps(Proxy)} method to determine if the proxy should be connected to using HTTPS.
+ */
+public final class HttpsProxies {
+
+    public static Proxy create(InetSocketAddress address, boolean https) {
```
carterkozak: I don't think we need the `boolean https` here, the utility class name already implies https: `HttpsProxies.create(address)` is clear and concise imo :-)

carterkozak: Let's add a `Preconditions.checkNotNull(address, "address is required")` to guard against nulls, since this is public API

Eric-Alvarez: Updated!

carterkozak: We still need to drop the parameter:
```suggestion
    public static Proxy create(InetSocketAddress address) {
```

Eric-Alvarez: Whoops, yes of course

carterkozak (dismissed):
lgtm!

carterkozak (approved):

PR #2973, opened 2024-09-26
Cache Feign endpoints

pkoenig10:
The ultimate goal of this PR is to allow us to revert https://github.com/palantir/dialogue/pull/1414, or to at least add an exemption for `FeignEndpoint`.

The only part of the `feign.Request` that we use in the `Endpoint` implementation is the request URL. Because we don't use Dialogue path templates and path parameters here, we can (ab)use the Dialogue path parameters to pass the request URL.

Although this may seem a little strange at first glance, it's not entirely unlike the static/dynamic split between path templates and path parameters - it's just as if the entire path template is a single path parameter.

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
@@ -31,8 +31,8 @@ import com.palantir.conjure.java.client.jaxrs.feignimpl.InputStreamDelegateDecod
 import com.palantir.conjure.java.client.jaxrs.feignimpl.InputStreamDelegateEncoder;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.Java8OptionalAwareContract;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.Java8OptionalAwareDecoder;
+import com.palantir.conjure.java.client.jaxrs.feignimpl.MethodHeaderEnrichmentContract;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.NeverReturnNullDecoder;
-import com.palantir.conjure.java.client.jaxrs.feignimpl.PathTemplateHeaderEnrichmentContract;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.SlashEncodingContract;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.TextDelegateDecoder;
 import com.palantir.conjure.java.client.jaxrs.feignimpl.TextDelegateEncoder;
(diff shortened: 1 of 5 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -341,26 +352,32 @@ public Exception decode(String _methodKey, feign.Response response) {
         }
     }
 
-    private final class FeignRequestEndpoint implements Endpoint {
-        private final Request request;
-        private final String endpoint;
-        private final HttpMethod method;
+    private final class FeignEndpoint implements Endpoint {
```
pkoenig10, line 356: I've renamed this since it no longer contains a `Request`.

This has the added benefit of being able to differentiate these endpoint implementations before and after this PR if wanted to add an exemption for `TimingEndpointChannel`.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -57,18 +57,23 @@
...
  */
 final class DialogueFeignClient implements feign.Client {
 
+    private static final String REQUEST_URL_PATH_PARAM = "request-url";
     private static final String PATH_TEMPLATE = "hr-path-template";
     private static final Splitter pathSplitter = Splitter.on('/');
     private static final Splitter querySplitter = Splitter.on('&').omitEmptyStrings();
     private static final Splitter queryValueSplitter = Splitter.on('=');
 
+    private final ConcurrentHashMap<FeignEndpointKey, EndpointChannel> endpointChannelCache = new ConcurrentHashMap<>();
```
pkoenig10: Using a `ConcurrentHashMap` instead of a Caffeine cache because I don't think we have to worry about the fact that this is unbounded. The number of entires here will be bounded by the fact that this client comes from some JAX-RS interface.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
carterkozak: I think it may be reasonable to declare an endpoint tag in dialogue to opt into caching, when the endpoint itself isn't an enum constant -- that way dialogue doesn't depend on this library.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -401,12 +418,29 @@ public String serviceName() {
...
         }
 
         @Override
         public String version() {
             return version;
         }
     }
+
+    @Value.Immutable(builder = false, prehash = true)
+    interface FeignEndpointKey {
```
carterkozak, line 432: Ideally the cache key should be more closely based on the method signature, rather than the method name. Otherwise, we'll end up with collisions for overloads that take different args (e.g. `@GET @Path("one") void ping()` and `@GET @Path("two") void ping(@QueryParam("q") String val)`.
That probably wouldn't be the end of the world, but it wouldn't match dialogue behavior either.

pkoenig10: > Otherwise, we'll end up with collisions for overloads that take different args

But there's no consequence to these collisions, right?

The `Endpoint` instances created for these two methods will behave identically, so there's no reason they can't be the same instance. All difference between these endpoints are captured in the request object and applied per request.

pkoenig10: Talked offline, there is some concern around the per-endpoint concurrency limiters.

Currently, the per-endpoint concurrently limiters have no effect on Feign clients because the `Endpoint` instances are all unique per-request. With the current implement, the concurrency limiters will be too aggressive when there are overloaded methods because overloaded methods will use the same `Endpoint` instance.

To address this we will introduce an additional enrichment header that capture `Method.toString()` and include that in the endpoint cache key.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/AbstractFeignJaxRsClientBuilder.java
```diff
@@ -180,28 +180,35 @@ public int hashCode() {
     }
 
     private static Contract createContract() {
-        return new EndpointNameHeaderEnrichmentContract(
-                new PathTemplateHeaderEnrichmentContract(new SlashEncodingContract(new Java8OptionalAwareContract(
```
pkoenig10, line 184: AFAICT, the `hr-path-template` header is, and only ever has been, used by the `OkhttpTraceInterceptor`. This was introduced in https://github.com/palantir/conjure-java-runtime/pull/577.

We no longer use OkHttp clients here and this header was an implementation detail, so I think we're good to remove `PathTemplateHeaderEnrichmentContract` and the related logic in `DialogueFeignClient`.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -88,20 +94,22 @@ final class DialogueFeignClient implements feign.Client {
     @Override
     public feign.Response execute(Request request, Request.Options _options) throws IOException {
         com.palantir.dialogue.Request.Builder builder = com.palantir.dialogue.Request.builder();
-        Optional<RequestBody> body = requestBody(request);
-        if (body.isPresent()) {
-            builder.body(body);
-            builder.putHeaderParams(HttpHeaders.CONTENT_LENGTH, Integer.toString(request.body().length));
-        }
```
pkoenig10, line 95: `RequestBody` got a `contentLength` method in https://github.com/palantir/dialogue/pull/1227, so we can move this responsibility into Dialogue.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/feignimpl/MethodHeaderEnrichmentContract.java
```diff
@@ -0,0 +1,40 @@
...
+import feign.Contract;
+import feign.MethodMetadata;
+import java.lang.reflect.Method;
+
+/**
+ * Contract to capture the method name and pass it to a feign client.
+ *
+ * This should be considered internal API and should not be depended upon.
+ */
+public final class MethodHeaderEnrichmentContract extends AbstractDelegatingContract {
```
carterkozak, line 28: can we make this package private and move it to the same package as DialogueFeignClient? We have more flexibility to modify this sort of thing if we don't expose it to consumers. A fair bit of the legacy feign implementation is visible for historical reasons, but we needn't make new things follow that pattern

pkoenig10: `AbstractDelegatingContract` is currently package-private, so we have to pick our poison. Would we rather make this public or make `AbstractDelegatingContract` public?

carterkozak: Ah, fair enough — I think it’s good as is

pkoenig10: I took at stab at this in https://github.com/palantir/conjure-java-runtime/pull/2973/commits/80e54cae7c8b17d472fb4f2910042097fa979127.

It will allow us to make specific contracts/encoders/decoders package-private at the expense of adding a couple public utility methods and having to duplicate these header name constants. Let me know which you'd prefer. I'm pretty indifferent, maybe with a slight preference for leaving this how it was.

carterkozak: Let’s keep it simple — agree that the minimal change is probably ideal

carterkozak (approved):

PR #3070, opened 2025-03-03
Add Bytes deserialization module (BinaryBytesDeserializationModule)

jackwickham:
## Before this PR
When deserializing binary data (from a smile or cbor input) as an unknown type (`Object`), jackson defaults to `byte[]`, which does not provide a reasonable `equals` implementation. This means that two objects deserialized from the same input will not be equal to each other.

## After this PR

==COMMIT_MSG==
Add Bytes deserialization module to enable equality when deserializing untyped binary data
==COMMIT_MSG==

## Possible downsides?
I had to pick up a dependency on conjure-lib, which wasn't used by this repo before (other than as a dependency of the generated code).

We aren't using this module in any CJR code, just making it available for clients to enable, because enabling it by default could break clients that introspect the deserialized data.

### DIFF
--- a/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
+++ b/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
@@ -1,5 +1,17 @@
 /*
  * (c) Copyright 2025 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 
 package com.palantir.conjure.java.serialization;
--- a/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
+++ b/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
@@ -27,21 +27,14 @@ import java.util.Base64;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public final class BinaryBytesDeserializationModuleTest {
 
-    private ObjectMapper smileMapper;
-
-    @BeforeEach
-    void before() {
-        smileMapper = ObjectMappers.newSmileClientObjectMapper();
-        smileMapper.registerModule(new BinaryBytesDeserializationModule());
-    }
-
     @Test
     void deserializesRawByteArrayToBytes() throws IOException {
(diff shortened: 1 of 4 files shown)


conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
```diff
@@ -0,0 +1,124 @@
...
+        String yaml = "testData: !!binary " + base64Value;
+
+        Object deserialized = yamlMapper.readValue(yaml, Object.class);
+        assertThat(deserialized).isInstanceOf(Map.class);
+        Map<?, ?> deserializedMap = (Map<?, ?>) deserialized;
+        assertThat(deserializedMap.get("testData")).isInstanceOf(Bytes.class).isEqualTo(Bytes.from(value));
+    }
+
+    @SuppressWarnings({"DangerousRecordArrayField", "ArrayRecordComponent"}) // Intentional
+    private record RecordContainingObject(Object obj, byte[] byteArray, Bytes bytes) {}
```
carterkozak: I think we can define this within `deserializesNestedObjectFields` for readability

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
```diff
@@ -0,0 +1,124 @@
...
+        ObjectMapper yamlMapper = YAMLMapper.builder().build();
+        yamlMapper.registerModule(new BinaryBytesDeserializationModule());
+
+        byte[] value = new byte[] {1, 2, 3};
+        String base64Value = new String(Base64.getEncoder().encode(value), StandardCharsets.UTF_8);
+        String yaml = "testData: !!binary " + base64Value;
+
+        Object deserialized = yamlMapper.readValue(yaml, Object.class);
+        assertThat(deserialized).isInstanceOf(Map.class);
+        Map<?, ?> deserializedMap = (Map<?, ?>) deserialized;
```
carterkozak: I might define a record for type safety, avoiding the cast to `Map`:
```java
private record TestData(Object testData) {}
```

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.jupiter.api.Test;
+
+public final class BinaryBytesDeserializationModuleTest {
+
+    private ObjectMapper smileMapper;
+
+    @BeforeEach
+    void before() {
+        smileMapper = ObjectMappers.newSmileClientObjectMapper();
+        smileMapper.registerModule(new BinaryBytesDeserializationModule());
```
carterkozak: Slight preference to either make this a factory method returning a smile mapper, or copy/paste it into the relevant tests.
Not a strong opinion though, It's just a little odd when a test declares fields that aren't used in every test.

conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/BinaryBytesDeserializationModuleTest.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    void deserializesRawByteArrayToBytes() throws IOException {
+        byte[] value = {1, 2, 3, 4};
+        byte[] serialized = smileMapper.writeValueAsBytes(value);
+
+        Object deserialized = smileMapper.readValue(serialized, Object.class);
+        assertThat(deserialized).isInstanceOf(Bytes.class).isEqualTo(Bytes.from(value));
+    }
```
carterkozak: Thoughts on adding a similar test using the JSON mapper? The result will be a string, not bytes, but a test should make that clear (and we definitely don't want to start coercing strings into `Bytes` objects!

bjlaub (approved):
lgtm

Issue #1582, opened 2020-04-30
Placeholders missing from log messages


gatesn:
## What happened?

I'm using several components from conjure-java-runtime in my Gradle plugin where I don't have enough control over the logging framework to configure SLS logging.

As such, I don't get any arguments in any of the log messages from CJR.

## What did you want to happen?

Conjure Java Runtime uses SLF4J, it shouldn't make assumptions about how SLF4J is configured or the context in which its jars are used.

PR #3102, opened 2025-05-02
ServiceExceptionMapper log messages use placeholders

helenyugithub:
## Before this PR
My logging framework does not have SLS logging so I don't get any log arguments
See https://github.com/palantir/conjure-java-runtime/issues/1582

## After this PR
ServiceExceptionMapper log messages use placeholders
==COMMIT_MSG==
==COMMIT_MSG==


conjure-java-jersey-server/src/main/java/com/palantir/conjure/java/server/jersey/ServiceExceptionMapper.java
```diff
@@ -42,15 +42,15 @@ public Response toResponseInner(ServiceException exception) {
         int httpStatus = exception.getErrorType().httpErrorCode();
         if (httpStatus / 100 == 4 /* client error */) {
             log.info(
-                    "Error handling request",
-                    SafeArg.of("errorInstanceId", exception.getErrorInstanceId()),
+                    "Error handling request. {}: {}",
```
helenyugithub, line 45: this matches JsonExceptionMapper

bmarcaur (approved):
Thanks!

PR #3114, opened 2025-05-12
Optimize DialogueFeignClient small response reader

schlosna:
## Before this PR

Small responses <= 8KiB would always allocate 8KiB `ByteBuffer` as `InputStreamReader` creates a `StreamDecoder` that allocates a fixed 8192 byte `ByteBuffer`. This allocation becomes a scalability bottleneck for high throughput RPCs with small responses (think something returning timestamps, locks, authorization results, etc.)

See https://github.com/openjdk/jdk/blob/4c03e5938df0a9cb10c2379af81163795dd3a086/src/java.base/share/classes/sun/nio/cs/StreamDecoder.java#L248

## After this PR

==COMMIT_MSG==
Avoid `InputStreamReader` / `HeapByteBuffer` overhead for small (less than 8KiB) inputs, see https://github.com/FasterXML/jackson-core/pull/1081 and https://github.com/FasterXML/jackson-benchmarks/pull/9#issuecomment-1685036954 for benchmarks showing between 2x and 10x speedup handling deserialization of small values.

==COMMIT_MSG==

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
@@ -50,7 +50,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
-import java.io.UnsupportedEncodingException;
+import java.io.StringReader;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Collection;
@@ -143,11 +143,7 @@ final class DialogueFeignClient implements feign.Client {
     }
 
     private static String urlDecode(String input) {
-        try {
-            return URLDecoder.decode(input, "UTF-8");
-        } catch (UnsupportedEncodingException e) {
-            throw new SafeUncheckedIoException("Failed to decode path segment", e, UnsafeArg.of("encoded", input));
-        }
+        return URLDecoder.decode(input, StandardCharsets.UTF_8);
     }
 
     private static Optional<RequestBody> requestBody(Request request) {
@@ -239,6 +235,17 @@ final class DialogueFeignClient implements feign.Client {
 
         @Override
         public Reader asReader() {
+            Integer maybeLength = length();
+            if (maybeLength != null && maybeLength < 8192) {
+                // Avoid InputStreamReader / HeapByteBuffer overhead for small (less than 8KiB) inputs,
+                // see https://github.com/FasterXML/jackson-core/pull/1081
+                try (InputStream inputStream = asInputStream()) {
+                    return new StringReader(new String(inputStream.readAllBytes(), StandardCharsets.UTF_8));
+                } catch (IOException e) {
+                    throw new SafeUncheckedIoException(
+                            "Failed to read response body", e, SafeArg.of("length", maybeLength));
+                }
+            }
             return new InputStreamReader(asInputStream(), StandardCharsets.UTF_8);
         }
 
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
@@ -248,7 +248,7 @@ final class DialogueFeignClient implements feign.Client {
                     byte[] bytes = new byte[toRead];
                     try {
                         int read = inputStream.readNBytes(bytes, 0, toRead);
-                        if (read == length) {
+                        if (read <= length) {
                             // fully read input
                             inputStream.close();
                             return new StringReader(new String(bytes, 0, read, StandardCharsets.UTF_8));
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
@@ -50,6 +50,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
+import java.io.SequenceInputStream;
 import java.io.StringReader;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -239,6 +235,17 @@ public InputStream asInputStream() {
...
+            if (maybeLength != null && maybeLength < 8192) {
+                // Avoid InputStreamReader / HeapByteBuffer overhead for small (less than 8KiB) inputs,
+                // see https://github.com/FasterXML/jackson-core/pull/1081
+                try (InputStream inputStream = asInputStream()) {
+                    return new StringReader(new String(inputStream.readAllBytes(), StandardCharsets.UTF_8));
+                } catch (IOException e) {
+                    throw new SafeUncheckedIoException(
+                            "Failed to read response body", e, SafeArg.of("length", maybeLength));
+                }
+            }
```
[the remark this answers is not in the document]
schlosna: rejected. Yes `readAllBytes()` will load the full 8KiB or less response in memory, but `InputStreamReader` creates a `StreamDecoder` that allocates a fixed 8192 byte `ByteBuffer` that we are avoiding.

https://github.com/openjdk/jdk/blob/4c03e5938df0a9cb10c2379af81163795dd3a086/src/java.base/share/classes/sun/nio/cs/StreamDecoder.java#L248

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -239,6 +235,17 @@ public InputStream asInputStream() {
 
         @Override
         public Reader asReader() {
+            Integer maybeLength = length();
+            if (maybeLength != null && maybeLength < 8192) {
+                // Avoid InputStreamReader / HeapByteBuffer overhead for small (less than 8KiB) inputs,
+                // see https://github.com/FasterXML/jackson-core/pull/1081
+                try (InputStream inputStream = asInputStream()) {
```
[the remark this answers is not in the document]
schlosna: when the response is 8KiB or less we want to fully consume the response input stream and release the response after converting to an immutable `String` & `StringReader` for consumption by Jackson deserialization

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -239,6 +235,17 @@ public InputStream asInputStream() {
...
+                    return new StringReader(new String(inputStream.readAllBytes(), StandardCharsets.UTF_8));
```
aldexis: Is there any way to allocate the string more directly? `inputStream.readAllBytes()` will allocate a `byte[]` and `new String` is going to copy that byte array into a new one, leading to twice the allocation. I'm not clear whether that is in fact more efficient than the current behavior if the length is more than 4096

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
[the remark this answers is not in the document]
aldexis: Is there a guarantee that if `asReader()` has been called, nothing will then call `asInputStream()`? I imagine it doesn't make much sense to call both, but I'm not very familiar with this API.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
aldexis: Other question: Should we guard against malicious/bad responses that return an incorrect length? (e.g. I imagine there is a corner case where the length is set to 0, which doesn't make sense)
Here we'll buffer the entire response in memory. If the content length header lied to us, we might buffer a _lot_ more than we expected, since we just read the entire thing in memory. Should we just read the first `length` bytes only?

schlosna: unfortunately there is not a more direct way to allocate the `String` without extra array copy outside of JDK.

The relevant bits of benchmark https://github.com/FasterXML/jackson-benchmarks/pull/9#issuecomment-1685041665 show that `StringReader` version is beneficial or equivalent up to 32KiB, so would prefer to just keep the 8KiB limit here

```
Benchmark                                       (mode)  (size)               (type)  Mode  Cnt   Score    Error  Units
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE     128  INPUT_STREAM_READER  avgt    3   0.475 ±  0.006  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE     128        STRING_READER  avgt    3   0.144 ±  0.002  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE     512  INPUT_STREAM_READER  avgt    3   0.786 ±  0.035  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE     512        STRING_READER  avgt    3   0.388 ±  0.096  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    1024  INPUT_STREAM_READER  avgt    3   1.105 ±  0.004  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    1024        STRING_READER  avgt    3   0.698 ±  0.004  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    2048  INPUT_STREAM_READER  avgt    3   1.537 ±  0.088  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    2048        STRING_READER  avgt    3   1.456 ±  3.178  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    4096  INPUT_STREAM_READER  avgt    3   3.047 ±  0.010  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    4096        STRING_READER  avgt    3   3.113 ±  0.040  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    7168  INPUT_STREAM_READER  avgt    3  10.650 ±  2.855  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    7168        STRING_READER  avgt    3  11.025 ± 23.020  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    8192  INPUT_STREAM_READER  avgt    3  13.048 ±  0.202  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE    8192        STRING_READER  avgt    3  12.708 ±  0.513  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE   16384  INPUT_STREAM_READER  avgt    3  32.922 ±  0.386  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE   16384        STRING_READER  avgt    3  32.588 ±  0.890  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE   32768  INPUT_STREAM_READER  avgt    3  72.577 ± 11.414  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE   32768        STRING_READER  avgt    3  72.497 ± 13.592  us/op

```

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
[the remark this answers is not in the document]
schlosna: adjusted the logic per below

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
schlosna: improved the handling here in cases where content-length is incorrect

schlosna: Also worth noting that when `InputStreamReader` is returned, `feign.jackson.JacksonDecoder#decode` must wrap it in a `BufferedReader` with a `char[1]` buffer as `InputStreamReader` doesn't support mark/reset, while `StringReader` does.

aldexis (approved):

PR #3161, opened 2025-07-14
DNM: Add support for no-proxy

mpritham:
Do not merge.

Depends on this change in[ c-j-r-api merging first.](https://github.com/palantir/conjure-java-runtime-api/pull/1388)
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -270,6 +298,59 @@ public String toString() {
...
+
+        @Override
+        public List<Proxy> select(URI uri) {
+            return shouldSkipProxy(uri) ? NO_PROXIES : proxies;
+        }
+
+        private boolean shouldSkipProxy(URI uri) {
+            String host = uri.getHost().toLowerCase(Locale.ROOT);
+            for (String noProxyHost : noProxyHosts) {
+                if (host.equals(noProxyHost) || host.endsWith("." + noProxyHost)) {
```
mpritham: I need to verify this logic. What are the expected formats of this environment variable?

[This article goes over various implementations](https://about.gitlab.com/blog/we-need-to-talk-no-proxy/#no_proxy-format)

aldexis: I think you're missing logic around the port, since the port may or may not be present in the no_proxy configuration?

I believe you also want to strip a potential leading `.` if you're going to do the check in this way (which I personally think is the best way to go about this)

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -270,6 +298,59 @@ public String toString() {
...
+            this.noProxyHosts = noProxyHosts;
+        }
+
+        @Override
+        public List<Proxy> select(URI uri) {
+            return shouldSkipProxy(uri) ? NO_PROXIES : proxies;
+        }
+
+        private boolean shouldSkipProxy(URI uri) {
+            String host = uri.getHost().toLowerCase(Locale.ROOT);
```
aldexis, line 341:
```suggestion
            // hostnames are case-insensitive for DNS resolution
            String host = uri.getHost().toLowerCase(Locale.ROOT);
```

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -163,6 +168,19 @@ private static ClientConfiguration of(
                 .build();
     }
 
+    private static ImmutableSet<String> parseNoProxyEnvironmentVar() {
+        String env = System.getenv(ENV_NO_PROXY);
+        if (env == null || env.isEmpty()) {
+            log.debug(ENV_NO_PROXY + " environment variable not set.");
```
aldexis: To match the other `FROM_ENVIRONMENT` logic, I would set this to 
```suggestion
            log.info(ENV_NO_PROXY + " environment variable not set.");
```

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -163,6 +168,19 @@ private static ClientConfiguration of(
...
+            return ImmutableSet.of();
+        }
+        return Arrays.stream(env.split(","))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .map(String::toLowerCase)
+                .collect(ImmutableSet.toImmutableSet());
```
aldexis: I think it would be useful to add an info log line indicating which domains are going to bypass the proxy

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -175,7 +193,13 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
                 }
                 log.info("Using proxy from environment variable", UnsafeArg.of("proxy", defaultEnvProxy));
                 InetSocketAddress address = createInetSocketAddress(defaultEnvProxy);
-                return fixedProxySelectorFor(new Proxy(Proxy.Type.HTTP, address));
+                ImmutableSet<String> noProxyHosts = parseNoProxyEnvironmentVar();
+                if (noProxyHosts.isEmpty()) {
+                    return fixedProxySelectorFor(new Proxy(Proxy.Type.HTTP, address));
+
```
aldexis:
```suggestion
```

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
aldexis: As a matter of testing, it might be useful to test IP addresses in the no_proxy config too (not sure if this is before or after DNS-resolution, so this might be either matching on the IP-address itself, or the DNS-resolved IP address from the hostname?)

If we don't want to support IP address at all, maybe a log line indicating we found IP addresses in the no_proxy config and these aren't supported?

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -270,6 +298,59 @@ public String toString() {
...
+            this.proxies = ImmutableList.of(proxy);
+            this.noProxyHosts = noProxyHosts;
+        }
+
+        @Override
+        public List<Proxy> select(URI uri) {
+            return shouldSkipProxy(uri) ? NO_PROXIES : proxies;
+        }
+
+        private boolean shouldSkipProxy(URI uri) {
```
aldexis, line 340: Thoughts on adding some debug/trace log lines to indicate when a proxy is used vs not? (can be useful to debug what's happening in certain networking setups)

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -270,6 +298,59 @@ public String toString() {
...
+        }
+
+        @Override
+        public String toString() {
+            return "NoProxyAwareProxySelector{proxy=" + proxies.stream().findFirst() + ", noProxyHosts=" + noProxyHosts
+                    + '}';
+        }
+
+        @Override
+        public void connectFailed(URI _uri, SocketAddress _sa, IOException _ioe) {}
```
aldexis: vnit/style: let's maybe move this method to just below the `select` or `shouldSkipProxy` methods, since it's also part of `ProxySelector` (also more consistent with `FixedProxySelector`)

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -202,6 +207,44 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
...
+    }
+
+    private static ImmutableSet<String> parseNoProxyEnvironmentVar(@Nullable String noProxyEnvVariable) {
+        if (noProxyEnvVariable == null || noProxyEnvVariable.isEmpty()) {
+            log.info(ENV_NO_PROXY + " environment variable not set.");
+            return ImmutableSet.of();
+        }
+        return Arrays.stream(noProxyEnvVariable.split(","))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
```
mpritham, line 242: I'm not stripping the leading dot. [Some implementations do while others don't](https://about.gitlab.com/blog/we-need-to-talk-no-proxy/#no_proxy-format). Opt-ing to not support stripping it for now to maintain parity with the `go` client.

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
```diff
@@ -202,6 +207,44 @@ public static ProxySelector createProxySelector(ProxyConfiguration proxyConfig)
...
+            log.info("Proxy environment variable not set, using no proxy");
+            return fixedProxySelectorFor(Proxy.NO_PROXY);
+        }
+        log.info("Using proxy from environment variable", UnsafeArg.of("proxy", defaultEnvProxy));
+        InetSocketAddress address = createInetSocketAddress(defaultEnvProxy);
+        return fixedProxySelectorFor(new Proxy(Proxy.Type.HTTP, address));
+    }
+
+    @VisibleForTesting
+    static ProxySelector proxySelectorFromEnvironmentNoProxyAware(
```
mpritham, line 222: It's tricky to set the OS level environment variables in a unit test. When I tried using JUnit Pioneer (which has an annotation for setting environment variables in JUnit tests), I ran into an error pointing to [this page](https://junit-pioneer.org/docs/environment-variables/#warnings-for-reflective-access) which says:

> The best way to prevent these warnings/errors, is to change the code under test, so this extension is no longer needed

I'm opting to use these helper methods instead.

Happy to split out this PR so that I refactor existing `FROM_ENVIRONMENT` logic first.

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
mpritham: I added the logic for handling ports. The go implementation does not strip the leading `.` (according to the Gitlab article). I was opting to follow their implementation and not strip either. No strong opinion here.

client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfigurations.java
mpritham: The URIs aren't DNS resolved. Still, it's possible for a client to supply a resolved IP when making a call, and so we should support IP addresses in the `no_proxy` list.

I added a test for IP addresses.

mpritham:
Closing this PR. Opened #3166 to track.

PR #3170, opened 2025-07-21
Support both okhttp 4.x and 5.x while staying on 4.x for now

aldexis:
## Before this PR
https://github.com/palantir/conjure-java-runtime/pull/3154 was trying to force an upgrade of okhttp to 5.x, which changed the okhttp maven coordinates as well as introduced some ABI breaks.

Because it requires manual fixing, bumping okhttp in CJR would likely cause breaks in downstream consumers, either blocking conjure upgrades or (worse) causing runtime errors.

Instead, this PR makes CJR compatible with both okhttp 4.x and okhttp 5.x classes, such that regardless of which one we get, we should be fine (even if both okhttp:4.x and okhttp-jvm:5.x are present on the classpath, it should be fine given the classnames are the same, and it doesn't matter here which one we get).

It's also important to fix this to avoid runtime errors due to ABI breaks for consumers that may have upgraded to 5.x.

## After this PR

==COMMIT_MSG==
Support both okhttp 4.x and 5.x while staying on 4.x for now
==COMMIT_MSG==

https://github.com/palantir/conjure-java-runtime/pull/3170/commits/dfcf1f2418720a4a405a4fcef1705e54630994dc tests bumping to 5.x to verify compatibility with the new code ([circle build](https://app.circleci.com/pipelines/github/palantir/conjure-java-runtime/7410/workflows/5f20eb21-3a55-415b-bb01-fc3ea17615c5))

## Possible downsides?
This perpetuates usage of okhttp. Unfortunately, we haven't gotten rid of all usage, even internally, and until this is done, we are stuck needing to support this. Fortunately, this PR is pretty small and shouldn't be much of a concern.


bjlaub (approved):

bjlaub:
i assume that [dfcf1f2](https://github.com/palantir/conjure-java-runtime/pull/3170/commits/dfcf1f2418720a4a405a4fcef1705e54630994dc) was only to verify that tests indeed pass on okhttp 5?

aldexis:
Exactly! (I mentioned it in the PR description as well)

PR #3192, opened 2025-08-12
Add `withConjureErrorParameterFormat` option to `WithClientOptions`

mpritham:
## Before this PR
I'd like to enable clients to set the desired error parameter format. We currently don't have a method on Dialogue's `ReloadingFactory` to do so.

## After this PR

==COMMIT_MSG== 
Add `withConjureErrorParameterFormat` option to `WithClientOptions`
==COMMIT_MSG==

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -79,3 +79,8 @@ acceptedBreaks:
     - code: "java.class.removed"
       old: "enum com.palantir.conjure.java.server.jersey.ErrorCause"
       justification: "ErrorCause is never used by consumers"
+  "8.22.0":
+    com.palantir.conjure.java.runtime:client-config:
+    - code: "java.method.addedToInterface"
+      new: "method T com.palantir.conjure.java.clients.ConjureClients.WithClientOptions<T>::withConjureErrorParameterSerializationFormat(com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat)"
+      justification: "Allowing clients to request error parameter serialization format"
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -82,5 +82,6 @@ acceptedBreaks:
   "8.22.0":
     com.palantir.conjure.java.runtime:client-config:
     - code: "java.method.addedToInterface"
-      new: "method T com.palantir.conjure.java.clients.ConjureClients.WithClientOptions<T>::withConjureErrorParameterSerializationFormat(com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat)"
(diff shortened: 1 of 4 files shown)


bjlaub (dismissed):
approving for RC

client-config/src/main/java/com/palantir/conjure/java/clients/ConjureClients.java
```diff
@@ -88,6 +89,8 @@ public interface WithClientOptions<T> {
 
         /** Per-host success/failure information will be recorded to this sink. */
         T withHostEventsSink(HostEventsSink hostEventsSink);
+
+        T withConjureErrorParameterSerializationFormat(ConjureErrorParameterFormat format);
```
bjlaub: nit: what about calling this just `withConjureErrorParameterFormat`?

mpritham: That works. Updated.

client-config/build.gradle
```diff
@@ -3,6 +3,7 @@ apply plugin: 'org.revapi.revapi-gradle-plugin'
 
 dependencies {
     api 'com.palantir.conjure.java.api:service-config'
+    api 'com.palantir.conjure.java.api:errors'
```
mpritham, line 6: This is needed because `ConjureErrorParameterFormat` is defined in `com.palantir.conjure.java.api:errors`

.palantir/revapi.yml
```diff
@@ -79,3 +79,13 @@ acceptedBreaks:
...
+  "8.24.0-rc1":
+    com.palantir.conjure.java.runtime:client-config:
+    - code: "java.method.addedToInterface"
+      new: "method T com.palantir.conjure.java.clients.ConjureClients.WithClientOptions<T>::withConjureErrorParameterFormat(com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat)"
+      justification: "Adding method to allow clients to request error parameter serialization\
+        \ format"
+    - code: "java.method.removed"
+      old: "method T com.palantir.conjure.java.clients.ConjureClients.WithClientOptions<T>::withConjureErrorParameterSerializationFormat(com.palantir.conjure.java.api.errors.ConjureErrorParameterFormat)"
+      justification: "Adding method to allow clients to request error parameter serialization\
+        \ format"
```
aldexis: This change seems weird?

mpritham: Which part was weird? If it was because the break was on an RC release, that has been fixed.

aldexis: I think I might have meant that it was saying a method was removed

aldexis (approved):

PR #3200, opened 2025-08-20
Optimize query parameter key=value parsing

schlosna:
## Before this PR

In services that make high throughput RPCs with small responses (think similar clients of timestamps, locks, authorization results, etc. as https://github.com/palantir/conjure-java-runtime/pull/3114 ), `String#charAt` & `com.google.common.base.Splitter#splitToList` pop up as hot methods & allocators due to `DialogueFeignClient.FeignEndpoint#renderPath` query parameter `key=value` parsing.

image

## After this PR

==COMMIT_MSG==
Use vectorized `indexOf` to improve query parameter parsing and reduce allocation overhead.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
+++ b/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
@@ -73,7 +73,6 @@ final class DialogueFeignClient implements feign.Client {
     private static final Splitter PATH_SPLITTER = Splitter.on('/');
     private static final Splitter QUERY_SPLITTER = Splitter.on('&').omitEmptyStrings();
     private static final char QUERY_KEY_VALUE_SEPARATOR = '=';
-    private static final Splitter QUERY_KEY_VALUE_SPLITTER = Splitter.on(QUERY_KEY_VALUE_SEPARATOR);
 
     private final ConjureRuntime runtime;
     private final Channel channel;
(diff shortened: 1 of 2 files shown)


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -417,14 +418,17 @@ public void renderPath(ListMultimap<String, String> params, UrlBuilder url) {
             if (queryParamsStart != -1) {
                 String querySegments = trailing.substring(queryParamsStart + 1);
                 for (String querySegment : QUERY_SPLITTER.split(querySegments)) {
-                    List<String> keyValuePair = QUERY_VALUE_SPLITTER.splitToList(querySegment);
-                    if (keyValuePair.size() != 2) {
+                    int equalsIndex = querySegment.indexOf(QUERY_KEY_VALUE_SEPARATOR);
```
pkoenig10, line 421: We probably want to ensure that there aren't more than 2 segments.

conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java
```diff
@@ -417,14 +418,17 @@ public void renderPath(ListMultimap<String, String> params, UrlBuilder url) {
...
+                    if (equalsIndex > 0) {
+                        String key = querySegment.substring(0, equalsIndex);
+                        String value = querySegment.substring(equalsIndex + 1);
+                        url.queryParam(urlDecode(key), urlDecode(value));
+                    } else {
```
pkoenig10, line 426: nit: I'd prefer if we wrote this as if-throw to avoid additional nesting in the non-exceptional case.

PR #3307, opened 2025-11-19
Bump jersey to latest 2.x and 3.x versions

aldexis:
## Before this PR
We're relying on Jersey 3.1.3, which contains https://nvd.nist.gov/vuln/detail/CVE-2025-12383.

Sounds like there is 4.0.0 that has released, but bumping an entire major version seems riskier, and the vuln has been addressed as of 3.1.10.

Here are the release notes for the various versions:
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.11
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.10
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.9
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.8
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.7
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.6
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.5
- https://github.com/eclipse-ee4j/jersey/releases/tag/3.1.4

## After this PR

==COMMIT_MSG==
Bump jersey to latest 3.x version
==COMMIT_MSG==

## Possible downsides?
There are a lot of changes (two years worth of them), so determining potential impact is hard.


conjure-java-jersey-server/baseline-class-uniqueness.lock
```diff
@@ -0,0 +1,50 @@
+# Danger! Multiple jars contain identically named classes. This may cause different behaviour depending on classpath ordering.
+# Run ./gradlew checkClassUniqueness --fix to update this file
+
+## runtimeClasspath
+[javax.el:javax.el-api, org.glassfish:jakarta.el]
```
aldexis: I don't understand why this is needed, when `org.glassfish:jakarta.el` seems to be removed from our versions.lock 🤔

aldexis: This is because 
- We're forcing down the version in https://github.com/palantir/conjure-java-runtime/blob/75af32290670939e68817f0c1f6aaf86207976a0/conjure-java-jersey-server/build.gradle#L4
- We were _not_ forcing down the version of `org.glassfish:jakarta.el`, so we would get the higher version from the versions.lock file
  image
- The jersey bump removed `org.glassfish:jakarta.el` as a transitive dependency, thus not forcing the version up anymore, leading to us using the requested 3.0.4, which has the conflicts

Because we were actually using a higher version anyway, which didn't contain the classes, we can just exclude it as a transitive dependency, to continue relying on `javax.el:javax.el-api`.

conjure-java-jersey-server/build.gradle
```diff
@@ -45,6 +45,8 @@ dependencies {
         version {
             strictly jerseyVersion
         }
+        // This conflicts with org.glassfish:jakarta.el (which contains the same classes), when at the version requested by 2.40
```
bjlaub, line 48: the details of CVE-2025-12383 are a bit unclear right now, but perhaps we should bump https://github.com/palantir/conjure-java-runtime/pull/3307/files#diff-5f9e3e7b8a9bffc2554597f5c9408adc3b41614674d053b9cdf6c1b8ba08c4cbR4 to the latest version as well (2.47)?

bjlaub:
👍

PR #3312, opened 2025-11-25
Make DefaultCas public to allow building custom truststores

Eric-Alvarez:
## Before this PR

DefaultCas is package-private which does not allow building custom truststores that trust default cas
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==
DefaultCas class is public to allow clients to build custom truststores


pkoenig10:
I don't think it should be this libraries responsibility to provide this utility for arbitrary clients. Different clients may have very different expectations or requirements for what is included here. Exposing this API greatly increases the responsibility shouldered by this library for things that are mostly not it's concern.

If we wanted to expose a library for this, it should be pulled out to a dedicated library where that is expressly it's goal.

PR #3355, opened 2026-01-13
Bump feign to openfeign 13

ilyanep:
## Before this PR
Conjure-java-runtime depends on Netflix Feign 8, which is very old and winds up blocking consumers from upgrading their own Feign dependencies

## After this PR

==COMMIT_MSG==
Conjure-java-runtime now depends on Openfeign 13. Consumers relying on transitive dependencies will need to update their own dependencies.
==COMMIT_MSG==

## Possible downsides?

Difficulty to upgrade


conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/CompatibleJaxRsContract.java
```diff
@@ -180,4 +184,12 @@ private static Object getAnnotationValueInternal(Annotation annotation) {
                     "Failed to read annotation value", e, SafeArg.of("annotationType", annotation.annotationType()));
         }
     }
+
+    private Collection<String> addTemplatedParam(Collection<String> possiblyNull, String name) {
```
ilyanep, line 188: This was removed from `BaseContract` upstream. I just copied the old version in.

PR #3427, opened 2026-03-26
Remove some Feign dependencies

pkoenig10:
`conjure-java-runtime` has a dependency on an old, unmaintained version of Feign. This is causing classpath conflicts with consumers are using a modern version of Feign, as the package has changed to `io.github.openfeign:feign-core` and the library has gone through a couple major version changes.

This PR is a small first step that eliminates our Feign dependency in a couple places where it is no longer needed.

After this PR, `conjure-java-jaxrs-client` is the only project that still has a Feign dependency. In a follow-up, I'd like to fork and vendor Feign as a private implementation detail of this project.


client-config/src/main/java/com/palantir/conjure/java/client/config/ClientConfiguration.java
```diff
@@ -133,22 +131,9 @@ default void check() {
             checkArgument(maxNumRetries() == 0, "If meshProxy is configured then maxNumRetries must be 0");
             checkArgument(uris().size() == 1, "If meshProxy is configured then uris must contain exactly 1 URI");
         }
-        // Assert that timeouts are in milliseconds, not any higher precision, because feign only supports millis.
-        checkTimeoutPrecision(connectTimeout(), "connectTimeout");
-        checkTimeoutPrecision(readTimeout(), "readTimeout");
-        checkTimeoutPrecision(writeTimeout(), "writeTimeout");
```
pkoenig10, line 139: This is no longer relevant. Since https://github.com/palantir/conjure-java-runtime/pull/1596 we no longer use these values to configure the Feign `Options`. We no longer use the default Feign client (which respects these options) and instead use our Dialogue-based client, which ignores them (because this is handled by the underlying Dialogue client).

https://github.com/palantir/conjure-java-runtime/blob/df1a6d4082e106bd1da5fa80826ba2f41f4d3863/conjure-java-jaxrs-client/src/main/java/com/palantir/conjure/java/client/jaxrs/DialogueFeignClient.java#L96

conjure-java-jersey-jakarta-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
```diff
@@ -44,7 +44,6 @@ private static boolean configure(FeatureContext context, ExceptionListener excep
         // Exception mappers
         context.register(new NoContentExceptionMapper());
         context.register(new IllegalArgumentExceptionMapper(exceptionListener));
-        context.register(new RetryableExceptionMapper(exceptionListener));
```
pkoenig10, line 47: This isn't providing any value. Without this, these errors will just get handled by `RuntimeExceptionMapper`, which has identical behavior.

Not to mention that:
- Most client errors should be handled inside the Dialogue client and not make it to the Feign wrapper where they may be wrapped in `RetryableExceptionMapper`.
- We have relatively few endpoints that still use Jersey.

okhttp-clients/src/main/java/com/palantir/conjure/java/okhttp/OkHttpClients.java
```diff
@@ -257,8 +257,6 @@ private static RemotingOkHttpClient createInternal(
         client.addInterceptor(UserAgentInterceptor.of(augmentUserAgent(agent, serviceClass)));
 
         // timeouts
-        // Note that Feign overrides OkHttp timeouts with the timeouts given in FeignBuilder#Options if given, or
-        // with its own default otherwise.
```
pkoenig10, line 261: No longer correct, for the reason mentioned above.

iamdanfox (dismissed):
I'm supportive of this because severing the compile dependency from `conjure-java-jersey-jakarta-server -> com.netflix.feign:feign-core` means that I think I can safely exclude this old feign-core version

conjure-java-jersey-jakarta-server/src/main/java/com/palantir/conjure/java/server/jersey/ConjureJerseyFeature.java
aldexis: I'm not terribly familiar with the jersey codepaths and what you say makes sense. I still wonder: do we have a test exercising the `RetryableException` -> `ErrorType.INTERNAL` mapping somewhere?

pkoenig10: The answer is no, in the sense that we don't have specific tests for arbitrary `RuntimeException` subtypes that might be thrown. After this PR, there is no functional difference between `RetryableException` and a `RuntimeException`.

aldexis (dismissed):

aldexis (approved):

PR #3429, opened 2026-03-27
Remove Feign dependency

pkoenig10:
We currently depend on a very old and no longer maintained version of Feign. Upgrading is tricky because there are a number of subtle behavior changes in the latest version of Feign - see https://github.com/palantir/conjure-java-runtime/pull/1337. This code is quite ossified, so even small behavior changes are likely to cause problems.

This is starting to cause conflicts with consumers are also depending on modern version of Feign.

This PR addresses the issue by removing our Feign dependency. This is accomplished by vendoring the code from the version of Feign we currently depend on.

This PR is divided into a series of commits in order to make reviewing easier:
1. **Vendor Feign as is into `conjure-java-jaxrs-client`**
1. **Move all classes to the `com.palantir.conjure.java.client.jaxrs`**

    `conjure-java-jaxrs-client` currently exposes a lot of it's internals, mostly because internal classes don't all exist in the same package. Changing the interfaces that our classes implement is an already an ABI, so we should take the opportunity to address this deficiency while we are breaking the API.

    There are a small handful of internal consumers that are importing some of the `feignimpl` classes. But it is small enough (roughly 13 files across 8 repos), that this change should be acceptable.

1. **Remove unused code, suppress warnings, and make internal things package private**

    This includes some small refactors if those refactors allowed us to eliminate more code.

There are many additional follow up refactors that can be done to further clean up and simplify this code.


pkoenig10:
I may take the time to experiment with a more substantive change, where we instead build this more from the ground up and remove all the unnecessary Feign intermediate parts. Basically all we need from Feign is the logic in `Contract` to parse a JAX-RS annotated class into metadata that we can use to make requests. For example, I'm imaging that we can parse the JAX-RS annotated class directly into `Endpoint` implementations.

PR #3512, opened 2026-06-17
Add JacksonWarmup to pre-settle JIT

schlosna:
## Before this PR

OpenJDK could encounter pathological C2 JIT deoptimization storm in Jackson [BeanPropertyWriter.java:731](https://github.com/FasterXML/jackson-databind/blob/3c95742e2db8fbe360e7bedfa900e9d2b996c8d7/src/main/java/com/fasterxml/jackson/databind/ser/BeanPropertyWriter.java#L731-L735) leading to ~10x performance degradation and significant increases in CPU utilization.

See https://bugs.openjdk.org/browse/JDK-8330258 & https://bugs.openjdk.org/browse/JDK-8374307

OpenJDK has an open PR https://github.com/openjdk/jdk/pull/28966 that may address this issue by switching from checking `too_many_traps` to `too_many_traps_or_recompiles`.

## After this PR
==COMMIT_MSG==
Adds a JacksonWarmup type that serializes a representative set of payload shapes through fresh server and dialogue/client `ObjectMapper`s (constructed via the same conjure-java factories used at runtime) so the JIT settles on the bimorphic `BeanPropertyWriter.serializeAsField` branch (typed vs untyped serializer) with full type-profile coverage before customer traffic perturbs it.

This works around OpenJDK C2 JIT deoptimization storm where `uncommon_trap` `isnonnull` triggers continuous deoptimization as seen in https://bugs.openjdk.org/browse/JDK-8330258 &
https://bugs.openjdk.org/browse/JDK-8374307 . The repeated deoptimization can be observed by `jdk.Deoptimization` JFR events at `com.fasterxml.jackson.databind.ser.BeanPropertyWriter` `if (_typeSerializer == null)` site, and then re-enters `unstable_if` deopt thrash on the serialization path when a perturbing workload arrives.

The warmup payloads interleave monomorphic fields (concrete types) with polymorphic fields (sealed interface + `@JsonTypeInfo`) in the same object so each iteration exercises both branches, letting C2 record a bimorphic profile during the warmup pass.

Warmup failures are caught and logged rather than aborting as a degraded JIT profile is preferable to refusing to serve traffic.
==COMMIT_MSG==

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
@@ -24,6 +24,9 @@ import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -35,17 +38,31 @@ import java.util.stream.Stream;
  * (typed vs untyped serializer) with full type-profile coverage before customer traffic perturbs it.
  *
  * <p>Without this warmup, a hot host that gets restarted and then receives a perturbing workload can re-enter
- * {@code unstable_if} deopt thrash on the JSON serialization path. The bytecode site is
- * {@code if (_typeSerializer == null)} in {@code BeanPropertyWriter.serializeAsField}: properties feeding
- * monomorphic serializers take one branch while properties feeding polymorphic serializers (Conjure unions,
- * {@code @JsonTypeInfo} types) take the other. The JIT speculates one branch, hits the other, deopts, recompiles,
- * and repeats - manifesting as kernel/system CPU from safepoints, code-cache invalidation, and recompilation.
+ * {@code unstable_if} deopt thrash on the JSON serialization path. The bytecode site is the
+ * {@code if (_typeSerializer == null)} check in {@code BeanPropertyWriter.serializeAsField}
+ * (<a href="https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.21.4/src/main/java/com/fasterxml/jackson/databind/ser/BeanPropertyWriter.java#L731-L735">jackson-databind
+ * 2.21.4 source</a>): properties feeding monomorphic serializers take one branch while properties feeding
+ * polymorphic serializers (Conjure unions, {@code @JsonTypeInfo} types) take the other. The JIT speculates
+ * one branch, hits the other, deopts, recompiles, and repeats - manifesting as kernel/system CPU from
+ * safepoints, code-cache invalidation, and recompilation. Throughput regressions of ~10x have been
+ * observed in production when this fires.
+ *
+ * <p>One might expect C2 to learn the bimorphic profile after a few deopts and stop. It does not: HotSpot
+ * suppresses re-profiling at sites that have already produced an {@code uncommon_trap} (the
+ * {@code too_many_traps} predicate, see Shipilev's
+ * <a href="https://shipilev.net/jvm/anatomy-quarks/29-uncommon-traps/">"Uncommon Traps" article</a>), so the
+ * site stays compiled for the wrong branch and keeps deoptimizing. The upstream fix is
+ * <a href="https://github.com/openjdk/jdk/pull/28966">openjdk/jdk#28966</a>, which switches the predicate to
+ * {@code too_many_traps_or_recompiles}. Until that lands and propagates, this class works around the bug
+ * by feeding C2 a bimorphic profile at startup for the Jackson serialization path.
  *
  * <p>The warmup payloads include both kinds of fields in the same object so each iteration exercises both branches,
  * letting C2 record a bimorphic profile for {@code serializeAsField} during warmup rather than learning it from
- * production traffic.
+ * production traffic. The bimorphic dispatch this leaves behind is essentially free (a single null check plus
+ * an inline cache with two entries), and is many orders of magnitude cheaper than the recompilation storm it
+ * prevents, even on hosts that only ever exercise one branch in steady state.
  *
- * <p>See OpenJDK issues:<ul>
+ * <p>See also OpenJDK issues:<ul>
  * <li> <a href="https://bugs.openjdk.org/browse/JDK-8330258">JDK-8330258</a> </li>
  * <li> <a href="https://bugs.openjdk.org/browse/JDK-8374307">JDK-8374307</a> </li>
  * </ul>
@@ -54,8 +71,12 @@ public final class JacksonWarmup {
 
     private static final SafeLogger log = SafeLoggerFactory.get(JacksonWarmup.class);
 
-    // Comfortably above HotSpot's tiered-compilation threshold (10_000) so C2 compilation completes during warmup.
-    static final int ITERATIONS = 20_000;
+    // Each iteration runs every mapper against every payload once, so per (mapper, payload) we hit ITERATIONS
+    // top-level writeValue invocations and many more invocations of the shared inner methods like
+    // BeanPropertyWriter.serializeAsField (one per field per payload). This is comfortably above HotSpot's
+    // tiered-compilation invocation thresholds (Tier4InvocationThreshold defaults to 5_000) so C2 settles on
+    // the bimorphic profile during warmup. Total warmup wall-clock is O(1s) on modern hardware.
+    static final int ITERATIONS = 10_000;
 
     private JacksonWarmup() {}
 
@@ -89,19 +110,27 @@ public final class JacksonWarmup {
         long count = 0;
         try (OutputStream sink = new DiscardOutputStream()) {
             List<Object> payloads = buildPayloads();
-            for (ObjectMapper mapper : mappers) {
-                try {
-                    for (int i = 0; i < ITERATIONS; i++) {
+            // Iteration is the outer loop and mappers is the inner loop so each iteration exercises every mapper.
+            // The hot methods being warmed (BeanPropertyWriter.serializeAsField, the per-field serializers) are
+            // JVM-wide compiled, so interleaving lets every mapper contribute to the type profile rather than
+            // letting the first mapper dominate and the rest only confirm it.
+            List<ObjectMapper> active = new ArrayList<>(Arrays.asList(mappers));
+            for (int i = 0; i < ITERATIONS && !active.isEmpty(); i++) {
+                for (Iterator<ObjectMapper> it = active.iterator(); it.hasNext(); ) {
+                    ObjectMapper mapper = it.next();
+                    try {
                         for (Object payload : payloads) {
                             mapper.writeValue(sink, payload);
                             count++;
                         }
+                    } catch (IOException | RuntimeException e) {
+                        log.warn(
+                                "Jackson ObjectMapper warmup failed; continuing without it",
+                                SafeArg.of("mapperClass", mapper.getClass().getCanonicalName()),
+                                e);
+                        // Drop the mapper from the active set so we don't spam logs on every subsequent iteration.
+                        it.remove();
                     }
-                } catch (IOException | RuntimeException e) {
-                    log.warn(
-                            "Jackson ObjectMapper warmup failed; continuing without it",
-                            SafeArg.of("mapperClass", mapper.getClass().getCanonicalName()),
-                            e);
                 }
             }
             long elapsedMillis = (System.nanoTime() - startNanos) / 1_000_000L;
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
@@ -1,0 +1,224 @@
+/*
+ * (c) Copyright 2026 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.java.serialization;
+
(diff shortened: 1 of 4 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
```diff
@@ -0,0 +1,224 @@
...
+ * Synchronously serializes a representative set of payload shapes through both the server and dialogue/client
+ * {@link ObjectMapper}s so the JIT settles on the bimorphic {@code BeanPropertyWriter.serializeAsField} branch
+ * (typed vs untyped serializer) with full type-profile coverage before customer traffic perturbs it.
+ *
+ * <p>Without this warmup, a hot host that gets restarted and then receives a perturbing workload can re-enter
+ * {@code unstable_if} deopt thrash on the JSON serialization path. The bytecode site is
+ * {@code if (_typeSerializer == null)} in {@code BeanPropertyWriter.serializeAsField}: properties feeding
+ * monomorphic serializers take one branch while properties feeding polymorphic serializers (Conjure unions,
+ * {@code @JsonTypeInfo} types) take the other. The JIT speculates one branch, hits the other, deopts, recompiles,
+ * and repeats - manifesting as kernel/system CPU from safepoints, code-cache invalidation, and recompilation.
```
aldexis: Maybe add links to the relevant pieces of code in Jackson (and maybe also openjdk) so it's easier to understand what this means? Without knowing a lot about how JIT compilation works, it's a bit hard to follow

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
```diff
@@ -0,0 +1,224 @@
...
+ * <li> <a href="https://bugs.openjdk.org/browse/JDK-8330258">JDK-8330258</a> </li>
+ * <li> <a href="https://bugs.openjdk.org/browse/JDK-8374307">JDK-8374307</a> </li>
+ * </ul>
+ */
+public final class JacksonWarmup {
+
+    private static final SafeLogger log = SafeLoggerFactory.get(JacksonWarmup.class);
+
+    // Comfortably above HotSpot's tiered-compilation threshold (10_000) so C2 compilation completes during warmup.
+    static final int ITERATIONS = 20_000;
```
aldexis: It would be good to confirm whether we need such a high threshold, especially since we're serializing many payloads across many mappers.

More generally, it would be nice to be able to observe and confirm the effects somehow, as otherwise we're just blindly introducing changes. Is there maybe some way to use JMH for this? (to compare both with/without warmup, as well as warmup perf)

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
aldexis: Shouldn't the JIT instead recompile based on the new profile after some point?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
```diff
@@ -0,0 +1,224 @@
...
+ * <p>Without this warmup, a hot host that gets restarted and then receives a perturbing workload can re-enter
+ * {@code unstable_if} deopt thrash on the JSON serialization path. The bytecode site is
+ * {@code if (_typeSerializer == null)} in {@code BeanPropertyWriter.serializeAsField}: properties feeding
+ * monomorphic serializers take one branch while properties feeding polymorphic serializers (Conjure unions,
+ * {@code @JsonTypeInfo} types) take the other. The JIT speculates one branch, hits the other, deopts, recompiles,
+ * and repeats - manifesting as kernel/system CPU from safepoints, code-cache invalidation, and recompilation.
+ *
+ * <p>The warmup payloads include both kinds of fields in the same object so each iteration exercises both branches,
+ * letting C2 record a bimorphic profile for {@code serializeAsField} during warmup rather than learning it from
+ * production traffic.
```
aldexis: Would this risk decreasing perf for services that only use one branch and can benefit from a more dedicated compilation?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
```diff
@@ -0,0 +1,224 @@
...
+     * @return count of payloads warmed up
+     */
+    public static long run(ObjectMapper... mappers) {
+        long startNanos = System.nanoTime();
+        long count = 0;
+        try (OutputStream sink = new DiscardOutputStream()) {
+            List<Object> payloads = buildPayloads();
+            for (ObjectMapper mapper : mappers) {
+                try {
+                    for (int i = 0; i < ITERATIONS; i++) {
```
aldexis: Shouldn't we flip these two? Otherwise, aren't we going to land on optimizing for the first mapper, and not the rest, since we're already running 20k iterations just on the first mapper?

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: We could probably drop this a bit. Overall warmup is `O(1 second)` on systems I've tested recently.

I've manually confirmed locally via coverage analysis. I don't necessarily want to pull in jtreg and JIT Whitebox into CJR, but can try to put together a reproducer with JIT compiler output.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: The JIT bug is the repeated recompilation with unstable `if` uncommon trap. This PR works around the JIT bug only for the Jackson case by feeding it the bimorphic profile path at startup so it optimizes both if & else blocks. The real fix is something like https://github.com/openjdk/jdk/pull/28966/ to handle `too_many_recompiles` in addition to `too_many_traps` when deciding suitability for uncommon trap use (i.e. it is no longer uncommon enough), see https://shipilev.net/jvm/anatomy-quarks/29-uncommon-traps/

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
aldexis: Having a reproducer would be great, so we can easily confirm whether the issue is still present in the future

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: Done in fd397903 — the class JavaDoc now links to the Jackson source line (`BeanPropertyWriter.serializeAsField` at `jackson-databind-2.21.4`), the OpenJDK PR (`openjdk/jdk#28966`), and Shipilev's "Uncommon Traps" article, and adds a paragraph explaining why C2 doesn't recover on its own (the `too_many_traps` predicate suppresses re-profiling after the first uncommon trap).

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: Done in fd397903 — dropped `ITERATIONS` from 20_000 to 10_000 (still comfortably above `Tier4InvocationThreshold`=5_000; per (mapper, payload) we still hit ~10k top-level invocations and far more inner `BeanPropertyWriter` invocations). And see the reproducer below — answering the more important question of whether the workaround actually works.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: The expanded JavaDoc covers this in fd397903 — the short version is HotSpot suppresses re-profiling at sites that have already produced an `uncommon_trap` (the `too_many_traps` predicate, see [Shipilev's article](https://shipilev.net/jvm/anatomy-quarks/29-uncommon-traps/)), so the site stays compiled for the wrong branch and keeps deoptimizing. The upstream fix is [openjdk/jdk#28966](https://github.com/openjdk/jdk/pull/28966) which switches the predicate to `too_many_traps_or_recompiles`.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: In short: no, the bimorphic dispatch is essentially free. The JavaDoc now spells it out — the leftover cost is a single null check plus an inline cache with two entries, which is many orders of magnitude cheaper than the deopt storm it prevents (~10x throughput hit when it fires), even on hosts that only ever exercise one branch in steady state. Picking up production traffic perturbations cheaply is exactly the point.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: Done in fd397903 — iteration is the outer loop, mappers is inner, and failed mappers drop out of the active set after the first exception so we don't spam logs on every subsequent iteration.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: Done in 85492267 — `JacksonWarmupReproducerTest` forks a pair of child JVMs with `-XX:+PrintCompilation`, drives a monomorphic-train (`Variant.StringVariant`) → bimorphic-perturb (`Payload`) workload through the package-protected types, parses the resulting PrintCompilation stream, and asserts:

- without warmup, at least one `made not entrant: uncommon trap` event on `BeanPropertyWriter.serializeAsField`
- with warmup, zero such events

It's opt-in (`-Djackson.warmup.reproducer=true`) so it doesn't slow normal CI; runtime is ~2s end-to-end on a recent Mac. Sample output today on Corretto 25.0.3:

**Without warmup:**
```
259 1218       3       BeanPropertyWriter::serializeAsField
396 1300       4       BeanPropertyWriter::serializeAsField
404 1218       3       BeanPropertyWriter::serializeAsField   made not entrant: not used
837 1300       4       BeanPropertyWriter::serializeAsField   made not entrant: uncommon trap   ← bug
847 1434       3       BeanPropertyWriter::serializeAsField
860 1560       4       BeanPropertyWriter::serializeAsField
```

**With warmup:**
```
257 1262       3       BeanPropertyWriter::serializeAsField
276 1472       4       BeanPropertyWriter::serializeAsField
280 1262       3       BeanPropertyWriter::serializeAsField   made not entrant: not used
```

If a future JDK fixes JDK-8330258/JDK-8374307, the first assertion will fail loudly and signal that `JacksonWarmup` may no longer be necessary.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: updated

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: updated

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: Dropped this to 10k and validated with added https://github.com/palantir/conjure-java-runtime/blob/davids/jackson-warmup/conjure-java-jackson-serialization/src/test/java/com/palantir/conjure/java/serialization/JacksonWarmupReproducerTest.java to validate no warmup vs. warmup with `-XX:+PrintCompilation` output

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: There's potentially a very small optimization and inlining that we may miss out on by prewarming so that it keeps both branches, but I would say the benefits outweigh that potential loss given the catastrophic overhead of deoptimization storm on performance where that JVM's serialization throughput is degraded by orders of magnitude until it is killed, in addition to the noisy neighbor CPU overhead degradation.

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
```diff
@@ -0,0 +1,253 @@
...
+ * polymorphic serializers (Conjure unions, {@code @JsonTypeInfo} types) take the other. The JIT speculates
+ * one branch, hits the other, deopts, recompiles, and repeats - manifesting as kernel/system CPU from
+ * safepoints, code-cache invalidation, and recompilation. Throughput regressions of ~10x have been
+ * observed in production when this fires.
+ *
+ * <p>One might expect C2 to learn the bimorphic profile after a few deopts and stop. It does not: HotSpot
+ * suppresses re-profiling at sites that have already produced an {@code uncommon_trap} (the
+ * {@code too_many_traps} predicate, see Shipilev's
+ * <a href="https://shipilev.net/jvm/anatomy-quarks/29-uncommon-traps/">"Uncommon Traps" article</a>), so the
+ * site stays compiled for the wrong branch and keeps deoptimizing. The upstream fix is
```
aldexis, line 54: You indicate that HotSpot does not re-profile when it has produced an `uncommon_trap`. What about when it has not? Could we actually get into a state where we run this, but then use the monomorphic serializer codepath for a while, such that HotSpot ends up reprofiling and recompiling, this time reintroducing the `uncommon_trap` we tried to avoid?

Edit: Seems like this is working as expected, based on reading your tests, since you train in both cases on the monomorphic string variant

aldexis (approved):

conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/JacksonWarmup.java
schlosna: I think there could potentially be cases today or in future where JIT may decide to reprofile and insert an `uncommon_trap`, but I think the likelihood after this shape of warmup should be quite low. This `JacksonWarmup` is a very targeted mitigation against one observed pathological deoptimization condition -- the systematic solution is fixing this upstream in JDK.

PR #3534, opened 2026-07-09
Add opt-in Jackson Blackbird optimization to ObjectMappers

parker89:
Adds opt-in Jackson Blackbird support to `ObjectMappers` (off by default). `ObjectMapperOptimizations.createModules(boolean)` plus new server/client mapper overloads register `BlackbirdModule` only when explicitly requested.

### DIFF
--- a/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
+++ b/conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
@@ -50,7 +50,12 @@ public final class ObjectMappers {
      * </ul>
      */
     public static JsonMapper newClientJsonMapper() {
-        return withDefaultModules(JsonMapper.builder(jsonFactory()))
+        return newClientJsonMapper(false);
+    }
+
+    /** Like {@link #newClientJsonMapper()}, optionally enabling Jackson optimization modules (Blackbird). */
+    public static JsonMapper newClientJsonMapper(boolean useOptimizations) {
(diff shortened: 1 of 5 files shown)


conjure-java-jackson-serialization/src/main/java/com/palantir/conjure/java/serialization/ObjectMappers.java
```diff
@@ -127,7 +152,12 @@ public static CBORMapper newServerCborMapper() {
      * </ul>
      */
     public static SmileMapper newServerSmileMapper() {
-        return withDefaultModules(SmileMapper.builder(smileFactory()))
+        return newServerSmileMapper(false);
+    }
+
+    /** Like {@link #newServerSmileMapper()}, optionally enabling Jackson optimization modules (Blackbird). */
+    public static SmileMapper newServerSmileMapper(boolean useOptimizations) {
```
mpritham, line 159: Did you have other optimizations in mind down the line? 

I wonder if we should name this something like "newSingleInstantiatedServerSmileMapper" or something to make it clear that this should not be called multiple times, and then enable blackbird by default for this case.

Maybe we can write an errorprone check(? e.g. only created in a final variable in a class - though this is a bit restrictive)

bjlaub:
I think a big issue I have with this change is that it is not obvious that calling `ObjectMappers.new*Mapper(true)` should be done with great care due to https://github.com/FasterXML/jackson-modules-base/issues/147.

We have quite a bit of code internally that uses this pattern; existing code won't be affected (which is good), but new code can very easily be written in a way that causes the leak. Using a memoization approach might be better, with a separate API that makes it very obvious that you are getting a shared ObjectMapper, not a new one.

